PR #7, opened 2018-06-28
Convert DGXie to Helm chart

dholt:
Needed to make DGXie configuration possible without modifying the kubernetes manifest

### DIFF
--- a/services/dgxie/values.yaml
+++ b/services/dgxie/values.yaml
@@ -1,0 +1,51 @@
+# Default values for dgxie.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# DGXie config
+bootMode: DGX
+isoPath: /iso
+dgxNetInt: enp1s0f0
+dgxDisk: sda
+dgxKbd: us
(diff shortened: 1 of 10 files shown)


dholt:
@nvhans do you mind testing this out when you get a chance?

services/dgxie/values.yaml
```diff
@@ -0,0 +1,51 @@
...
+netPrvDhcpStart: 192.168.1.100
+netPrvDhcpEnd: 192.168.1.199
+netPrvDhcpLease: 7200
+httpPort: 13370
+
+# Deployment config
+replicaCount: 1
+
+image:
+  repository: dgxie
```
ryanolson: should this be `registry.local/deepops/dgxie`?

dholt: currently you build the dgxie container locally and copy it over to the mgmt host and load it in via docker, so it's not coming from the registry. it probably would be a better idea to bootstrap the registry and put the container there, but this config correct for now

ryanolson: 👍

PR #9, opened 2018-06-29
Clean up Helm instructions, configure Ingress and add container registry with Helm

dholt:


config.example/registry.yml
```diff
@@ -0,0 +1,29 @@
+#
+# Container registry configuration
+#
+
+# Storage
+persistence:
+  # Use persistent storage to retain data
+  enabled: true
+  # Size of persistent storage volume
+  size: 10Gi
```
ryanolson: that's pretty small

dholt: what do you want for a default? my VM only has 40Gi :)

PR #31, opened 2018-11-15
Permit to not have a public interface

hightoxicity:
If no pub interface is set (empty env var), then no NAT rules are set, useful if we use a proxy to preseed.

Thanks for review.


hightoxicity:
In #39

PR #34, opened 2018-11-27
Make bootstrap works with proxies

hightoxicity:
Make possible to bootstrap using proxies.
Use the same way to declare proxies as the one used in kubespray.
Ensure backward compatibility (parameters declaration is optional).

Thanks for review.


hightoxicity:
In #39

PR #36, opened 2018-11-29
Permit to provide extra packages to install at preseed

hightoxicity:
Permit to provide extra packages to install at preseed phase


hightoxicity:
In #39

PR #43, opened 2018-12-10
dhcp ha by dhcp dyn range sharding

hightoxicity:


hightoxicity:
pixiecore is currently using randomly generated secret to sign urls... It does not permit to run multiple instances able to check signed urls on same key basis.
cf.  https://github.com/google/netboot/blob/cc33920b4f3296801a64d731d269978116f40d92/pixiecore/booters.go#L137

PR #51, opened 2019-01-16
General updates to README and additional config comments (#1)

supertetelman:
* Fix DGX compute  nodes link

* Add specific error message for ceph install issue

* Move firmware upgrade to other section of README

* Add kibana link

* Add TODO comments to dgxie.yml

* Create ~/.kube dir

* Add links to/from services & provision

* Add optional step to delete iso-loader

* remove extra space and make it clear that you need to provision & configure your dgx before continuing

* don't have them install the container runtime on DGX that already have it

* Remove slurm  mention from provisioning steps

### DIFF
--- a/scripts/ceph_poll.sh
+++ b/scripts/ceph_poll.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+while true; do
+    kubectl -n rook-ceph exec -ti rook-ceph-tools ceph status # Run once to print output
+    kubectl -n rook-ceph exec -ti rook-ceph-tools ceph status | grep "mds: cephfs" | grep "up:active" | grep "standby-replay" # Run again to check for completion
+    if [ "${?}" == "0" ]; then
+    	echo "Ceph has completed setup."
+        break
+    fi
+    sleep 15
+done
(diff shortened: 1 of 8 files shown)


scripts/ceph_poll.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+while true; do
+    kubectl -n rook-ceph exec -ti rook-ceph-tools ceph status # Run once to print output
```
ryanolson: Does this work?  Or do you need to provide the actual name of the pod?

```
ryan@login:/homes/ryan/deepops$ kubectl -n rook-ceph exec -ti rook-ceph-tools ceph status
Error from server (NotFound): pods "rook-ceph-tools" not found
```

```
ryan@login:/homes/ryan/deepops$ kubectl -n rook-ceph exec -ti rook-ceph-tools-5bc8b8f97-77bnl ceph status
  cluster:
    id:     5e174c88-f3d5-4550-b4b9-ac2c85a95ad2
    health: HEALTH_OK
  ...
```

supertetelman: Good catch. We had a little hiccup during QA/check-in. I just pushed the updated script that dynamically pulls the name.

supertetelman:
Pushing into dev/training branch instead of master.

PR #63, opened 2019-01-31
Training

nvhans:
add execution bit to all scripts in deepops/scripts directory


dholt:
lets do this in the dev, this looks like it's pulling in a bunch of unwanted stuff from the training branch

PR #71, opened 2019-02-13
Install slurm deb from github release

lukeyeager:
Build files: https://github.com/mknoxnv/ubuntu-slurm/compare/master...lukeyeager:docker

Build artifacts: https://github.com/lukeyeager/ubuntu-slurm/releases


dholt:
Looks good, I'd like to get the builds into this repo if possible. I'm also unsure if we can distribute debs due to licensing but this is a step in the right direction

PR #75, opened 2019-02-14
Fix driver package name

lukeyeager:
The PPA package names are "nvidia-XXX".
https://launchpad.net/~graphics-drivers/+archive/ubuntu/ppa


dholt:
Did you test this? I thought when I did it didn't work

lukeyeager:
Which part didn't work? The package installed just fine with the standard ppa name given under `config.example/`.

And nvidia-smi works on the machine I just used it on. But I haven't tested other systems.

dholt:
Good enough, I'll tag you in any bug reports ;)

PR #80, opened 2019-02-14
Revive slurm

lukeyeager:


playbooks/slurm.yml
```diff
@@ -21,10 +21,14 @@
 - hosts: slurm-masters
   become: yes
   roles:
-    - { role: slurm, is_controller: yes }
+    - role: slurm
+      user: "{{ ansible_user }}"
```
lukeyeager: @dholt is this a dumb thing to do? If I don't make this change then I get errors in the slurm role about "user" not being defined. Not sure if this is the right value for that variable.

dholt: shouldn't hurt but we probably need to fix the role

PR #82, opened 2019-02-15
Slurm prolog+epilog changes

lukeyeager:
* Fix lint (using `shellcheck`)
* Stop hardcoding power levels


dholt:
We're going to have to re-visit the Max-P/Max-Q settings, Max-Q shouldn't be minimum power levels, but it will depend on GPU. I don't think we're using this anywhere yet however.

PR #88, opened 2019-02-19
nvidia-driver: updates

lukeyeager:
* Skip the slow driver reinstall if it looks like it's already installed
  * CLI override to reinstall
    `ansible-playbook playbooks/nvidia-driver.yml -e '{nvidia_driver_force_reinstall: yes}'`
* Only reboot if we actually changed something
  * CLI override to skip reboot
    `ansible-playbook playbooks/nvidia-driver.yml -e '{nvidia_driver_skip_reboot: yes}'`

### DIFF
--- a/playbooks/nvidia-driver.yml
+++ b/playbooks/nvidia-driver.yml
@@ -10,10 +10,6 @@
         name: nvidia-driver
       when: has_gpu.rc == 0
 
-    - include_role:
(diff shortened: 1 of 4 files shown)


dholt:
I like this but I can't avoid a reboot now like I could before when it was only in the playbook, can we have a flag to skip a reboot too?

PR #90, opened 2019-02-21
reboot - use ansible task instead of custom role

lukeyeager:


roles/slurm/tasks/compute.yml
```diff
@@ -22,9 +22,8 @@
 - name: update grub
   block:
     - command: update-grub
-    - file:
-        path: /var/run/reboot-required
-        state: touch
+    - name: reboot after updating grub config
+      reboot:
```
lukeyeager: @dholt was there an advantage to waiting to reboot? I just tried this and it seems to work fine. Better, in fact, since I didn't need to `scontrol update state=idle` for some reason.

dholt: i didn't write this one, but seems fine with me

ryanolson: I wrote it.  I don't think `reboot` was an option in ansible at the time.  Or if it was, it didn't work great.

PR #94, opened 2019-02-22
Refactor virtual

michael-balint:
Runs (almost) all deepops steps in a virtual environment. Now works with `refactor` branch mods.

@lukeyeager - Can you look at roles/nvidia-driver/tasks/main.yml? Had to force-overwrite for this to work in the base Ubuntu OS in the VM.

@dholt - Can you look at virtual/cluster_up.sh, lines 59-79? The documentation refers to older ansible playbooks that I think we're moving away from, so want to make sure I'm not missing anything.


roles/nvidia-driver/tasks/main.yml
```diff
@@ -27,6 +27,7 @@
   apt:
     name: "{{ item }}"
     install_recommends: no
+    dpkg_options: "force-overwrite"
```
lukeyeager: Yuck. What does Ubuntu OS come preinstalled with?

lukeyeager: Can we just ensure that a specific package is absent before running this command?

michael-balint: Kubespray installs `18.06.1~ce~3-0~ubuntu` in the `container-engine/docker : ensure docker packages are installed` task. When installing nvidia-docker, it complains that it needs `5:18.09.2~3-0~ubuntu-bionic`.

lukeyeager: Which package are you referring to? docker-ce? Wouldn't the change need to go into either the docker or nvidia-docker role instead of this one, then? Sorry, I think I'm missing something.

michael-balint: I'm going back and replaying it. I vaguely remember so *.so which could not get overwritten. There is also the conflict between the versions of docker-ce, which I alluded to above. Perhaps the latter is what is causing the former.

virtual/cluster_up.sh
```diff
@@ -1,37 +1,96 @@
...
 # Start vagrant via libvirt - set up the VMs
 vagrant up --no-parallel --provider=libvirt
 
+# Show the running VMs
+virsh list
+
 cd ..
 
+# Create the config for deepops servers (and use the virtual inventory)
+cp -r config.example/ config/
```
lukeyeager: It would be great if we could maintain two inventories at the same time (e.g. `./hosts.yml` and `./virtual/hosts.yml`). Then, we could use deepops for multiple clusters at the same time.

I think I nice way to support that would be to collapse all the variables in `./config/group_vars/*` into a single, YAML, config file.

But I also think we could do this even now by changing this line to `cp -r config.example/ virtual/config/`, and use `-i virtual/config/inventory` in the rest of the ansible commands in this file.

What do you think?

virtual/cluster_up.sh
```diff
@@ -1,37 +1,96 @@
...
 # Launch the apt repo service on k8s
-KUBECONFIG=$KUBECONFIG ./virtual/kubectl apply -f services/apt.yml
+./virtual/kubectl apply -f services/apt.yml
+
+# Launch the container registry
+helm repo add stable https://kubernetes-charts.storage.googleapis.com
+helm install --values config/registry.yml stable/docker-registry --version 1.4.3
+
+# Install nvidia drivers on dgx-servers
+ansible-playbook -l dgx-servers -e "ansible_user=vagrant" playbooks/nvidia-driver.yml
```
lukeyeager: I think it makes more sense to use the group `gpu-servers` for your machines instead of `dgx-servers`. Since your VM is obviously not actually a DGX.

virtual/cluster_up.sh
```diff
@@ -1,37 +1,96 @@
...
+
+# Downgrade docker-ce on dgx-servers
+ansible dgx-servers -e "ansible_user=vagrant" -b -a "apt remove -y docker-ce"
+# Install docker-ce compatible with nvidia-docker2
+ansible dgx-servers -e "ansible_user=vagrant" -b -a "apt install docker-ce=5:18.09.2~3-0~ubuntu-bionic"
```
lukeyeager: This workaround shouldn't be necessary anymore. Just use the standard playbooks.

See https://github.com/NVIDIA/nvidia-docker/issues/913#issuecomment-463524739

virtual/cluster_up.sh
michael-balint: I was thinking of that as well, actually. Totally agree.

virtual/cluster_up.sh
michael-balint: It's a DGX Station, but that's fair since we're just using one GPU from it. I cant change it.

virtual/cluster_up.sh
michael-balint: Thank you!

roles/nvidia-driver/tasks/main.yml
michael-balint: Removed this since it _should_ work with the nvidia-docker changes.

michael-balint:
> I haven't actually tested the code. Just read through it.
> 
> @michael-balint would it help for me to actually try this out?

It'd be helpful to try it out on a different system as a sanity check.

virtual/cluster_up.sh
michael-balint: @lukeyeager - The `nvidia-driver.yml` playbook hangs for me after installing the drivers and waiting to reboot. When I ssh to the gpu node, it appears to still be up. Any ideas on how to resolve this?

ryanolson: This doesn't have to do with swapping the `reboot` role for the `reboot` ansible module, does it?

michael-balint: Yes, it likely does. The reboot was working previously.

michael-balint: Weird, tracked this down and apparently has to do with authentication once the server/VM boots back up. Adding `ansible_password` back into the command fixes the issue.

PR #98, opened 2019-02-25
config - clean out unused variables

lukeyeager:
Why?

1. The number of configuration options under `./config/group_vars/` is, IMHO, a bit overwhelming at the moment. I'd like to keep it lean so users are more likely to actually read through all the options.
2. I'm not a fan of moving code to places like `roles/old/`. We've got the code in git history - if we want it back it's easy to resurrect. I'd like to mostly limit the checked-in code to what is currently working/quasi-supported.

Let me know if any of these old roles or vars are actually still useful. If so, let's resurrect them out of `old` and get them up and running again.


config.example/group_vars/all.yml
```diff
@@ -3,13 +3,6 @@
 
 # Account to use for SSH connections
 ansible_user: ubuntu
- 
-# SSH Configuration
-ssh_password_authentication: 'yes'
```
dholt: I use this var on nearly every install

config.example/group_vars/all.yml
```diff
@@ -3,13 +3,6 @@
...
-
-# Configure SUDO to not require a password for users in the `sudo` group
-# If this is `false`, pass the `-K` flag to all Ansible commands to prompt for the SUDO password
-enable_sudo_without_password: true
```
dholt: I think you just added this, and enabling sudo without a password is something that's done by default, so if you want it to be an option then we'll need this one.

config.example/group_vars/all.yml
```diff
@@ -3,13 +3,6 @@
 
 # Account to use for SSH connections
 ansible_user: ubuntu
```
dholt: `ansible_user` needs to go away. For SSH connections, you can configure your SSH client, or pass `-u <user>` to `ansible-playbook`

config.example/group_vars/all.yml
ryanolson: It’s usually done in the bootstrap playbook

lukeyeager:
I'm able to use the `refactor` branch to get a basic slurm cluster running. I'm still figuring out what is the minimum required product. It sounds like the `refactor` branch is currently useless, since all of these necessary roles are under `roles/old/`. I'll close this and work on resurrecting those roles one-by-one, instead. Once it's useful, we can promote it to either `master` or `dev`.

PR #104, opened 2019-02-28
nfs: revive and make minor changes

lukeyeager:


config.example/inventory
```diff
@@ -63,6 +63,15 @@ slurm-masters
 dgx-servers
 gpu-servers
 
+######
+# NFS
+######
+[nfs-server:children]
+login
```
lukeyeager: @dholt I can't decide if maybe `management` is a better default here or not? For now, I'm matching what you had before.

dholt: Not the way we used to do things, the management servers ran k8s and the login server ran nfs to export home directories. We could add a note about making management a child of `[nfs-server]` but I don't think we want to default to management nodes becoming NFS servers.

lukeyeager:
> I'm not sure I like leaving this uncommented by default

Fixed.

PR #105, opened 2019-02-28
RHEL/CentOS driver

dholt:
Update Nvidia driver role and playbook to support RHEL/CentOS


dholt:
Just for the playbook or the whole thing? The role might be more difficult since they share some parts

lukeyeager:
Here's what I was picturing:
* Playbook is still just `include_role`
* `tasks/main.yml` switches out to `tasks/{ubuntu,redhat}.yml` for distro-specific stuff
* After the distro-specific stuff, `tasks/main.yml` continues with the shared stuff. Maybe look into `package` instead of `apt`/`yum` for the tasks?

dholt:
The only trouble is that installing the driver on RHEL variants pulls in the latest kernel-headers package, but not the kernel, so it has to be resolved. I did it the way I did so that you wouldn't always get a kernel update if you didn't need it. I think that makes it hard to split out distro-specific stuff because you'd have to do it more than once. If we can keep the same behavior I'm fine with however it's done.

lukeyeager:
@dholt I took the liberty of updating this branch. Highlights from an ansible log with two compute nodes - one ubuntu and the other centos:
```
...

TASK [nvidia-driver : ubuntu tasks] *********************************************************************************************************************************************************************************************************************************************
skipping: [lyeager-compute-centos-01]                                                                                                                                                                                                                                            
included: /home/lyeager/code/deepops/roles/nvidia-driver/tasks/ubuntu.yml for lyeager-compute-01                                                                                                                                                                                 

...

TASK [nvidia-driver : redhat family tasks] **************************************************************************************************************************************************************************************************************************************
skipping: [lyeager-compute-01]                                                                                                                                                                                                                                                   
included: /home/lyeager/code/deepops/roles/nvidia-driver/tasks/redhat.yml for lyeager-compute-centos-01                                                                                                                                                                          

...

TASK [test nvidia-smi] **********************************************************************************************************************************************************************************************************************************************************
skipping: [lyeager-mgmt]                                                                                                                                                                                                                                                         
skipping: [lyeager-login]                                                                                                                                                                                                                                                        
ok: [lyeager-compute-centos-01]                                                                                                                                                                                                                                                  
ok: [lyeager-compute-01]                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                 
PLAY RECAP **********************************************************************************************************************************************************************************************************************************************************************
lyeager-compute-01         : ok=14   changed=6    unreachable=0    failed=0                                                                                                                                                                                                      
lyeager-compute-centos-01  : ok=20   changed=12   unreachable=0    failed=0                                                                                                                                                                                                      
lyeager-login              : ok=3    changed=0    unreachable=0    failed=0                                                                                                                                                                                                      
lyeager-mgmt               : ok=3    changed=0    unreachable=0    failed=0                                                                                                                                                                                                      
```

playbooks/nvidia-driver.yml
```diff
@@ -1,6 +1,15 @@
+---
 - hosts: all
   become: true
   tasks:
+    - name: run setup to detect OS
+      setup:
+
+    - name: install lspci
+      package:
```
dholt: I think you need an `update_cache`, when I installed CentOS 7 via MAAS it hadn't been updated.

dholt:
It looks to me like this is going to run into the problem I mentioned. Kernel version is going to match kernel header version, then the driver gets installed and updates the kernel headers but not the kernel, and the driver is built for the wrong kernel and doesn't work. The hacky kernel stuff needs to go after the driver install screws up the packages.

playbooks/nvidia-driver.yml
lukeyeager: Ah. Ok. My test setup ran it at some point, I guess.

lukeyeager: Fixed. I couldn't decide whether I wanted to have two `update_cache` blocks and a `package` block or a `yum` and an `apt` 🤷‍♂️

dholt:
Tested on Ubuntu and CentOS, looks good to me

PR #107, opened 2019-02-28
Move all SSH configuration into inventory

lukeyeager:
This logical grouping makes sense to me. The inventory tells you where your machines are, and how to connect to them. Then, all the stuff under `./config/group_vars/` is for configuring the playbooks+roles.

Then, if the user wants just "the default stuff", they should only ever need to edit the inventory.

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -8,9 +8,6 @@ ssh_password_authentication: 'yes'
 # If this is `false`, pass the `-K` flag to all Ansible commands to prompt for the SUDO password
 enable_sudo_without_password: true
 
-# Configure Ansible to use a proxy if remote machines are unroutable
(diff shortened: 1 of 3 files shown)


lukeyeager:
If you don't set `ansible_user`, then it defaults to the username on your provisioning machine, right? I don't think it's fair to assume that those will match.

I agree it'd be nice to remove the reference to `ansible_user` elsewhere in the repo, though. We should make each role require you to be more specific about which user should be used.

dholt:
Correct, but that's because that's the user Ansible uses for SSH. You should either set up SSH, set `ansible_user` (though we should not depend on that, or pass `-u <user>` to `ansible-playbook`. It's OK to use this var in places, but I don't want to rely on making the user set it. If we default it to `ubuntu` for example, people will have to go and change the config for CentOS. If we leave it out that's not the case, and it's on the user to connect to the system properly, which will set `ansible_user`.

PR #108, opened 2019-02-28
Minor fixes to virtual cluster turn-up after walkthrough

ajdecon:
I walked through setting up a new virtual cluster, on a fresh Ubuntu 18.04 release. There were a few minor issues that required fixes:

* Default user was not in `libvirt` group after running `bootstrap_virtual.sh`. Added a note to README.md to recommend starting a fresh login session.
* Fixed setting `ansible_user` to `vagrant` in `cluster_up.sh`.
* `cluster_up.sh` was breaking because it couldn't find `kubectl`, added a line to include ` virtual/k8s-config/artifacts/kubectl` in the PATH.
* Fixed locations of a couple config files for the monitoring setup script.


michael-balint:
@ajdecon - Thanks for testing/walking through this! Did you also test with GPU passthrough?

ajdecon:
@michael-balint : Not yet, was just about to get started on that.

config.example/group_vars/all.yml
```diff
@@ -8,6 +8,9 @@ ssh_password_authentication: 'yes'
 # If this is `false`, pass the `-K` flag to all Ansible commands to prompt for the SUDO password
 enable_sudo_without_password: true
 
+# Configure the default ansible user
+#ansible_user: ubuntu
```
lukeyeager: See https://github.com/NVIDIA/deepops/pull/107

scripts/k8s_deploy_monitoring.sh
```diff
@@ -26,7 +26,7 @@ fi
 # Deploy Monitoring stack
 helm status kube-prometheus >/dev/null 2>&1
 if [ $? -ne 0 ] ; then
-    helm install coreos/kube-prometheus --name kube-prometheus --namespace monitoring --values config.example/kube-prometheus.yml
+    helm install coreos/kube-prometheus --name kube-prometheus --namespace monitoring --values config.example/helm/kube-prometheus.yml
```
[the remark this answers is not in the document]
dholt: It doesn't technically matter, but since `virtual/cluster_up.sh` is doing `cp -r config.example/ virtual/config/`, we should point at `config` instead of `config.example` since the former may have been modified.

config.example/group_vars/all.yml
ajdecon: Ack. I'll rebase this branch after #107 lands and fix the `sed` command to refer to inventory.

lukeyeager: Done

scripts/k8s_deploy_monitoring.sh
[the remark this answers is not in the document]
michael-balint: I did that because I was trying to mimic what regular (non-virtual) users would do - cp the config.example to config. Is that still the case?

This is modifying a file outside of virtual, which is why I ask.

config.example/group_vars/all.yml
michael-balint: @ajdecon - I don't think we need it. Should take out the sed command since all ansible-playbook commands explicitly have `-e "ansible_user=vagrant"` set. Should retest without the sed modification.

scripts/k8s_deploy_monitoring.sh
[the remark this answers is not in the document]
dholt: Oh sorry, didn't notice that. The reason I pointed to `config.example` instead of `config` was that in the K8s cluster deployment you don't need anything in `config`. It should probably check for `config` and use that if it exists, `config.example` otherwise.

config.example/group_vars/all.yml
ajdecon: Confirmed that simply removing the `sed` command works fine, went ahead and made that change. Removed addition to `config.example/group_vars/all.yml`.

PR #111, opened 2019-03-01
New user and ssh roles

dholt:


playbooks/user-password.yml
```diff
@@ -0,0 +1,33 @@
...
+        ssh_server_password_login: true
+        ssh_use_pam: true
+        ssh_max_auth_retries: 10
+    - name: Set user password
+      include_role:
+        name: users
+      vars:
+        # User Configuration
+        users:
+          - username: "{{ ansible_user }}"
```
lukeyeager: I thought we were trying to get rid of this?

playbooks/user-password.yml
```diff
@@ -0,0 +1,33 @@
...
+# Generate password for user
+# You can also generate a password with `mkpasswd -m sha-512`
+- hosts: all
+  become: true
+  gather_facts: false
+  vars:
+    pwgen: "{{ lookup('password', '/dev/null length=8 chars=ascii_letters') }}"
+  tasks:
+    - set_fact:
+        passwd: "{{ pwgen }}"
```
lukeyeager: Maybe I don't understand the value of facts yet, but why not simply merge all this into the users section below?
```
users:
  password: "lookup('password', '/dev/null length=8 chars=ascii_letters') | password_hash('sha512', 'salty') }}"
```
Also, I think you can get some free extra security by using e.g. the hostname as the salt instead of something static.

playbooks/user-password.yml
```diff
@@ -0,0 +1,33 @@
...
+      include_role:
+        name: dev-sec.ssh-hardening
+      vars:
+        ssh_client_hardening: false
+        ssh_server_password_login: true
+        ssh_use_pam: true
+        ssh_max_auth_retries: 10
+    - name: Set user password
+      include_role:
+        name: users
```
lukeyeager: Do you want `unxnn.ansible_users` here?

playbooks/user-password.yml
dholt: I want to get rid of SETTING it, it's ok to read it. I don't want to require people to change config files and set that variable. Ansible sets it to the SSH user so it's OK to read.

playbooks/user-password.yml
dholt: I can't print the password that way. Really we should be using the vault

playbooks/user-password.yml
dholt: fixed

michael-balint:
There's one thing failing for me, but it's not a result of these changes.

`scripts/k8s_deploy_monitoring.sh` refers to `config.example/kube-prometheus*` files which don't exist

dholt:
They were moved to the `helm` dir, which is fixed in another PR. I'll rebase after this is merged.

PR #115, opened 2019-03-01
Setup script: rename, reduce printing, add config dir copy

dholt:


scripts/setup.sh
```diff
@@ -0,0 +1,97 @@
...
+if [ $? -eq 0 ] ; then
+    git submodule update --init
+else
+    echo "ERROR: Unable to update Git submodules"
+fi
+
+# Copy default configuration
+CONFIG_DIR=${CONFIG_DIR:-./config}
+if [ ! -d "${CONFIG_DIR}" ] ; then
+    cp -rfp ./config.example "${CONFIG_DIR}"
```
michael-balint: Not blocking, but just want to point out - we are creating a config directory which we will never use for the k8s cluster script. Should we think about how to converge a generated k8s-config w/ the config that is cp-ed?

scripts/setup.sh
```diff
@@ -0,0 +1,98 @@
...
+
+        # Install Ansible
+        type ansible >/dev/null 2>&1
+        if [ $? -ne 0 ] ; then
+            echo "Installing Ansible..."
+            sudo apt-add-repository -y ppa:ansible/ansible >/dev/null
+            sudo apt-get update >/dev/null
+            sudo apt-get -y install ansible >/dev/null
+        fi
+        ansible --version | head -1
```
michael-balint: Is there a way we can force an upgrade if ansible is not at the right version? Or better yet, freeze the version? It seems like lots of issues have arisen out of users having older versions and/or a specific bug in a particular version.

lukeyeager: I agree this is a problem. It's not clear how we should fix it, though.

PR #119, opened 2019-03-01
Purge ansible_user

dholt:
Don't rely on `ansible_user` being set. This is the case when playbooks are run against the local host and there's no SSH connection.

### DIFF
--- a/playbooks/bootstrap.yml
+++ b/playbooks/bootstrap.yml
@@ -21,12 +21,12 @@
 - hosts: all
   become: true
   tasks:
-    - name: Allow password-less sudo to "{{ ansible_user }}"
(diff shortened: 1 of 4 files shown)


playbooks/bootstrap.yml
```diff
@@ -21,12 +21,12 @@
 - hosts: all
   become: true
   tasks:
-    - name: Allow password-less sudo to "{{ ansible_user }}"
+    - name: Allow password-less sudo to ansible user"
```
lukeyeager: typo: trailing quote

lukeyeager:
(sorry, meant to hit approve the first time)

PR #122, opened 2019-03-02
Update kube.yml

ryanolson:
adding stub for `docker_insecure_registries` in kube.yml

We should document the [Kubernetes Configuration Guide](https://github.com/NVIDIA/deepops/blob/refactor/docs/kubernetes-cluster.md) the how to install a local docker registry.  This document still appears to be a WIP on the refactor branch.

### DIFF
--- a/config.example/kube.yml
+++ b/config.example/kube.yml
@@ -16,7 +16,11 @@ kube_apiserver_admission_control:
 nvidia_accelerator_enabled: true
 nvidia_gpu_flavor: tesla
 # nvidia_driver_version: "384.111"
+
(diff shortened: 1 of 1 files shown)


dholt:
The Kubernetes guide is not intended to replace the old readme, which is now the DGX POD doc. We don't use this file to install Kubernetes in the refactor branch, so we need a new component and docs to set up the registry.

ryanolson:
FYI - `docker_insecure_registries` is a kubespray configurations variable. 

How do we pass options to kubespray on the refactor branch?

PR #124, opened 2019-03-02
Install container registry Helm chart with Ansible

dholt:


dholt:
The Helm module requires `host` to point to the tiller host. We're current running Tiller in k8s with a service type ClusterIP, so we may have to expose the port to the local host or find an access method that doesn't change, since `localhost` does not work.

ryanolson:
I added the ability to increase the size of message proxied to allow for large layers to pass thru the proxy.

ryanolson:
Since this is a docker role, we should also run two further roles as a dependency:
- a fixed hostname role which adds `container_registry_hostname` to /etc/hosts pointing to the k8s ingress service
- a docker role, preferably the kubespray docker role with the `docker_insecure_registries` marked to allow `container_registry_hostname`

dholt:
Can you do a new PR for that? You shouldn't be committing code to mine and able to review :)

dholt:
@ryanolson that was easier than I originally thought so I went ahead and made the changes.

By default, `/etc/hosts` is populated with the names of all hosts in the cluster, so if we default `container_registry_hostname` to `node1.cluster.local`, it's resolvable in the entire cluster without any config changes required.

This also solves your second suggestion; setting `docker_insecure_registries` to `node1.cluster.local` by default requires no config changes and no other playbooks/roles to be run.

PR #126, opened 2019-03-04
added proxy-body-size parameter set to 0

mfoco:
see https://github.com/kubernetes/contrib/issues/2444


dholt:
I think we'll want to defer this due to #124

mfoco:
Oh, I see.
Feel free to reject and move the line to those new configs.

PR #129, opened 2019-03-04
virtual - add prefix to hostnames

lukeyeager:
* Less chance of collision with real, external hosts
* Allows us to rename "login01" to "login" without colliding with the
    group name

### DIFF
--- a/virtual/README.md
+++ b/virtual/README.md
@@ -37,7 +37,7 @@ To start the cluster, run the `cluster_up.sh` script...
 $ ./cluster_up.sh
 ```
 
-Vagrant will spin up three VMs - login01, mgmt, and dgx01. Afterwards, the script will use ansible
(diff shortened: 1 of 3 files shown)


lukeyeager:
It seems to me you'd probably only want a single login node. Whereas you might actually want multiple mgmt nodes. But @dholt probably has a better informed opinion on this.

dholt:
I agree in most cases you'd have one login, and one or more management nodes. The `login01`/`mgmt01` names were just placeholders that would hopefully be overwritten with real hostnames on deployment, so I have no preference what we call them. Probably should be consistent everywhere to avoid confusion.

lukeyeager:
Agree on consistency. I went out of my way to be consistent in this PR. The vagrant boxes don't really need the prefix, but I kept it just to minimize confusion. I'm avoiding `vagrant ssh mgmt`/`ssh virtual-mgmt`.

PR #130, opened 2019-03-04
virtual - move SSH connection params to inventory

lukeyeager:
Does this seem cleaner to y'all? It does to me. Super minor nitpick.


lukeyeager:
> we need to change the generated k8s_hosts.ini as well

Done.

PR #143, opened 2019-03-06
Refactor bootstrap playbooks

lukeyeager:
* Renamed python playbook
* Renamed ssh playbook
* Split out sudo stuff into new playbook + extended it
* Added debug messages


playbooks/bootstrap-python.yml
```diff
@@ -0,0 +1,15 @@
...
+  gather_facts: False
+  tasks:
+    # Fortunately, python is a dependency of yum. So RedHat systems always have it.
+    - name: install python
+      raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
+      register: output
+      changed_when: output.stdout != ""
+
+    - debug:
+        msg: Now you shouldn't need to use the 'raw' module for these hosts anymore.
```
dholt: Not entirely sure we need this ;)

playbooks/bootstrap-sudo.yml
```diff
@@ -0,0 +1,35 @@
...
+      when: sudo.rc != 0
+
+    - name: install sudo package
+      package: name=sudo
+      become: true
+      become_method: "{{ method }}"
+
+    - name: "grant password-less sudo to user \"{{ ansible_user }}\""
+      copy:
+        content: "{{ ansible_user }} ALL=(ALL) NOPASSWD:ALL"
```
dholt: I think this var needs to be `{{ ansible_user | default(ansible_env.SUDO_USER) }}`

playbooks/bootstrap-ssh.yml
```diff
@@ -2,31 +2,27 @@
 - hosts: 127.0.0.1
   connection: local
   gather_facts: false
-  vars:
-    ansible_become: false
```
dholt: If I run a playbook with `become`, i don't think this will have the intended effect any longer. This playbook gets included in k8s-cluster which is run with `become`.

It might have the same outcome if we set permissions on the generated files/dirs to be owned by the non-root user.

playbooks/bootstrap-ssh.yml
```diff
@@ -2,31 +2,27 @@
...
         path: "{{ lookup('env','HOME') + '/.ssh' }}"
         state: directory
-    # Requires Ansible v2.8
-#    - name: Generate default SSH keypair if it doesn't exist
-#      openssh_keypair:
-#        path: "{{ lookup('env','HOME') + '/.ssh/id_rsa' }}"
+
     - name: Generate default SSH keypair if it doesn't exist
-      command: ssh-keygen -q -t rsa -f "{{ lookup('env','HOME') + '/.ssh/id_rsa' }}" -C "" -N ""
+      command: ssh-keygen -q -t rsa -f "{{ lookup('env','HOME') + '/.ssh/id_rsa_ansible' }}" -C "" -N ""
```
dholt: I don't really like this, now I have to specify the SSH key every time I access the server

playbooks/bootstrap-sudo.yml
lukeyeager: I saw you keep using that, but it never seemed necessary for me. Can you give me a repro case where that's useful?

playbooks/bootstrap-python.yml
lukeyeager: Haha true. I got a little excited about my debug messages :)

playbooks/bootstrap-ssh.yml
lukeyeager: Ok, I can set it back to `id_rsa`. I don't have one of those on my system, and my SSH config isn't configured to use it. So, for me, I was planning to add a line to my inventory file if I ever needed to use this playbook. Seemed like I might as well be explicit about the purpose of the key.

But you're probably right that many people have `IdentityFile ~/.ssh/id_rsa` in their ssh configs. Which would make this playbook "just work".

dholt: The default SSH client on any system I've ever used has looked for that file by default, which is why I used it here.

I like this in theory but it just adds extra steps most of the time I think. Maybe we could make it a var and set `id_rsa` as the default?

playbooks/bootstrap-sudo.yml
dholt: If I pass `-b` to `ansible-playbook`, `{{ ansible_user }}` is set to `root`. In this case it would grant password-less sudo to `root`.

lukeyeager: I want to believe you, but I just can't find a way to replicate that.
```
$ ansible localhost -m debug -a 'var=ansible_user'
localhost | SUCCESS => {
    "ansible_user": "vagrant"
}
$ ansible localhost -b -m debug -a 'var=ansible_user'
localhost | SUCCESS => {
    "ansible_user": "vagrant"
}
```
(tried with other hosts besides localhost, too)

dholt: You are defining `ansible_user`, so the variable exists. If I have the same user on the control and target machines, for example, meaning I don't need to set that variable by hand, then it does not work:

```$ ansible lab-gk8-worker1 -m debug -a 'var=ansible_user'
lab-gk8-worker1 | SUCCESS => {
    "ansible_user": "VARIABLE IS NOT DEFINED!"
}
```

I don't remember the case where it comes back as `root`, the conversation is buried in slack/github somewhere. Maybe it was a different situation. Regardless, my suggestion always returns the correct user without requiring someone to set `{{ ansible_user }}`.

dholt:
I'm getting an error with `bootstrap-ssh.yml`, it's skipping the local host:

```
$ ansible-playbook -l lab-gk8-worker4 -k playbooks/bootstrap-ssh.yml
SSH password:

PLAY [127.0.0.1:all] *******************************************************************************************************************************************************************************************

TASK [set SSH key location] ************************************************************************************************************************************************************************************
ok: [lab-gk8-worker4]

PLAY [127.0.0.1] ***********************************************************************************************************************************************************************************************
skipping: no hosts matched

PLAY [all] *****************************************************************************************************************************************************************************************************

TASK [Add SSH public key to ansible user user authorized keys] *************************************************************************************************************************************************
 [WARNING]: Unable to find '/home/dholt/.ssh/id_rsa.pub' in expected paths (use -vvvvv to see paths)

fatal: [lab-gk8-worker4]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'file'. Error was a <class 'ansible.errors.AnsibleError'>, original message: could not locate file in lookup: /home/dholt/.ssh/id_rsa.pub"}
        to retry, use: --limit @/local/dholt/deepops/playbooks/bootstrap-ssh.retry

PLAY RECAP *****************************************************************************************************************************************************************************************************
lab-gk8-worker4            : ok=1    changed=0    unreachable=0    failed=1

```

playbooks/bootstrap-sudo.yml
lukeyeager: That was it, good call. I was setting the user in the inventory.

However, your solution is still not good. Check this out:

**ansible_env depends on whether `setup` was run with --become or not**
```bash
# run setup as root
$ ansible virtual-login -b -m setup

# ... and SUDO_USER is SET with or without --become
$ ansible virtual-login -m debug -a "msg='<{{ansible_env.USER|default(\"\")}}> <{{ansible_env.SUDO_USER|default(\"\")}}>'"
virtual-login | SUCCESS => {
    "msg": "<root> <vagrant>"
}
$ ansible virtual-login -b -m debug -a "msg='<{{ansible_env.USER|default(\"\")}}> <{{ansible_env.SUDO_USER|default(\"\")}}>'"
virtual-login | SUCCESS => {
    "msg": "<root> <vagrant>"
}

# run setup without root
$ ansible virtual-login -m setup

# ... and SUDO_USER is UNSET with or without --become
$ ansible virtual-login -m debug -a "msg='<{{ansible_env.USER|default(\"\")}}> <{{ansible_env.SUDO_USER|default(\"\")}}>'"
virtual-login | SUCCESS => {
    "msg": "<vagrant> <>"
}
$ ansible virtual-login -b -m debug -a "msg='<{{ansible_env.USER|default(\"\")}}> <{{ansible_env.SUDO_USER|default(\"\")}}>'"
virtual-login | SUCCESS => {
    "msg": "<vagrant> <>"
}

```
**`shell`, for reference, doesn't depend on `setup`**
```bash
$ ansible virtual-login -m shell -a 'echo "<$USER> <$SUDO_USER>"'
virtual-login | CHANGED | rc=0 >>
<vagrant> <>

$ ansible virtual-login -b -m shell -a 'echo "<$USER> <$SUDO_USER>"'
virtual-login | CHANGED | rc=0 >>
<root> <vagrant>
```

So ...

1. We can't rely on `ansible_user` because it might not be set
2. We can't rely on `SUDO_USER` because it's only set if `setup` was run as root
3. We can't really rely on `ansible_env` at all because we have fact caching turned on

I think this is the only safe thing to do:
```yaml
    # re-run setup, to guarantee the cache is good
    - setup: gather_subset=min
    # use SUDO_USER if set, otherwise USER
    - debug: msg={{ansible_env.SUDO_USER|default(ansible_env.USER)}}
```

playbooks/bootstrap-ssh.yml
```diff
@@ -0,0 +1,35 @@
+---
+- hosts: 127.0.0.1:all
+  gather_facts: false
+  tasks:
+    - name: set SSH key location
+      set_fact:
+        private_key: "{{ ansible_ssh_private_key_file | default(lookup('env','HOME') + '/.ssh/id_rsa') }}"
+
+- hosts: 127.0.0.1
```
dholt: This might be something wonky with my machine since this used to work, but I'm getting an error with `bootstrap-ssh.yml`, it's skipping the local host:

```
$ ansible-playbook -l lab-gk8-worker4 -k playbooks/bootstrap-ssh.yml
SSH password:

PLAY [127.0.0.1:all] *******************************************************************************************************************************************************************************************

TASK [set SSH key location] ************************************************************************************************************************************************************************************
ok: [lab-gk8-worker4]

PLAY [127.0.0.1] ***********************************************************************************************************************************************************************************************
skipping: no hosts matched

PLAY [all] *****************************************************************************************************************************************************************************************************

TASK [Add SSH public key to ansible user user authorized keys] *************************************************************************************************************************************************
 [WARNING]: Unable to find '/home/dholt/.ssh/id_rsa.pub' in expected paths (use -vvvvv to see paths)

fatal: [lab-gk8-worker4]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'file'. Error was a <class 'ansible.errors.AnsibleError'>, original message: could not locate file in lookup: /home/dholt/.ssh/id_rsa.pub"}
        to retry, use: --limit @/local/dholt/deepops/playbooks/bootstrap-ssh.retry

PLAY RECAP *****************************************************************************************************************************************************************************************************
lab-gk8-worker4            : ok=1    changed=0    unreachable=0    failed=1

```

This solution seems to work:

```
$ git diff playbooks/bootstrap-ssh.yml
diff --git a/playbooks/bootstrap-ssh.yml b/playbooks/bootstrap-ssh.yml
index 1ed4caf..7400a6b 100644
--- a/playbooks/bootstrap-ssh.yml
+++ b/playbooks/bootstrap-ssh.yml
@@ -6,19 +6,20 @@
       set_fact:
         private_key: "{{ ansible_ssh_private_key_file | default(lookup('env','HOME') + '/.ssh/id_rsa') }}"

-- hosts: 127.0.0.1
-  connection: local
+- hosts: all
   gather_facts: false
   tasks:
     - name: Create SSH directory if it doesn't exist
       file:
         path: "{{ lookup('env','HOME') + '/.ssh' }}"
         state: directory
+      delegate_to: 127.0.0.1

     - name: Generate default SSH keypair if it doesn't exist
       command: ssh-keygen -q -t rsa -f "{{ private_key }}" -C "" -N ""
       args:
         creates: "{{ private_key }}"
+      delegate_to: 127.0.0.1

 - hosts: all
   become: true
```

lukeyeager: Good call. I was able to clean up the playbook to use a single block now.

PR #144, opened 2019-03-07
Adding pytorch-job test (runs mnist example).

michael-balint:
Simple pytorch job that is useful as an example.

### DIFF
--- a/tests/pytorch-job.yml
+++ b/tests/pytorch-job.yml
@@ -6,8 +6,6 @@ spec:
   backoffLimit: 5
   template:
     spec:
-      imagePullSecrets:
-        - name: nvcr.dgxkey
       containers:
         - name: pytorch-container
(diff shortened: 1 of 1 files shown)


tests/pytorch-job.yml
```diff
@@ -0,0 +1,17 @@
...
+  template:
+    spec:
+      containers:
+        - name: pytorch-container
+          image: nvcr.io/nvidia/pytorch:19.02-py3
+          command: ["/bin/sh"]
+          args: ["-c", "python /workspace/examples/upstream/mnist/main.py"]
+          resources:
+            limits:
+              nvidia.com/gpu: 2
```
dholt: I think it makes more sense to use one GPU by default so that it's more likely this will work

michael-balint: Agreed - done!

PR #146, opened 2019-03-07
kerberos/nis/autofs

lukeyeager:


roles/dns-config/tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+- name: assert that variables are defined
+  fail:
+    msg: "Variable '{{ item }}' is not defined"
+  when: item not in vars
+  with_items:
+    - dns_config_servers
```
dholt: This doesn't match the var in `roles/dns-config/tasks/main.yml`. It's also ambiguous whether we're configuring nameservers or search domains, I would think we want to have vars for both.

roles/move-home-dirs/tasks/move_users.yml
```diff
@@ -0,0 +1,53 @@
+# We cannot move the home directory of the currently logged in user
+# So, we create a temporary user and login with that
```
dholt: Awesome solution

roles/dns-config/tasks/main.yml
dholt: I see what you're doing now, which is the same with the KRB stuff. I'm not sure about this, because it requires a user to set config vs. having sane defaults out of the box. Can we just make no changes to the file if the variable is not set?

dholt: Can we leave the target file unmodified in this case instead of failing? It requires someone to edit configuration vs providing sane defaults.

roles/kerberos-client/tasks/main.yml
```diff
@@ -1,16 +1,20 @@
 ---
+- name: assert that variables are defined
+  fail:
+    msg: "Variable '{{ item }}' is not defined"
+  when: item not in vars
+  with_items:
+    - kerberos_client_realm_name
```
dholt: Same as other comment about DNS

roles/dns-config/tasks/main.yml
lukeyeager: I'd argue there are some scenarios where there are no sane defaults. `dns_config_servers` and `kerberos_client_realm_name` seem like two of those scenarios. `ls` has a sane default arg (`.`), but what's a sane default for `ping`?

This playbook would never be included in a generic `*-cluster.yml` playbook, anyway. I think it's OK to have this one fail loudly.

roles/move-home-dirs/tasks/move_users.yml
```diff
@@ -0,0 +1,53 @@
...
+- name: grant sudo to temporary user
+  copy:
+    content: "{{ tmp_user }} ALL=(ALL) NOPASSWD:ALL"
+    dest: "/etc/sudoers.d/{{ tmp_user }}"
+
+- name: move each user
+  include_tasks: move_user.yml
+  vars:
+    ansible_user: "{{ tmp_user }}"
+    ansible_password: "{{ tmp_user_password }}"
```
lukeyeager: @dholt this requires that `sshpass` is installed on the provisioning machine. That wasn't a requirement before. Do you want me to add more steps in here to do this with a public/private key pair instead?

dholt: That would be my preference if it avoids another dependency

lukeyeager: kk will do

lukeyeager: Done

PR #148, opened 2019-03-08
dgxie iso nfs wrapper

hightoxicity:
Offer an alternative deployment for dxie iso-loader.
Idea behind that (and experience) is that in a small cluster, if you aggregate your worker nodes storage in a hyperconvergent spirit and offer it to ceph as osds, fact is that your DHCP (dgxie) running into the kube cluster have a hard dependency on a ceph flex volume, and if for any reason (just a kube worker node reboot) your volume gets unavailable, you lose your dgxie pod, so you lose your DHCP and then you lose all your worker nodes, that is something we faced to many time with deepops, the solution we found to take over incident was to modify the dgxie helm chart bundle to mount an empty volume to make dgxie run again and we retrieve all worker nodes thanks to dhcp and then the ceph cluster gone back healthy.
Here the idea, is to use a NFS mount command from dgxie pod to offer the ISO, that way, dgxie become ceph fault tolerant. I tested it turning off all my worker nodes, my dgxie pod stayed up and I "just" lost the NFS iso wrapper pods, when I turned on the worker nodes, the dgxie container retrieved its NFS mount itself.
Maybe it looks like a tricky thing but we have a great robustness gain that I would like to share.

You can find the iso loader container build chain here: https://github.com/hightoxicity/docker-nfs-server/tree/feat-qwant if you want to replace by one you build.

Thx.

### DIFF
--- a/containers/dgxie/Dockerfile
+++ b/containers/dgxie/Dockerfile
@@ -18,6 +18,8 @@ RUN mkdir -p /www /var/run/vsftpd/empty
     #echo 'pasv_max_port=10100' >> /etc/vsftpd.conf && \
     #echo 'pasv_min_port=10090' >> /etc/vsftpd.conf
 
+RUN apt-get -y install nfs-common
+
 COPY get_hosts.py /usr/local/bin
 COPY rest_api.py /usr/local/bin
 COPY api.py /api.py
--- a/containers/dgxie/start
+++ b/containers/dgxie/start
@@ -2,6 +2,7 @@
 
 ## CONFIG
 MNT=${MNT:-/data}
+NFS_DIR=${NFS_DIR:-}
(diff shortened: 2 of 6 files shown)


michael-balint:
Thanks for this! Can you sign a copy of our CLA (Contributor License Agreement) so we can accept this and future PRs?

https://raw.githubusercontent.com/NVIDIA/deepops/master/CLA

michael-balint:
@hightoxicity - see my comment above?

michael-balint:
@hightoxicity - can you fix this merge conflict and we'll test/merge... thank you!

michael-balint:
Closing due to merge conflicts - please address these conflicts and we can reopen + accept + merge.

PR #150, opened 2019-03-08
New role to install cuDNN and NCCL libraries

dholt:


dholt:
Yes, containers usage is still not very widespread in a lot of places, like HPC. I was asked for this yesterday, which prompted the role.

PR #151, opened 2019-03-08
Add method to install extra software packages

dholt:


playbooks/software.yml
```diff
@@ -24,6 +24,12 @@
         state: present
         update_cache: yes
 
+    - name: install extra packages
+      apt:
+        name: "{{ extra_sw_packages }}"
```
lukeyeager: 1. Can we use `software_extra_packages` instead? (prefix var name with role name)
2. I'm pretty sure this will error out if you don't define the variable. Maybe put the empty list in `defaults/main.yml`?

dholt: We don't have a role for this one, just the playbook. Thinking we should go back to the role since the playbook will get longer when we add RHEL support

lukeyeager: Oh right! My bad, sloppy reading. We can go to a role now if you like. Or you can just add a `when` block to this task.

dholt: I don't know how to set a default without a role, if I set ```      vars:
        software_extra_packages: []``` in the playbook, it takes precedence over the variable in `config/group_vars/all.yml`

lukeyeager: I was thinking something like this:
```
    - name: install extra packages
      apt:
        name: "{{ software_extra_packages }}"
        state: present
        update_cache: yes
      when: software_extra_packages is defined
```

dholt: Added

PR #152, opened 2019-03-08
cachefilesd - revive old role

lukeyeager:


playbooks/nfs.yml
```diff
@@ -8,3 +8,4 @@
   become: yes
   roles:
     - { role: nfs, nfs_is_client: yes }
+    - cachefilesd
```
dholt: Can we not add this here? Cachefilesd can be really buggy and I rarely use it, but use NFS all the time.

lukeyeager: Yep, can do.

PR #161, opened 2019-03-13
Add NTP client role and playbook

dholt:


ryanolson:
👍  Nice.  geerlingguy does some good work.

PR #160, opened 2019-03-13
Add a simple Slurm + MPI hello world example

ajdecon:
Getting the examples section started with a very simple MPI example, demonstrated to work in a virtual Slurm DeepOps cluster.

While we will want to have some more impressive examples, e.g. Deep Learning training and all that jazz, we should also have some simple and clear examples that are easy for new users to build on and get running. This example is about as simple as you can get while still being an MPI, but also provides a Slurm usage example.

Next one I'm working on is an equivalent "hello, world" Flask service running in k8s, but running into fun with the registry which is delaying that. So might as well open a PR for this on its own. :)

PR #162, opened 2019-03-14
Add a simple Slurm + MPI Hello, World example

ajdecon:
(Take 2! This is a re-do after I messed up PR #160.)

Getting the examples section started with a very simple MPI example,
demonstrated to work in a virtual Slurm DeepOps cluster.

While we will want to have some more impressive examples, e.g. Deep
Learning training and all that jazz, we should also have some simple and
clear examples that are easy for new users to build on and get running.
This example is about as simple as you can get while still being an MPI,
but also provides a Slurm usage example.

Next one I'm working on is an equivalent "hello, world" Flask service
running in k8s, but running into fun with the registry which is delaying
that. So might as well open a PR for this on its own. :)

### DIFF
--- a/examples/slurm-mpi-hello/bootstrap-mpi.yml
+++ b/examples/slurm-mpi-hello/bootstrap-mpi.yml
@@ -1,0 +1,13 @@
+---
+# Install OpenMPI packages from Ubuntu repos
+- hosts: all
+  become: true
+  tasks:
+  - name: install openmpi packages
+    apt:
+      name: "{{ item }}"
+    with_items:
+    - openmpi-bin
+    - libopenmpi-dev
+    - libpmi2-pmix
+    - libpmi-pmix-dev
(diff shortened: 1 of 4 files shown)


ajdecon:
@dholt I believe this should include the fixed file paths you suggested in #160.

examples/slurm-mpi-hello/bootstrap-mpi.yml
```diff
@@ -0,0 +1,13 @@
+---
+# Install OpenMPI packages from Ubuntu repos
+- hosts: all
+  become: true
+  tasks:
+  - name: install openmpi packages
+    apt:
```
dholt: Would be good to use `package` if the package names are the same or have a RHEL section, but ok for now

ajdecon: Annoyingly, the package names are different.

Also, the packages are built with slightly different options, and it's not clear if the RHEL packages include PMI support. This is relevant to the example because without PMI we need to manually supply a hosts file to `mpirun`.

I'm going to merge this as is for now, but note a follow-up to generalize this to RHEL.

PR #163, opened 2019-03-14
Add deployment script to add Dask/RAPIDS/K8S integration

supertetelman:
* deployment script
* helm value files
* k8s service yml files
* external git repo for custom image build
* *May not fully support multi-node systems without additional docker push commands.


supertetelman:
The current helm chart supports scaling via ReplicaSets, not scaling via the dask_kubernetes package.

Scaling from within Python code using dask_kubernetes is the preferered method. Until the helm charts and docker images are updated to reflect this we need to create a custom docker image with these libraries and a ServiceAccount to support them.

scripts/k8s_deploy_rapids_dask.sh
```diff
@@ -0,0 +1,125 @@
...
+# The Dask workers can be scaled from  the K8S command line using `kubectl scale` commands or through jupyter using dask_kubernetes commands
+# See the included sample notebooks for details.
+
+NAMESPACE="rapids" # TODO: Pass through different ns end to end
+if [ -z "${NAMESPACE}" ]; then
+    NAMESPACE="rapids"
+fi
+
+
+function build_image() {
```
supertetelman: The way this script is currently implemented it will build the Docker image locally, for multi-node systems we might need to add some scripting to push it to the local docker repository.

Within the next few months required library changes should be pushed upstream to the official RAPIDS or NGC docker images and we can remove this step altogether.

ajdecon:
Given that `config.example/helm/rapids-dask-svc.yml` is just straight k8s config, I suggest we create a new `config.example/k8s/` directory and move it there. This makes it clear that the file is meant to be consumed by `kubectl` rather than `helm`... and it doesn't seem unlikely that we'll have more k8s config files at some point.

supertetelman:
I agree. Adding a new directory for straight k8s yml files seems the right thing to do here and moving forward.

docs/rapids-dask.md
```diff
@@ -0,0 +1,30 @@
...
+Deploy the [LoadBalancer](ingress.md#on-prem-loadbalancer)
+
+Deploy Dask:
+
+```sh
+# Optionally, Modify chart configuration
+vi config/helm/rapids-dask.yml
+
+# Optionally, modify the K8S resources
+vi config/helm/rapids-dask-svc.yml
```
dholt: Needs an update

supertetelman: Updated

tvincereed:
Does this provide the right Dask/K8s infrastructure for the RAPIDS demos/benchmarks that Ty Mckercher is working on?

ajdecon:
@tvincereed: Based on the demo Ty shared with me on Friday, the demo code would require some minor changes to work on this deployment. It does look pretty easy to port: that demo uses a bash script to launch python on the cluster (not needed here), and uses the CLI to pass parameters to the calculation (would need to be changed to Jupyter notebook parameters).

I'm planning to spend some time tomorrow adapting that demo to a Jupyter deployment, as well as documenting how to run it in its CLI mode using Slurm.

tvincereed:
Thanks Adam, The ultimate goal is to provide a solid RAPIDS/Dask foundation that works for our demo/benchmarks. This makes those demos/benchmarks great QA tests for DeepOps deployments and also helps others who want to quickly get RAPIDS/Dask GPU accelerated infrastructure stood up fast.

PR #168, opened 2019-03-18
Re-add default kubespray config

dholt:
Re-adding config to fix pod doc until we come up with a better merged config plan

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,4 +1,0 @@
-export PATH=${PATH}:.
-export PDSH_RCMD_TYPE='ssh'
-export PDSH_SSH_ARGS='-l root'
-export KUBECONFIG=./admin.conf
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -60,20 +60,3 @@
 #software_extra_packages:
 #  - build-essential
 #  - "{{ 'linux-headers-' + ansible_kernel }}"
-
-# Kubernetes configuration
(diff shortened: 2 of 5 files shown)


michael-balint:
@dholt - I'm testing this, but am running into an issue with `kubeadm_enabled`...

```
TASK [deploy warning for non kubeadm] *************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'not kubeadm_enabled and not skip_non_kubeadm_warning' failed. The error was: error while evaluating conditional (not kubeadm_enabled and not skip_non_kubeadm_warning): 'kubeadm_enabled' is undefined\n\nThe error appears to have been in '/home/nvidia/Code/deepops/kubespray/cluster.yml': line 19, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  tasks:\n    - name: deploy warning for non kubeadm\n      ^ here\n"}
```

dholt:
How are you running? That variable is set by default already in Kubespray so we shouldn't need to re-define it ourselves:

```$ grep kubeadm_enabled kubespray/roles/kubespray-defaults/defaults/main.yaml
kubeadm_enabled: true```

michael-balint:
Updated comment above to show how we're running.

michael-balint:
Works after adding the `kubeadm_enabled` var in the config.

PR #171, opened 2019-03-19
Slurm example with RAPIDS + Dask

ajdecon:
Provide an example of running a RAPIDS + Dask benchmark on a Slurm cluster. Tested using a virtual DeepOps deployment with two compute nodes, each using 8 CPU cores and 2 Tesla P4 GPUs.

Actual benchmark code (`run.sh` and `sum.py`) provided by Ty Mckercher.


ajdecon:
cc @supertetelman @tvincereed

dholt:
Seems like we need to add `andrewrothstein.anaconda` to `requirements.yml`?

I think this is a reasonable role to include globally if we add a playbook to install Anaconda.

```
$ ./scripts/setup.sh
Updating apt cache...
sudo: unable to resolve host vca10
ansible 2.7.2
git version 2.7.4
ipmitool version 1.8.16
- dev-sec.ssh-hardening (6.0.0) is already installed, skipping.
- unxnn.users (56f9436e0d1e92a743e9f960d0b978ac8b72e070) is already installed, skipping.
- geerlingguy.docker (2.5.2) is already installed, skipping.
- lukeyeager.hosts (711ba98571f068a8bc6739fa1055ac38fc931010) is already installed, skipping.
- geerlingguy.ntp (1.6.2) is already installed, skipping.
Configuration directory './config' exists, not overwriting

dholt@vca10:/local/dholt/deepops
$ ansible-playbook -l slurm-cluster -u centos examples/slurm-bare-metal-dask-rapids/ansible-prereqs.yml
ERROR! the role 'andrewrothstein.anaconda' was not found in /local/dholt/deepops/examples/slurm-bare-metal-dask-rapids/roles:/local/dholt/deepops/galaxy-roles:/local/dholt/deepops/roles:/local/dholt/deepops/examples/slurm-bare-metal-dask-rapids

The error appears to have been in '/local/dholt/deepops/examples/slurm-bare-metal-dask-rapids/ansible-prereqs.yml': line 6, column 5, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  roles:
  - andrewrothstein.anaconda
    ^ here
```

ajdecon:
@dholt : D'oh! I had added that to requirements, but this got lost in the course of some rebasing fun.

Just added a commit which fixes the `requirements.yml` and adds a playbook for doing this globally.

dholt:
Also getting a failure checking CUDA:

```
TASK [ensure cuda is installed] *************************************************************************
failed: [mgmt-216-02] (item=cuda=10.0*) => {"changed": false, "item": "cuda=10.0*", "msg": "No package matching 'cuda=10.0*' found available, installed or updated", "rc": 126, "results": ["No package matching 'cuda=10.0*' found available, installed or updated"]}
failed: [mgmt-216-03] (item=cuda=10.0*) => {"changed": false, "item": "cuda=10.0*", "msg": "No package matching 'cuda=10.0*' found available, installed or updated", "rc": 126, "results": ["No package matching 'cuda=10.0*' found available, installed or updated"]}
failed: [hsw230] (item=cuda=10.0*) => {"changed": false, "item": "cuda=10.0*", "msg": "No package matching 'cuda=10.0*' found available, installed or updated", "rc": 126, "results": ["No package matching 'cuda=10.0*' found available, installed or updated"]}
        to retry, use: --limit @/local/dholt/deepops/examples/slurm-bare-metal-dask-rapids/ansible-prereqs.retry
```

requirements.yml
```diff
@@ -15,3 +15,6 @@
 
 - src: geerlingguy.ntp
   version: "1.6.2"
+
+- src: andrewrothstein.anaconda
+  version: "v4.0"
```
dholt: This isn't a valid version, `v4.0.8` works

michael-balint: Also - is there a reason to use Anaconda, or will miniconda suffice? Anaconda is a lot larger.

ajdecon: No strong reason for using anaconda vs miniconda afaik. I’ll see if I can swap this out and fix role dependencies.

I’ll also test the CUDA install issue. I suspect this is a CentOS vs Ubuntu issue.

Have meetings this morning but will likely push an updated version early this afternoon.

(I wish Github let me mark “needs changes” on my own PR, it’s a good way to flag that there are changes pending!)

ajdecon:
@dholt @michael-balint, updated the following and should be good to re-review.

- Replaced anaconda role with miniconda, verified valid role version
- Fixed `ansible-prereqs.yml` playbook to ensure it works with CentOS
- Tested full workflow on CentOS, following steps in `README.md`
- Added some detail to the `README.md` around SSH setup required to launch Dask under Slurm

examples/slurm-bare-metal-dask-rapids/ansible-prereqs.yml
```diff
@@ -0,0 +1,29 @@
...
+  roles:
+  - andrewrothstein.miniconda
+
+# CUDA components just on workers
+- hosts: slurm-workers
+  become: yes
+  roles:
+  - nvidia-ml
+
+  tasks:
```
dholt: Should we perhaps include a CUDA role or the driver playbook here instead? The driver role checks for GPUs and skips if they don't exist, but in a mixed CPU/GPU cluster this task section fails.

Alternatively, we could include something like this somewhere (from the driver playbook):

```
  tasks:
    - name: install custom facts
      include_role:
        name: facts

    - name: install nvidia driver
      include_role:
        name: nvidia-driver
      when: ansible_local['gpus']['count']
```

ajdecon: @dholt: 👍 I'll see if I can refactor this into a CUDA role that does the right checks.

examples/slurm-bare-metal-dask-rapids/ansible-prereqs.yml
```diff
@@ -0,0 +1,29 @@
+---
```
dholt: Can we install `git` via package manager as a dependency?

dholt:
Ditto about cudf failing to install, otherwise looks great to me.

ajdecon:
Based on our call this morning with Ty, I'd also like to move the RAPIDS Conda environment and the Dask deployment scripts into a role that can copy these files directly to the login node. This would provide a better user experience and also let us include a RAPIDS environment in Jenkins later if we want.

ajdecon:
Ping @michael-balint @dholt : When convenient, this needs re-reviewed after changes were addressed.

PR #177, opened 2019-03-21
updated scripts to isolate vagrant VM tasks from DeepOps tasks and also support CentOS

tvincereed:
new file:   vagrant_startup.sh
	deleted:    cluster_destroy.sh
	modified:   cluster_up.sh
	deleted:    scripts/setup_vagrant.sh
	deleted:    setup.sh


virtual/vagrant_startup.sh
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bash
+
+. /etc/os-release
+
+# Install Software
+case "$ID_LIKE" in
+    rhel*)
+        # Install Vagrant
+	set -xe
```
ajdecon: Nit: probably this line should just appear once at the top of the file.

virtual/vagrant_startup.sh
```diff
@@ -0,0 +1,126 @@
...
+	export DEEPOPS_VAGRANT_FILE="/home/$(whoami)/deepops/virtual/Vagrantfile-centos"
+
+
+
+	# End Install Vagrant
+        ;;
+    debian*)
+        # Install Vagrant
+
+set -xe
```
ajdecon: As above

virtual/vagrant_startup.sh
```diff
@@ -0,0 +1,126 @@
...
+
+	#set up Vagrantfile and start up the configuration in Vagrant
+	export DEEPOPS_VAGRANT_FILE="/home/$(whoami)/deepops/virtual/Vagrantfile-ubuntu"
+
+
+	# End Install Vagrant
+        ;;
+    *)
+        echo "Unsupported Operating System $ID_LIKE"
+        echo "Please install Ansible, Git, and python-netaddr manually"
```
ajdecon: Nit: Should mention Vagrant here as well, since we'll have skipped those install steps.

virtual/vagrant_startup.sh
```diff
@@ -0,0 +1,126 @@
...
+	# update yum
+	sudo yum  update
+
+	# install essential packages and tools
+	sudo yum -y install wget
+	sudo yum group install -y "Development Tools"
+	sudo yum install -y centos-release-qemu-ev qem-kvm-ev qemu-kvm libvirt virt-install bridge-utils libvirt-devel  libxslt-devel libxml2-devel libvirt-devel libguestfs-tools-c
+
+	# set up networking for Vagrant VMs
+	#sudo echo "net.ipv4.ip_forward = 1"|sudo tee /etc/sysctl.d/99-ipforward.conf
```
ajdecon: Nit: leave these lines out rather than comment out.

virtual/vagrant_startup.sh
```diff
@@ -0,0 +1,126 @@
...
+	# install kvm packages
+	sudo yum install -y qemu-kvm libvirt-bin libvirt-dev bridge-utils libguestfs-tools
+	sudo yum install -y qemu virt-manager firewalld OVMF
+
+	# install vagrant plugins
+	vagrant plugin install vagrant-hostmanager vagrant-libvirt
+	vagrant plugin install vagrant-host-shell vagrant-scp vagrant-mutate
+
+	#set up Vagrantfile and start up the configuration in Vagrant
+	export DEEPOPS_VAGRANT_FILE="/home/$(whoami)/deepops/virtual/Vagrantfile-centos"
```
ajdecon: 1. This assumes the location of the deepops repo and home directory. Instead, maybe use the pattern for setting `VIRT_DIR` further down to get the absolute path of the script directory? 
1. This line (and the one like it for Ubuntu) assumes we want the same host OS and VM OS, but there are use cases where we'll want these to be different. Instead of hard-coding this, can we get it from the environment and set a default? E.g.,

```
DEEPOPS_VAGRANT_FILE=${DEEPOPS_VAGRANT_FILE:-${VIRT_DIR}/Vagrantfile-centos}
```

tvincereed:
Thanks for the great feedback. Sorry I've been tied up with other tasks and haven't been able to implement the changes yet. I'll work on it this weekend as best I can. Thanks again Adam, Doug, Mike

virtual/vagrant_startup.sh
tvincereed: Sure. I still think the logic below this line permits that but was making it a setting that would work OOB. Long term I talked to Doug and agree that the Vagrant file should be auto generated based on config files rather than be static. I think he was going to have a .yml file and ansible the appropriate Vagrantfile.

virtual/vagrant_startup.sh
tvincereed: good catch.. This was unchanged remnants of the scripts/setup.sh file that I copied. Completely agree

virtual/vagrant_startup.sh
tvincereed: totally agree!

michael-balint:
Looking at the changes due to the last commit, here are some pooled comments...

* Is the intention to use `vagrant_startup.sh` from now? If so, we should also update the README.md to reference this instead of `cluster_up.sh`. Also the Jenkinsfile should reflect this change or else the build test won't pass (it is failing now and this will impact all further checks from other PRs).

* In `vagrant_startup.sh`, we also reference `./scripts/setup.sh`. Do we want to remove this? We don't make any calls to the k8s or slurm setup scripts. Is this intentional?

* `cluster_up.sh` still exists - did we want to delete this?

* The `setup.sh` is pulled into `vagrant_startup.sh`. This means that even if you have everything installed, it'll run through this process. It isn't a big deal for the packages that are installed via apt/yum, but it is for the files that are downloaded and installed that way. It adds extra install time and will deplete the disk space on the test server. My suggestion here is that either we pull the setup back into a separate script that can be run once - or implement some nice checks to make sure we aren't re-installing components.

* `vagrant_shutdown.sh` should do more than just `vagrant destroy -f`. There are cases where the destroy command doesn't work and we must resort to using virsh. This is relevant to the test server since it is somewhat ok if this occurs on someone's personal machine, but not ok if it backs up the test queue. Also, it's just nice to have for the personal case too...

Instead of `vagrant destroy -f`, replace with...

```
if vagrant destroy -f ; then
  echo "Destroyed cluster."
else
  echo "Unable to vagrant destroy. Manually removing the VMs."
  virsh destroy virtual_virtual-mgmt
  virsh undefine virtual_virtual-mgmt
  virsh vol-delete --pool default virtual_virtual-mgmt.img
  virsh destroy virtual_virtual-login
  virsh undefine virtual_virtual-login
  virsh vol-delete --pool default virtual_virtual-login.img
  virsh destroy virtual_virtual-gpu01
  virsh undefine virtual_virtual-gpu01
  virsh vol-delete --pool default virtual_virtual-gpu01.img
fi
```

tvincereed:
Michael, in-lined comments below. Sorry for my poorly documented changes.

Summary of changes:

  1.  Move all vagrant/VM related actions to two files (vagrant_startup.sh, vagrant_shutdown.sh). The rational is to eventually have the virtual cluster be less “stand-a-lone” and perhaps no longer need its own directory of config and script files. The longer term concept would be to use ansible or other to build the Vagrantfile to match what is in the baseline deepops config files (user modified) and then the actual DeepOps deployment in “virtual” is using the identical scripts and playbooks that a physical cluster deployment would use.  I think this has merit for both Jenkins testing and also for realworld users who want to test drive a config virtually before they touch real HW.

There are two future efforts that need to be taken:

i)                    We need to construct the Vagrantfile data from the deepops configuration data in an automated way

ii)                  We need to add GPU pass thru setting into Vagrantfile in an automated way.

  1.  Cluster_up.sh is really a simple execution script of deploying DeepOps. I think it belongs in DeepOps/scripts and renamed something like “virtual_cluster_test.sh”.
  2.  Cluster_destroy.sh is really just vagrant/VM related work so better named vagrant_shutdown.sh to keep all the vagrant/VM related stuff in two scripts.
  3.  The vagrant related actions in cluster_up.sh were moved to vagrant_startup.sh to keep all vagrant/VM related stuff to two scripts.

Vagrant_startup.sh was modified to work for both CentOS and Ubuntu and contain all the vagrant related installation and setup.

michael-balint:
@tvincereed - Agree with the philosophy 100%. We need to make some changes to this PR though - otherwise the build test won't pass and all future PR builds will also fail. Let's catch up today and go through the changes that need to happen so we can merge this.

michael-balint:
@tvincereed - Encapsulated the changes into a PR: https://github.com/tvincereed/deepops/pull/1

michael-balint:
Closing in favor of #188, which includes these changes and other fixes that make Jenkins work.

PR #184, opened 2019-03-24
Set up MetalLB by default in virtual cluster

ajdecon:
There are a lot of options when it comes to setting up external access
to a k8s cluster, but a simple L2 load balancer is a good default
option. So let's set it up by default, and let the user make changes
later if they like.

This follows the general principle that a virtual cluster should be as
useful as possible when turned up with the default config.

## Test plan

Turned up a fresh virtual cluster with this script included, then ran `examples/k8s-dask-rapids/deploy.sh` to deploy an example that depends on MetalLB. Confirmed that external IPs were assigned to services in this example:

```
ubuntu@ivb120:~/src/deepops/examples/k8s-dask-rapids$ kubectl -n rapids get services
NAME                    TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                       AGE
rapids-dask-jupyter     LoadBalancer   10.233.40.93   10.0.0.101    80:31807/TCP                  100s
rapids-dask-scheduler   LoadBalancer   10.233.33.72   10.0.0.100    8786:31028/TCP,80:31498/TCP   100s
```


michael-balint:
Nice - I was thinking we should do this by default as well. This would also make Kubeflow work by default without changing it to NodePort (which is I think what we do in the Kubeflow deploy script right now).

ryanolson:
I'd prefer if you set up a single MetalLB for an ingress controller like ingress-nginx or istio, then used ingress options to route to services.  

I think it's abusing MetalLB to create a LoadBalancer for every service.

michael-balint:
@ryanolson - Do you feel that way for virtual/testing as well? Do you have some sample setup you can point us to?

ryanolson:
Here is one example of exposing a Rook/Ceph Object Store as an S3 service:

https://github.com/NVIDIA/tensorrt-laboratory/tree/master/examples/Deployment/ObjectStore

For testing, I like the idea of resolving any hostname (*.local; or whatever domain suffix is being used) to a single MetalLB IP.  That way you can have as many services as you want and you don't have to tie them each to a new IP that MetalLB will provide you. 

Just a thought.

ajdecon:
@ryanolson : That makes sense. I agree that configuring an ingress is a good best practice that we should model in the examples we include with DeepOps. That's not something I think we do with most of our scripts now, so we may want to go back and adjust them if we think that's important.

Within the scope of this PR (for the default script to turn up a virtual cluster), would your suggested change be that we stand up an ingress controller by default in addition to the MetalLB load balancer? Or are you asking for a different specific change?

And I would think that the specific ingress configuration for any given service (such as the RAPIDS example I used to test) should be separate PRs, as the config details would be different for each one.

ajdecon:
Made a few changes:

* Added a default nginx ingress controller
* Refactored deployment of LB/ingress into its own script
* Added a test for the ingress to the `Jenkinsfile`

Successfully tested the script with a fresh virtual turnup, and can see that the nginx controller is getting an external IP:

```
ubuntu@ivb120:~/src/deepops/virtual/scripts$ kubectl get services -l app=nginx-ingress
NAME                            TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
nginx-ingress-controller        LoadBalancer   10.233.54.60    10.0.0.100    80:30678/TCP,443:32352/TCP   13m
nginx-ingress-default-backend   ClusterIP      10.233.51.129   <none>        80/TCP                       13m
```

Will monitor the Jenkins logs to see if the new Jenkins tests work.

ajdecon:
Ok, looks like the Jenkins job failed because the LB wasn't online yet:

```
+ kubectl get services -l app=nginx-ingress -l component=controller --no-headers

+ awk {print $4}

+ nginx_external_ip=<pending>

+ curl http://<pending>/

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current

                                 Dload  Upload   Total   Spent    Left  Speed


  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (6) Could not resolve host: <pending>

```

There's already a sleep present, but I'll see if there's a way to wait for the LB to be available in a smarter fashion.

dholt:
I disagree with the philosophy here. Every third-party service or k8s project I've seen has used a LoadBalancer service type by default because that's what works out of the box on cloud deployments. Adding or requiring an ingress controller adds extra complexity and configuration that aren't necessary. For example, using this method requires DNS setup. I still think we need an ingress controller but it should be separate.

michael-balint:
@ajdecon It might be time to use ansible 😂

Or use something like this: https://github.com/groundnuty/k8s-wait-for/blob/master/wait_for.sh

config.example/helm/ingress.yml
```diff
@@ -7,9 +7,11 @@
 controller:
   # Use host network to listen on ports 80 and 443
   hostNetwork: true
+  # Service type LoadBalancer requires a load balancer to be configured, e.g.
+  # MetalLB in an on-prem cluster. See metallb.yml for a sample definition.
   # Use NodePort for on-prem where we don't have a load balancer
   service:
-    type: NodePort
+    type: LoadBalancer
```
dholt: I think this is a good change, but we're going to need to update the docs to show that the LB is required before bringing up the ingress controller.

ajdecon:
@dholt : do you think we should only include a MetalLB load balancer in the default cluster? Or include both the load balancer and an ingress, but not configure our examples to use them by default?

Just trying to clarify for scope. :) I think this PR should include the LB/ingress infrastructure we want to include by default, but that any changes we make to our services config should be separate.

dholt:
If we can do something useful with the ingress controller by default, then it should be included. But I don't want it to be required when using the load balancer, and I don't want the ingress controller to be used by default if it requires complicated external configuration.

ajdecon:
> @ajdecon It might be time to use ansible 😂
> 
> Or use something like this: https://github.com/groundnuty/k8s-wait-for/blob/master/wait_for.sh

@michael-balint : And in fact I'm implementing some basic retry logic in my current script. :D If my current minimal retry logic works I'll leave it in, otherwise I'll have reached my threshold to convert shell to Python and/or Ansible.

michael-balint:
Hopefully it works. The thing I like about having `cluster_up`, etc in virtual as a bash script vs ansible is that it is easy for users to look at virtual line-by-line to set up their own cluster. W/ ansible, since roles are used, it's not as obvious.

ajdecon:
Ok, even after some long pauses, IPs aren't being successfully assigned in the Jenkins environment, even though I've tested this successfully on two hosts outside Jenkins. I suspect it's either an environment issue or the way I'm adjusting the IP numbering.

Unfortunately the only way I have to test in the Jenkins environment... is by pushing updates. (@michael-balint any better ideas?) And I'd like to discover the issue to make sure it's not something that would impact other parts of the environment.

All of which is a long-winded way of saying, apologies for spamming those subscribed here with multiple test commits to trigger Jenkins builds. :P

PR #186, opened 2019-03-25
Add NVIDIA recommended container settings

dholt:
Add recommended Docker settings as defaults for the daemon.
Clean up various places where we set config in the playbooks.


ajdecon:
* The new `docker_daemon_json` doesn't include the section specifying `nvidia-container-runtime`. Is that an intentional change, or did that get dropped by accident?
* Can we add a comment referencing whatever document recommends these settings? I'm sure they're a good idea, but I hate having magic numbers without a pointer to their source. :)

dholt:
> The new docker_daemon_json doesn't include the section specifying nvidia-container-runtime. Is that an intentional change, or did that get dropped by accident?
It does, in `config.example/group_vars/dgx-servers.yml` and `config.example/group_vars/gpu-servers.yml`

> Can we add a comment referencing whatever document recommends these settings? I'm sure they're a good idea, but I hate having magic numbers without a pointer to their source. :)
Will do

ajdecon:
OOPS, sorry, the diff view fooled me.

ajdecon:
@dholt : The Jenkins job for this PR is still running and appears to be stuck at the stage of running a GPU job. Can you check to see if this is due to the new config?

michael-balint:
@ajdecon - it's due to DNS issues, rebooting the server to apply some changes so hopefully we won't have that in the future

PR #192, opened 2019-03-27
Fix path in user message

dholt:


ajdecon:
LGTM (pending Jenkins success)

PR #194, opened 2019-03-28
Add scripts to download an offline copy of DeepOps dependencies

ajdecon:
Initial draft of scripts to download DeepOps dependencies and roll them into a big tarball that can be transferred offline. This is a starting point toward making offline/air-gap support easier to manage.

These are obviously incomplete at this point: we have additional dependency downloads hard-coded into a variety of Ansible playbooks, Helm charts, etc, and they may pull in more. But I think this captures all the major dependency *types*, and can serve as a starting point as we start testing offline installs.

I've tested each of the individual download scripts, and can confirm that the driver script successfully runs everything. I'm currently doing a full run to see what the final tarball size is, and will report back. :)


ajdecon:
Jenkins is failing, but it looks like a package dependency issue for `nvidia-docker2`, that wouldn't be impacted by the new scripts being added in this PR.

```
failed: [node2] (item=[u'nvidia-docker2', u'nvidia-container-runtime']) => {
    "cache_update_time": 1553798878, 
    "cache_updated": true, 
    "changed": false, 
    "invocation": {
        "module_args": {
            "allow_unauthenticated": false, 
            "autoclean": false, 
            "autoremove": false, 
            "cache_valid_time": 0, 
            "deb": null, 
            "default_release": null, 
            "dpkg_options": "force-confdef,force-confold", 
            "force": false, 
            "force_apt_get": false, 
            "install_recommends": null, 
            "name": [
                "nvidia-docker2", 
                "nvidia-container-runtime"
            ], 
            "only_upgrade": false, 
            "package": [
                "nvidia-docker2", 
                "nvidia-container-runtime"
            ], 
            "purge": false, 
            "state": "present", 
            "update_cache": true, 
            "upgrade": null
        }
    }, 
    "item": [
        "nvidia-docker2", 
        "nvidia-container-runtime"
    ], 
    "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"     install 'nvidia-docker2' 'nvidia-container-runtime'' failed: E: Unable to correct problems, you have held broken packages.\n", 
    "rc": 100, 
    "stderr": "E: Unable to correct problems, you have held broken packages.\n", 
    "stderr_lines": [
        "E: Unable to correct problems, you have held broken packages."
    ], 
    "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nSome packages could not be installed. This may mean that you have\nrequested an impossible situation or if you are using the unstable\ndistribution that some required packages have not yet been created\nor been moved out of Incoming.\nThe following information may help to resolve the situation:\n\nThe following packages have unmet dependencies:\n nvidia-docker2 : Depends: docker-ce (= 5:18.09.3~3-0~ubuntu-bionic) but 5:18.09.4~3-0~ubuntu-bionic is to be installed or\n                           docker-ee (= 5:18.09.3~3-0~ubuntu-bionic) but it is not installable\n", 
    "stdout_lines": [
        "Reading package lists...", 
        "Building dependency tree...", 
        "Reading state information...", 
        "Some packages could not be installed. This may mean that you have", 
        "requested an impossible situation or if you are using the unstable", 
        "distribution that some required packages have not yet been created", 
        "or been moved out of Incoming.", 
        "The following information may help to resolve the situation:", 
        "", 
        "The following packages have unmet dependencies:", 
        " nvidia-docker2 : Depends: docker-ce (= 5:18.09.3~3-0~ubuntu-bionic) but 5:18.09.4~3-0~ubuntu-bionic is to be installed or", 
        "                           docker-ee (= 5:18.09.3~3-0~ubuntu-bionic) but it is not installable"
    ]
}
```

Troubleshooting locally to see if the breakage is due to a problem in DeepOps or upstream.

dholt:
@ajdecon that nvidia-docker issue is what was fixed in my last PR to pin nvdocker/docker versions. Docker released an 18.09.4 but nvidia-docker only has a version built against 18.09.3.

ajdecon:
@dholt ACK. Rebased on master to trigger a rebuild with your fix included.

ajdecon:
Complete download and tar test succeeded in ~4 hours. Final tarball was 53 GB.

The bulk of the data is mirrors of CentOS and EPEL repos. We should include this functionality, because we definitely need to ensure this data exists in an offline context... but we should maybe make it optional, as any given site might already have these mirrors running. :)

dholt:
Could we possibly pull only the packages we need and give the user an option to add other packages to the list?

ajdecon:
That would be technically possible but tedious. Enumerating all the packages we request + dependencies basically requires that we build a cluster and then list installed packages, and this list would have to be re-generated frequently to account for changes to DeepOps and dependency changes upstream.

Mirroring everything in the repos we use is pretty simple and robust, it just takes a while and needs a lot of space. And the largest repo in the list is `cuda`, which I think we'd want to mirror anyway:

```
ubuntu@ivb120:/tmp/deepops/yum-mirror$ du -sh ./*
919M    ./CentOS-7-x86_64-Minimal-1810.iso
9.5G    ./base
577M    ./centosplus
19G     ./cuda
813M    ./docker-ce-stable
451M    ./docker-engine
15G     ./epel
507M    ./extras
7.1G    ./updates
```

I do think the right answer is to include this functionality and just put it behind a `--download-the-world`  sort of flag, to avoid wasting time for those who have downloaded the world in the past. ;-)

ajdecon:
Closing this PR, will replace with an Ansible-based version later.

PR #198, opened 2019-03-29
fix url for rook helm charts

ajdecon:
This broke in my earlier #196, but wasn't caught because all test hosts already had the helm repo present.


ajdecon:
Obsoleted by #201

PR #199, opened 2019-03-29
[WIP] Ansible playbook to build an offline cache of dependencies

ajdecon:
Ansible playbook to download DeepOps dependencies and roll them into a big tarball that can be transferred offline. This is a starting point toward making offline/air-gap support easier to manage. #194 was a previous version of this using shell scripts.

These are obviously incomplete at this point: we have additional dependency downloads hard-coded into a variety of Ansible playbooks, Helm charts, etc, and they may pull in more. But I think this captures all the major dependency types, and can serve as a starting point as we start testing offline installs.

I've tested this playbook and confirmed it does download all the enabled dependencies, including running reposync.


ajdecon:
> Is there a reason to have the offline_cache directory in config.example vs. just adding the inventory to the main inventory file and vars to a new group_vars file?

Agreed, good idea.

> Also, can we move playbooks/files/offline_cache/yum.conf to a role (i.e. "airgap")?

Yes, the whole thing should be migrated to be a role.

roles/cache_builder/defaults/main.yml
```diff
@@ -0,0 +1,67 @@
...
+- "docker-ce-stable"
+- "docker-engine"
+- "epel"
+
+helm_download_url: "https://storage.googleapis.com/kubernetes-helm/helm-v2.11.0-linux-amd64.tar.gz"
+helm_chart_git_repos:
+  base: "https://github.com/helm/charts"
+rook_helm_chart_url: "https://charts.rook.io/master/rook-ceph-v0.9.0-79.g1a1ffdd.tgz"
+
+docker_image_list:
```
ryanolson: The images that belong to kubespray should use the values found in kubespray's default role for the fixed version of kubespray you are installing.

roles/cache_builder/defaults/main.yml
```diff
@@ -0,0 +1,67 @@
...
+tar_file_path: "/tmp/deepops-archive.tar.bz2"
+deepops_repo_url: "https://github.com/NVIDIA/deepops"
+
+centos_iso_mirror: "http://mirrors.ocf.berkeley.edu/centos/7.6.1810/isos/x86_64"
+centos_iso: "CentOS-7-x86_64-Minimal-1810.iso"
+download_yum_repos: yes
+yum_repos_to_sync:
+- "libnvidia-container"
+- "nvidia-container-runtime"
+- "nvidia-docker"
```
dholt: I think this should be `nvidia-docker2`

roles/cache_builder/defaults/main.yml
ajdecon: Good call. I'll check to see if there's a good way to hoist that variable into this role.

roles/cache_builder/defaults/main.yml
ajdecon: The repo name is `nvidia-docker`, and is found at `https://nvidia.github.io/nvidia-docker`. It actually contains both the `nvidia-docker` and `nvidia-docker2` packages.

dholt: Sorry, thought these were package names. Where do the repo URLs come from?

ajdecon: All the URLs are in `roles/cache_builder/files/yum.conf`

roles/cache_builder/defaults/main.yml
ajdecon: Now including a run of the kubespray download role in `scripts/build_offline_cache.sh`.

ajdecon:
Marked this WIP, please don't merge for now. I'm working on the inverse process (building repo mirrors from the archive) and this will need some changes.

ajdecon:
This got a lot bigger than I expected, but I think this is ready to review. Feedback very much appreciated. cc @dholt, @tvincereed, @michael-balint 

Right now this PR does two things:

* Download all the DeepOps dependencies (that I've identified) and put them in a big tar file
* Extract that tar file, and spin up docker containers to serve repositories containing those dependencies

Instructions to do this are found in the PR as `docs/offline.md`. Ideally to test this you should perform the two steps on different hosts to make it clean. You can test them on the same host and it should work but it's not as good of a test.

The next step for this project is to set up a variables file pointing to the local repositories and attempt a DeepOps turn-up without Internet access, but I think that should probably be a separate PR.

PR #201, opened 2019-03-29
Fix default rook chart repo

supertetelman:


scripts/k8s_deploy_rook.sh
```diff
@@ -5,7 +5,9 @@
 # `helm search rook` # get latest version number
 # `helm upgrade --namespace rook-ceph-system rook-ceph rook-master/rook-ceph --version v0.9.0-174.g3b14e51`
 
-HELM_ROOK_CHART_REPO="${HELM_ROOK_CHART_REPO:-HELM_ROOK_CHART_REPO}"
+HELM_ROOK_CHART_REPO="${HELM_ROOK_CHART_REPO:-https://charts.rook.io/master}"
+echo $HELM_ROOK_CHART_REPO
+exit 1
```
ajdecon: This looks like it will always exit unconditionally, so the rest of the script never runs.

supertetelman: That's what I get for debugging and changing code on multiple boxes. Fixed.

PR #205, opened 2019-04-03
Configure monitoring to use Ingress

dholt:
Modify monitoring services to use ingress routes instead of NodePorts without requiring a load balancer or external DNS configuration.


ajdecon:
(Having made that script change locally, that change to `virtual/scripts/setup_k8s.sh` does succeed.)

ajdecon:
It passes! Accepted.

PR #207, opened 2019-04-04
Add deployment script/config files to standup a local BinderHub instance

supertetelman:
This setup script will deploy a running instance of BinderHub (https://binderhub.readthedocs.io).

The intended use case of this script is to quickly get a user up-and-running on DeepOps with an easy-to-use system for building Docker images and running demos from external Github repos.

Due to current BinderHub limitations, this will attempt to deploy notebooks with 1 GPU, this amount can be modified in the config.yml file, but cannot be adjusted per notebook.

### DIFF
--- a/config.example/helm/binderhub-config.yml
+++ b/config.example/helm/binderhub-config.yml
@@ -2,3 +2,12 @@ config:
   BinderHub:
     use_registry: false
     # hub_url: 
+
+# Require 1 GPU for each Notebook deployment
+jupyterhub:
+  singleuser:
+    extraResource:
+      guarantees:
+        nvidia.com/gpu: "1"
+      limits:
+        nvidia.com/gpu: "1"
--- a/config.example/helm/binderhub-config.yml
+++ b/config.example/helm/binderhub-config.yml
@@ -1,0 +1,4 @@
+config:
+  BinderHub:
+    use_registry: false
+    # hub_url: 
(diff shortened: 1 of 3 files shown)


atetelman:
Interesting. I've been testing/working with this on my DGX and in the cloud, not through virtual. I don't currently have a system up I can try virtual on.

The error you are seeing there is from this line. I'm just iterating through all the interfaces and trying to find the one that BinderHub will respond to. I'm making the assumption that the LoadBalancer isn't actually working correctly and I need to use the NodePort.

It looks like the not seeing `binderhub-binderhub` bit might be a bug/issues in the BinderHub helm chart. It's probably referring to the `binderhub` svc instead.

```
  for local_ip in `ip -br addr  | awk '{print $3}' | awk -F/ '{print $1}'`; do
    curl --max-time .1 --connect-timeout .1 -L ${local_ip}:${binderhub_port} && break
  done
```

michael-balint:
@atetelman - I'm still getting the output above (about not being able to connect to a server). That said, I _am_ able to reach the binderhub service when I use http://<external_ip>:80. Why aren't we setting `local_ip` to the external ip that the LoadBalancer assigns to the service?

```
$ kubectl get svc -n binderhub
NAME           TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
binder         LoadBalancer   10.233.14.240   10.0.0.101    80:31311/TCP                 35m
hub            ClusterIP      10.233.52.184   <none>        8081/TCP                     35m
proxy-api      ClusterIP      10.233.1.90     <none>        8001/TCP                     35m
proxy-public   LoadBalancer   10.233.15.17    10.0.0.100    80:31219/TCP,443:31593/TCP   35m
```

PR #212, opened 2019-04-06
Virtual fixes for test server

michael-balint:
On the test server, we're periodically encountering errors due to dpkg locking when builds coincide. This PR fixes that and speeds up the setup process if you already have virtual dependencies installed.

### DIFF
--- a/virtual/vagrant_startup.sh
+++ b/virtual/vagrant_startup.sh
@@ -11,26 +11,33 @@ VIRT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 case "$ID_LIKE" in
   rhel*)
-    # End Install Vagrant & Dependencies for RHEL Systems
+    # Install Vagrant & Dependencies for RHEL Systems
 
-    # Update yum
-    sudo yum update
+    if ! rpm -q wget "Development Tools" centos-release-qemu-ev qem-kvm-ev qemu-kvm libvirt virt-install \
+      bridge-utils libvirt-devel libxslt-devel libxml2-devel libguestfs-tools-c sshpass qemu-kvm libvirt-bin \
+      libvirt-dev bridge-utils libguestfs-tools qemu virt-manager firewalld OVMF 2>&1; then
 
-    # Install essential packages and tools
-    sudo yum -y install wget
-    sudo yum group install -y "Development Tools"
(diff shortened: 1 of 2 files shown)


virtual/vagrant_startup.sh
```diff
@@ -11,27 +11,23 @@ VIRT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
-    # End Install Vagrant & Dependencies for RHEL Systems
+    # Install Vagrant & Dependencies for RHEL Systems
+
+    export YUM_DEPENDENCIES="centos-release-qemu-ev qem-kvm-ev qemu-kvm libvirt virt-install \
+      bridge-utils libvirt-devel libxslt-devel libxml2-devel libguestfs-tools-c sshpass qemu-kvm libvirt-bin \
+      libvirt-dev bridge-utils libguestfs-tools qemu virt-manager firewalld OVMF"
 
-    # Update yum
-    sudo yum update
+    if ! (yum grouplist installed "Development Tools" && rpm -q $YUM_DEPENDENCIES) >/dev/null 2>&1; then
```
ajdecon: `yum grouplist installed` doesn't correctly fail if the group is not installed. To check the group directly, we'll need to do something like a pipe to grep that will handle return code correctly.

```
[vagrant@virtual-mgmt ~]$ sudo yum grouplist installed Scientific Support
Failed to set locale, defaulting to C
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: mirrors.syringanetworks.net
 * extras: mirrors.xtom.com
 * updates: mirrors.ocf.berkeley.edu
Warning: no environments/groups match: installed, Scientific, Support
[vagrant@virtual-mgmt ~]$ echo $?
0
[vagrant@virtual-mgmt ~]$ yum grouplist installed | grep "Scientific Support"
Failed to set locale, defaulting to C
[vagrant@virtual-mgmt ~]$ echo $?
1
[vagrant@virtual-mgmt ~]$ yum grouplist installed | grep "Development Tools"
Failed to set locale, defaulting to C
   Development Tools
[vagrant@virtual-mgmt ~]$ echo $?
0
```

`rpm -q` does appear to behave correctly.

ajdecon:
LGTM! Verified correct behavior on Ubuntu and RHEL test hosts.

PR #214, opened 2019-04-08
Make use of setup.sh clearer

ajdecon:
* Fix `setup.sh` so that it can be run from any directory (not just the repo root)
* Fix the instructions in `virtual/README.md` that refer to the wrong `setup.sh`

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -2,6 +2,9 @@
 
 . /etc/os-release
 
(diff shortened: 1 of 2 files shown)


ScottESanDiego:
Looks much clearer, thanks!

scripts/setup.sh
```diff
@@ -2,6 +2,9 @@
 
 . /etc/os-release
 
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
michael-balint: @dholt - calling this out so you're aware of it

PR #216, opened 2019-04-09
Issue 215, add slurm dependency

supertetelman:
* add kernel version to linux tools install


dholt:
I think it's better practice to install the `-generic` meta package vs the specific kernel version. This isn't going to fix the issue you ran into if you installed a newer kernel than the version you're currently running.

PR #221, opened 2019-04-11
Updates to Virtual README.md

satindern:
Centos 7.6

### DIFF
--- a/virtual/README.md
+++ b/virtual/README.md
@@ -22,76 +22,96 @@ Total: 6 vCPU, 20 GB RAM, 96 GB Storage
 * Ubuntu 16.04 (or greater)
 * CentOS 7.6 (or greater)
 
-Running DeepOps virtually assumes that the host machine's OS is an approved OS. If this is not the case, the `bootstrap_virtual.sh` and `cluster_up.sh` scripts may be modified to work with a different OS.
+Running DeepOps virtually assumes that the host machine's OS is an approved OS. If this is not the case, the scripts used in the steps below may be modified to work with a different OS.
 
 Also, using VMs and optionally GPU passthrough assumes that the host machine has been configured to enable virtualization in the BIOS. For instructions on how to accomplish this, refer to the sections at the bottom of this README: [Enabling virtualization and GPU passthrough](#enabling-virtualization-and-gpu-passthrough).
 
-## Bootstrap dependencies
+## Start the Virtual Cluster
 
-To install basic dependencies for running Ansible and managing a DeepOps cluster, run the [/scripts/setup.sh](/scripts/setup.sh) script from the root deepops directory.
+1. From the main deepops directory, run the setup script.
 
+   This will install Ansible and other software on the provisioning machine which will be used to deploy all other software to the cluster. For more information on Ansible and why we use it, consult the [Ansible Guide](/docs/ANSIBLE.md).
+
+   ```sh
+   ./scripts/setup.sh
+   ```
+
+2. In the virtual directory, startup vagrant. This will start 3 VMs:
+   * virtual-login
+   * virtual-mgmt
+   * virtual-gpu01
+
+   ```sh
+   # NOTE: The default VM OS is Ubuntu. If you wish the VMs to spawn CentOS,
+   #       configure the DEEPOPS_VAGRANT_FILE variable accordingly...
+   #       export DEEPOPS_VAGRANT_FILE=$(pwd)/Vagrantfile-centos
+   # NOTE: virtual-gpu01 requires GPU passthrough, by default it is not enabled
+   cd virtual
+   ./vagrant_startup.sh
+   ```
+
+3. Start the cluster.
+
+   ```sh
+   # NOTE: Only Kubernetes is deployed by default. To also deploy Slurm,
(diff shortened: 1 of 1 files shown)


virtual/README.md
```diff
@@ -8,18 +8,13 @@ Set up a virtual cluster with DeepOps. Useful for...
 
 ## Prerequisites
 
-Running DeepOps virtually assumes that the host machine's OS is Ubuntu 16.04 or greater. If this is
-not the case, the `bootstrap_virtual.sh` and `cluster_up.sh` scripts may be modified to work with
-a different OS.
+Running DeepOps virtually assumes that the host machine's OS is Ubuntu 16.04 or greater. DeepOps is also supported on CentOS 7.6 or greater. If this is not the case, the `bootstrap_virtual.sh` and `cluster_up.sh` scripts may be modified to work with a different OS.
```
michael-balint: Can we say instead...

Approved OS:
- Ubuntu 16.04 (or greater)
- CentOS 7.6 (or greater)

Running DeepOps virtually assumes that the host machine's OS is an approved OS. If this is not the case, the `bootstrap_virtual.sh` and `cluster_up.sh` scripts may be modified to work with a different OS.

virtual/README.md
```diff
@@ -102,8 +97,7 @@ $ lspci -nnk | grep NVIDIA
```
michael-balint: I thought the `./cluster_up.sh` instruction was confusing and you were going to add an explanation for what it does? Line 54.

michael-balint: Also, add a line explaining source k8s_environment.sh

virtual/README.md
```diff
@@ -8,18 +8,17 @@ Set up a virtual cluster with DeepOps. Useful for...
...
-Running DeepOps virtually assumes that the host machine's OS is Ubuntu 16.04 or greater. If this is
-not the case, the `bootstrap_virtual.sh` and `cluster_up.sh` scripts may be modified to work with
-a different OS.
+Approved OS:
 
-Also, using VMs and optionally GPU passthrough assumes that the host machine has been configured to
-enable virtualization in the BIOS. For instructions on how to accomplish this, refer to the sections
-at the bottom of this README: `Enabling virtualization and GPU passthrough`.
+Ubuntu 16.04 (or greater)
+CentOS 7.6 (or greater)
```
michael-balint: These should be bulleted (use a `* ` before each one). Also need an additional newline after the second bullet so that there is space for the next paragraph.

virtual/README.md
```diff
@@ -8,18 +8,17 @@ Set up a virtual cluster with DeepOps. Useful for...
...
+Running DeepOps virtually assumes that the host machine's OS is an approved OS. If this is not the case, the bootstrap_virtual.sh and cluster_up.sh scripts may be modified to work with a different OS.
```
michael-balint: Looks like we lost the code formatting here. Use ` to surround the scripts.

virtual/README.md
```diff
@@ -8,18 +8,17 @@ Set up a virtual cluster with DeepOps. Useful for...
...
+
+Also, using VMs and optionally GPU passthrough assumes that the host machine has been configured to enable virtualization in the BIOS. For instructions on how to accomplish this, refer to the sections at the bottom of this README: `Enabling virtualization and GPU passthrough`.
```
michael-balint: This portion should be a link: [Enabling virtualization and GPU passthrough](#enabling-virtualization-and-gpu-passthrough)

michael-balint:
Should also add a section on VM requirements...
- 2 CPUs per VM (6 total)
- 2048 MB RAM for mgmt, login
- 16384 MB RAM for gpu01

virtual/README.md
satindern: remind me here, `./cluster_up.sh` deploys the k8s cluster within the vagrant vms created in the `./vagrant_startup.sh` step. correct?

satindern: `source k8s_environment.sh` enables us to run `kubectl` from the host without having to ssh into the vm, correct? where should this be added?
after [Start the Cluster](https://github.com/satindern/deepops/blob/patch-1/virtual/README.md#start-the-cluster)?

michael-balint: Yes. `./vagrant_startup.sh` starts the VMs. `./cluster_up.sh` runs the DeepOps scripts. Specifically, it deploys Kubernetes and/or Slurm.

michael-balint: It enables us to run kubectl on the provisioning node (where we ran the deepops scripts). When `./cluster_up.sh` runs, it copies a version of `kubectl` and `admin.conf` to the `k8s-config/artifacts` directory. The `k8s_environment.sh` script adds `kubectl` to the path and sets the `KUBECONFIG` variable to reference `admin.conf`. Afterwards, you can run `kubectl` commands against the k8s cluster.

michael-balint:
@satindern - Let's try to have commits that achieve a single goal. Notice how each time you edit and save `README.md` from the browser, you're achieving a commit? Try to quash like changes together. Also try to be more specific in the commit message for each change. Ex: "Updated README.md to k8s_environment.sh explanation."

michael-balint:
@satindern - I think the instructions can be simplified a bit - can you give me read/write access to your fork so I can make my own edits directly?

PR #224, opened 2019-04-12
Use nvcr.io version of CUDA container 

ScottESanDiego:
Consistency with the other tests (and because that's where we want everyone to go).


ScottESanDiego:
Done - figured you had your reasons for using 8 here, and didn't want to muck with it. :-)

PR #227, opened 2019-04-12
pip install required openshift package for docker registry

supertetelman:
Following the install steps on a clean CentOS 7 management node is throwing errors like this:

```
TASK [Get tiller service info] ******************************************************************************************************************************************fatal: [mgmt-]: FAILED! => {"changed": false, "msg": "This module requires the OpenShift Python client. Try `pip install openshift`"}
        to retry, use: --limit @user/repo/sa/deepops/playbooks/k8s-services.retry
```

Installing pip on the mgmt node resolves this.


playbooks/k8s-services.yml
```diff
@@ -24,6 +24,34 @@
...
+  tasks:
+    - name: install virtualenv
+      apt:
+        name: 
+          - "virtualenv"
+        state: present
+      when: ansible_distribution == 'Ubuntu'
+    - name: create location for python virtual env
+      file:
+        path: "{{ deepops_dir }}"
```
dholt: These vars need to be defined somewhere. Probably this just needs to be a role(s) but it's ok to just add a vars segment in here for now

playbooks/k8s-services.yml
```diff
@@ -24,6 +24,34 @@
...
+        state: directory
+      when: ansible_distribution == 'Ubuntu'
+    - name: install openshift python client for k8s_raw module
+      pip:
+        name: openshift
+        virtualenv: "{{ deepops_venv }}"
```
dholt: Ditto var needs to be defined

playbooks/k8s-services.yml
supertetelman: Yeah, agreed on this being a role. I'll just stick with defining the vars in place for now.

supertetelman:
I went ahead and pulled this out into a role. I haven't had a chance to test this out on both a CentOS and an Ubuntu system yet though, may want to wait on the PR if the existing GPU plugin code (that I copied from) wasn't well tested.

supertetelman:
Let me know if you think we should pull the openshift role into the gpu plugin playbook and I can test that out as part of this PR.

playbooks/k8s-services.yml
```diff
@@ -42,6 +42,10 @@
 - hosts: kube-master
   become: true
   tasks:
+    - name: Install required OpenShift packages to support Docker registry
+      include_role:
+        name: openshift
+      run_once: true
```
dholt: I think we probably need to remove the `run_once` on installing openshift since we can't guarantee which node the next `run_once` will land on. It should be ok to have those packages on every node.

dholt:
Re: 2. we can't install globally on Ubuntu, the pip packages conflict with DGX packages, hence the virtual env.

supertetelman:
Moved openshift out of k8s-gpu-plugin and added dependency. This needs re-testing on CentOS and Ubuntu multi-node systems.

roles/openshift/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+deepops_dir: /opt/deepops
+deepops_venv: '{{ deepops_dir }}/venv'
```
supertetelman: I need to verify this gets picked up in the k8s-gpu-plugin role properly.

roles/openshift/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    state: directory
+  when: ansible_distribution == 'Ubuntu'
+
+- name: install openshift python client for k8s_raw module
+  pip:
+    name: openshift
+    virtualenv: "{{ deepops_venv }}"
+  when: ansible_distribution == 'Ubuntu'
+
+  # TODO: Verify that RHEL  does not require venv
```
supertetelman: We should probably add this for consistency?

playbooks/k8s-services.yml
```diff
@@ -42,6 +42,10 @@
 - hosts: kube-master
   become: true
   tasks:
+    - name: Install required OpenShift packages to support Docker registry
```
supertetelman: I'm not sure this will help without restarting Kubernetes. I believe OpenShift needs to be installed before helm/tiller is started (as-in in the k8s-cluster.yml).

playbooks/k8s-services.yml
supertetelman: Agreed.

dholt:
What's the state of this one? I still had to make the change Balint did:

```In playsbooks/k8s-services.yml, move the Install required OpenShift packages to support Docker registry task to the first step #Install packages required for Ansible Helm module. Looks like Helm is dependent on this.```

PR #230, opened 2019-04-15
Remove vagrant hostmanager plugin

michael-balint:


virtual/virtual_inventory
```diff
@@ -21,7 +21,7 @@ virtual-mgmt
 
 [kube-node]
 virtual-mgmt
-virtual-gpu01
+virtual-gpu01 ansible_host=10.0.0.11 ip=10.0.0.11
```
dholt: Can we put all nodes/IPs in `[all]` for consistency?

michael-balint: Yep - moved them in the latest commit.

PR #233, opened 2019-04-16
Support Docker in Docker in BinderHub script

supertetelman:
Bump up the BinderHub version and enable dind to support Docker in Docker builds in default cloud environments.

### DIFF
--- a/config.example/helm/binderhub-config.yml
+++ b/config.example/helm/binderhub-config.yml
@@ -1,6 +1,7 @@
 config:
   BinderHub:
     use_registry: false
+    # image_prefix: <docker-id|organization-name>/<prefix>- # Used to name docker images if use_registry is true
     # hub_url: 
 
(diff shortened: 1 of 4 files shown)


config.example/helm/binderhub-config.yml
```diff
@@ -15,3 +15,6 @@ jupyterhub:
     enabled: true
     timeout: 3600 # Delete Pods that have been inactive for over n seconds
     every: 600 # Check for inactive Pods ever n seconds
+
+dind:
+  enabled: true
```
ajdecon: This config is pretty opaque unless you know about Docker in Docker. Can you add a comment explaining what this config and why you'd want it enabled or disabled? Either here or in the doc.

(I'd appreciate the clarification myself! I may have missed it, but a quick browse of the BinderHub docs doesn't explain why you'd want to do this...)

PR #236, opened 2019-04-16
Configure monitoring ingress with loadbalancer if present

ajdecon:
During monitoring setup, check for the presence of a LoadBalancer
Ingress on the nginx-controller service. If it exists, use the IP
address for the LoadBalancer to generate the ingress URL for the
monitoring services. If not, just use the master node IP.


ajdecon:
With these changes, I was able to successfully turn up a cluster with a MetalLB load balancer present and `curl` the URLs for the monitoring ingresses.

I'm running a local test now with the load balancer disabled to verify that it sets up monitoring with the master IP instead.

PR #240, opened 2019-04-19
[WIP] Allow building a DeepOps cluster on CentOS without Internet access

ajdecon:
This is still a WIP, but opening it as a PR to get early feedback and let others test.

This PR overrides Ansible and environment variables to allow setting up a DeepOps cluster using local software repository mirrors, rather than reaching out to the Internet. The current iteration targets CentOS, but most of this work will also translate to Ubuntu.

The default paths and ports for repos are those used in #199, which downloads dependencies and sets up mirrors. However, in theory this PR could be used independently of that one by specifying alternate mirror locations.

Current status of goals for this PR:
- [x] Turn up base Kubernetes on a virtual cluster
- [x] Set up Ceph storage
- [x] Set up Ingress
- [ ] Set up DGXie
- [ ] Set up monitoring stack
- [ ] Set up Jupyterhub and Kubeflow
- [x] Set up Slurm
- [ ] Write documentation
- [ ] Test on bare metal


ajdecon:
Added miscellaneous fixes and demonstrated a successful Slurm turnup on virtual.

ajdecon:
Note: this is now rebased on top of #199.

ajdecon:
Rebased this PR on master. Reserved merge conflicts but may need a few fixes after testing. cc @supertetelman

atetelman:
As of now this is working great end-to-end. I ran into a few testing issues while pushing Docker images to the local registry, but I believe that was related to certain iptables I had in place and lifting them resolved the issue.

There are a few manual steps that need to be included in the documentation, otherwise this seems like it is in a good enough state to merge through.

PR #245, opened 2019-04-27
Revised README and getting-started docs

michael-balint:


michael-balint:
We still need to add...
- FAQ
- slurm-usage

README.md
```diff
@@ -5,61 +5,68 @@ GPU infrastructure and automation tools
...
+* An existing cluster running Kubernetes where DeepOps scripts are used to deploy Kubeflow and connect NFS storage
+* An existing cluster that needs a resource manager / batch scheduler, where DeepOps is used to install Slurm, Kubernetes, or a hybrid of both
 
-> NOTE: we recommend using the most recent release branch for stable code.
-> The `master` branch is used for development and as such may be unstable or even broken at any point in time.
+## Releases
+
+Latest release: [DeepOps 19.03 Release](releases/tag/19.03)
+
+It is recommended to use the latest release branch for stable code (linked above). The master branch is generally stable, but is used for development.
```
ajdecon: > The master branch is generally stable, but is used for development.

Master is not really "stable", in the sense that someone who frequently updates from master might find a lot changing under them. :) Maybe re-phrase to something like,

> All development takes place on the master branch, which is generally functional but may change significantly between releases.

README.md
```diff
@@ -5,61 +5,68 @@ GPU infrastructure and automation tools
...
 
 ## Copyright and License
 
 This project is released under the [BSD 3-clause license](https://github.com/NVIDIA/deepops/blob/master/LICENSE).
 
-## Issues and Contributing
+## Issues
 
-A signed copy of the [Contributor License Agreement](https://raw.githubusercontent.com/NVIDIA/deepops/master/CLA) needs to be provided to <a href="mailto:deepops@nvidia.com">deepops@nvidia.com</a> before any change can be accepted.
+NVIDIA DGX customers should file and NVES ticket via [NVIDIA Enterprise Services](https://nvid.nvidia.com/enterpriselogin/).
```
ajdecon: > NVIDIA DGX customers should file and NVES ticket

Typo, replace with:

> NVIDIA DGX customers should file an NVES ticket ...

docs/getting-started.md
```diff
@@ -1,22 +1,40 @@
 Getting Started
 ===
 
-## What is DeepOps?
+## Requirements
 
-The DeepOps project is a reference approach for provisioning and deploying GPU servers and multi-node GPU clusters. It is intended to be adapted to unique environments and requirements while helping you go faster by sharing the experience we have with GPU servers and NVIDIA SW stacks applied to our HPC and DL needs.
+* A control system to run the install process (AKA: a provisioning node)
```
ajdecon: Referring to this as a control system confused me initially, and "system" is an overloaded term anyway. Probably best to stick to one name for this host, something like:

* A pre-existing "provisioning" node which can be used to run Ansible and the install scripts

michael-balint:
@dholt @tvincereed - are you cool with the getting-started guide?

README.md
```diff
@@ -5,61 +5,68 @@ GPU infrastructure and automation tools
 
 ## Overview
 
-The DeepOps project facilitates deployment of GPU servers and multi-node GPU clusters for Deep Learning and HPC environments, in an on-prem, optionally air-gapped datacenter or in the cloud.
+The DeepOps project encapsulates best practices in the deployment of GPU server clusters and sharing single powerful nodes (such as [NVIDIA DGX Systems](https://www.nvidia.com/en-us/data-center/dgx-systems/)). DeepOps can also be adapted or used in a modular fashion to match site-specific cluster needs. For example:
 
-Use the provided Ansible playbooks and scripts to deploy Kubernetes, Slurm, or a hybrid of both. This repository encapsulates best practices to make your life easier, but can also be adapted or used in a modular fashion to suite your specific cluster needs. For example: if your organization already has Kubernetes deployed to a cluster, you can still use the optional services and scripts provided to install Kubeflow, enable authentication, or connect NFS storage.
+* An on-prem, air-gapped data center of NVIDIA DGX servers where DeepOps provides end-to-end capabilities to set up the entire cluster management stack
```
dholt: I don't think we should exclude non-DGX servers with Tesla GPUs

README.md
```diff
@@ -5,61 +5,68 @@ GPU infrastructure and automation tools
...
+* An existing cluster running Kubernetes where DeepOps scripts are used to deploy Kubeflow and connect NFS storage
+* An existing cluster that needs a resource manager / batch scheduler, where DeepOps is used to install Slurm, Kubernetes, or a hybrid of both
```
dholt: I don't understand what 'existing cluster' means, aren't we deploying the cluster with DeepOps?

README.md
```diff
@@ -5,61 +5,68 @@ GPU infrastructure and automation tools
 
 ## Overview
 
-The DeepOps project facilitates deployment of GPU servers and multi-node GPU clusters for Deep Learning and HPC environments, in an on-prem, optionally air-gapped datacenter or in the cloud.
+The DeepOps project encapsulates best practices in the deployment of GPU server clusters and sharing single powerful nodes (such as [NVIDIA DGX Systems](https://www.nvidia.com/en-us/data-center/dgx-systems/)). DeepOps can also be adapted or used in a modular fashion to match site-specific cluster needs. For example:
```
dholt: It seems like we've removed all references to single-node/non-scheduler deployments. I feel like we need some mention somewhere that you can provision a single machine without a scheduler (this is how I use DeepOps in production every day) or use just some part of the project vs. an entire cluster deployment.

docs/getting-started.md
```diff
@@ -1,22 +1,40 @@
 Getting Started
 ===
 
-## What is DeepOps?
+## Requirements
 
-The DeepOps project is a reference approach for provisioning and deploying GPU servers and multi-node GPU clusters. It is intended to be adapted to unique environments and requirements while helping you go faster by sharing the experience we have with GPU servers and NVIDIA SW stacks applied to our HPC and DL needs.
+* A pre-existing "provisioning" node which can be used to run Ansible and the install scripts
+* A cluster to deploy to (potentially a cluster of one server or a [virtual one](/virtual/README.md))
```
dholt: Can we say something like 'cluster or single server'? 'cluster of one server' doesn't make sense

README.md
michael-balint: It's meant to be a common example - definitely not meant to exclude. Are you suggesting we be more inclusive in the example?

README.md
michael-balint: No, in this case it's that there is already a K8s cluster set up and they can pick scripts (like Kubeflow) to run on top.

README.md
michael-balint: 👍

michael-balint: Added...
```
* A single machine where no scheduler is desired, only NVIDIA drivers, Docker, and the NVIDIA Container Runtime
```

docs/getting-started.md
michael-balint: Done

PR #249, opened 2019-04-29
Use separate repo for nvidia-driver role

lukeyeager:
By using a separate repo for this code, we can upload it to the Ansible
Galaxy. I expect this code will be used much more often if we extricate
it from the deepops repo. Which will give us more users and probably
more bug reports leading to better code.


lukeyeager:
@ajdecon thoughts on these changes?

https://github.com/NVIDIA/deepops/blob/a5b8632a0f9b4bea85c7ba969e11cd3ac3a077df/roles/nvidia-driver/defaults/main.yml
vs
https://github.com/lukeyeager/ansible-role-nvidia-driver/blob/v1.0.1/defaults/main.yml

ajdecon:
@lukeyeager : The changed variable names LGTM.

My only quibble is with the comment, "Settings for airgapped systems". These vars are needed in any context where you want to use a different download mirror, whether because you're behind an air gap, or just want a local cache to improve download times. Maybe something more generic, like "repository URLs".

PR #260, opened 2019-05-10
[enroot] Add role to install from PPA

lukeyeager:
Notes:
* This PPA currently only supports Bionic - not Xenial (see https://github.com/NVIDIA/enroot/pull/4)
* No support for centos yet
* No support for airgapped systems yet

/cc @3XX0

### DIFF
--- a/playbooks/enroot.yml
+++ b/playbooks/enroot.yml
@@ -1,0 +1,7 @@
+---
+- hosts: all
+  become: true
+  tasks:
+    - name: enroot role
+      include_role:
+        name: enroot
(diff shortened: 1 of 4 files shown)


dholt:
why no centos?

3XX0:
For RHEL, you also need to tune some [kernel settings](https://github.com/NVIDIA/enroot/blob/master/doc/requirements.md) which require a reboot/kexec.  
Not sure how you do that with ansible.

dholt:
We set kernel cmd line params and reboot in other places with Ansible. I'm ok with leaving it out for now as long as we generate a to-do item for making this work for RHEL/CentOS

lukeyeager:
Will reopen later when I get it working for centos, too.

PR #264, opened 2019-05-14
[nis] enable nscd by default, too

lukeyeager:


roles/nis-client/tasks/main.yml
```diff
@@ -70,3 +70,25 @@
...
+# ncsd
+#
+
+- name: nscd service use_nscd={{ nis_client_use_nscd }}
+  service: name=nscd state=stopped
+  failed_when: False
+  when: not nis_client_use_nscd
+  tags: nscd
+
+- name: nscd packages use_nscd={{ nis_client_use_nscd }}
```
ajdecon: The name of the package is `nscd` in both distros, so you may be able to refactor this to just use the `package` module.

roles/nis-client/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+nis_client_use_nscd: yes
```
ajdecon: Should this be enabled by default? `nscd` is often an operational headache, especially on large clusters where the caches can get out of sync. I'd argue that the user should have to explicitly turn this on and know what they're getting into. :)

lukeyeager:
Nevermind. I fixed my issue by reading this commit:
https://github.com/systemd/systemd/commit/2bcbffd6db8efe8f0cc2f2b01d407a326247176d

PR #271, opened 2019-05-21
Role to run Grafana via systemd/docker

dholt:


roles/grafana/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    dest: "{{ grafana_config_dir }}/grafana.ini"
+    owner: root
+    group: root
+    mode: 0644
+  notify: restart grafana
+
+- name: install systemd unit file
+  template:
+    src: templates/docker.grafana.service.j2
+    dest: "/etc/systemd/system/{{ grafana_svc_name }}"
```
lukeyeager: Consider this path instead? `/usr/local/lib/systemd/system/`

I think of `/etc/systemd/system/` as primarily for overrides and links.

dholt: My understanding has always been that best-practice is to use `/etc/systemd/system`. I also prefer to keep configuration confined to `/etc` because it's easier to back up.

https://unix.stackexchange.com/questions/224992/where-do-i-put-my-systemd-unit-file

lukeyeager: Nice link. I'm convinced.

PR #274, opened 2019-05-21
Role to run Prometheus Slurm exporter via systemd/docker

dholt:
Container build source: https://github.com/dholt/prometheus-slurm-exporter


roles/prometheus-slurm-exporter/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+slurm_exporter_container: "dholt/prometheus-slurm-exporter"
```
michael-balint: Can we move this to the deepops repo on docker hub?

PR #278, opened 2019-05-29
[WIP] Update DGXie container to work with less custom configurization

supertetelman:
* Pass through HTTP Port to machines.json
* Added wildcards for machines.json
* Improved comments and examples
* Update for additional 4.0.6 & 4.1.0 support


config.example/containers/dgxie/dnsmasq.conf
```diff
@@ -34,3 +34,10 @@ dhcp-option=3,#GATEWAY#
 #dhcp-boot=tag:bios,"pxelinux.0"
 
 conf-dir=/etc/dnsmasq.d,*.conf
+
+# If the dhcp-ignore flag is specified in this fashion, only hosts configured with dhcp-host will be given DHCP
+dhcp-ignore=tag:!known
```
ajdecon: With this config in place, do you need to list the MACs in two places? Here and in `machines.json`?

containers/dgxie/api.py
```diff
@@ -2,19 +2,32 @@
...
+        http_port = os.environ['HTTP_PORT']
+    else:
+        http_port = "80"
+
     # return profile in json for matching machine
     for machine in machines:
-        if 'mac' in machines[machine] and machines[machine]['mac'] == mac:
+        if 'mac' in machines[machine] and re.match(machines[machine]['mac'], mac):
+            machines[machine]['mac'] = mac
+            machines[machine]['kernel'] = machines[machine]['kernel'].replace("$HTTP_PORT", http_port)
```
dholt: I don't see these vars in the machines.json file, it still has hard-coded ports

containers/dgxie/api.py
```diff
@@ -2,19 +2,32 @@
...
+    '''See https://github.com/danderson/netboot/blob/master/pixiecore/README.api.md'''
     # load machine profiles for each call so we can re-load changes from disk
     jf = open('/etc/machines/machines.json', 'r')
     machines = json.load(jf)
     jf.close()
 
+    if "HTTP_PORT" in os.environ.keys():
+        http_port = os.environ['HTTP_PORT']
+    else:
+        http_port = "80"
```
dholt: we should default to the same value as the start file, and not use port 80

scripts/build_and_restart_dgxie.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -xe
+
+
+compose_file=config/containers/dgxie/docker-compose-dgxie.yml
+compose_directory="."
+compose_cmd="docker-compose --project-directory ${compose_directory} -f ${compose_file}"
```
dholt: I still think this adds unnecessary complexity, can we just leave the file called `docker-compose.yml` and avoid the script?

scripts/build_and_restart_dgxie.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
dholt: This script should be executable

config.example/containers/dgxie/docker-compose-dgxie.yml
```diff
@@ -20,11 +23,3 @@ services:
             #- DNS2=8.8.4.4
             #- DHCP_START=192.168.1.2
             #- DHCP_END=192.168.254
```
dholt: Missing an octet in the range, should be `192.168.1.254`

config.example/containers/dgxie/dnsmasq.conf
dholt: This is what assigns static dhcp leases, machines.json controls what 'profile' the pxe server assigns to the machine.
I don't think we should default to having this un-commented, because nothing will boot by default. Either way we need documentation on this.

containers/dgxie/api.py
```diff
@@ -2,19 +2,32 @@
...
+
     # return profile in json for matching machine
     for machine in machines:
-        if 'mac' in machines[machine] and machines[machine]['mac'] == mac:
+        if 'mac' in machines[machine] and re.match(machines[machine]['mac'], mac):
+            machines[machine]['mac'] = mac
+            machines[machine]['kernel'] = machines[machine]['kernel'].replace("$HTTP_PORT", http_port)
+            machines[machine]['cmdline'] = machines[machine]['cmdline'].replace("$HTTP_PORT", http_port)
```
dholt: According to the API spec (https://github.com/danderson/netboot/blob/master/pixiecore/README.api.md) only the kernel is a required parameter, but this requires cmdline and initrd to be specified or the API server fails

config.example/containers/dgxie/dnsmasq.conf
supertetelman: The logic behind leaving this here was so that DGXie would only start giving addresses out if it had been configured correctly. This was to prevent someone accidentally standing up multiple DHCP servers in their environment. Something worth documenting.

supertetelman:
Closing this PR. All these changes were merged into the pxe-vmware branch and we can wrap up the work there.

https://github.com/NVIDIA/deepops/pull/286

PR #284, opened 2019-06-03
deepops namespace

ryanolson:
Move DeepOps services that were previously installed under `default` to `deepops`.

TODOs
[ ] ENV override for `deepops` if you wish to customize the namespace used.
[ ] add other services that previously got deployed under default


michael-balint:
@ryanolson - Can we close this out for a release (in the next few days)? Most of the other services we deploy are in their own namespace (ex: kubeflow). Also, the ENV for the namespace override isn't critical IMO. We can forgo that in the interest of closing this out.

dholt:
Please create a branch on your fork in the future as well

PR #292, opened 2019-06-11
Add NGC-Ready playbook and doc

dholt:
Also adds default config to render human-readable Ansible output.

### DIFF
--- a/docs/ngc-ready.md
+++ b/docs/ngc-ready.md
@@ -1,0 +1,52 @@
+NGC Ready Server
+===
+
+NVIDIA® GPU Cloud (NGC) containers leverage the power of NGC-Ready servers with NVIDIA GPUs. This document describes how to set up your NGC-Ready server with a software stack optimized to run NGC containers.
+
+## Prerequisites
+
+These instructions assume the following:
+
+  * You have a NGC-Ready server. To determine if your server is NGC-Ready, please review the list of validated servers at the NGC-Ready Server documentation page - https://docs.nvidia.com/ngc/ngc-ready-systems/index.html
+  * Your NGC-Ready Server has a compatible Linux distribution installed:
+    * Ubuntu Server 18.04 LTS
+    * RHEL/CentOS 7
+
+## Setup
+
+NGC-Ready Server setup can be done from the NGC-Ready server itself or a 'control' machine such as a laptop with SSH access to the NGC-Ready Server.
+
+```sh
+# Clone code repo and install pre-requisite software
+git clone https://github.com/NVIDIA/deepops.git
+cd deepops
+./scripts/setup.sh
+```
+
+## Installation
+
+This process will install the latest NVIDIA GPU Drivers, and Docker with the NVIDIA container runtime.
+
+```sh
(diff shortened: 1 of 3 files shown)


docs/ngc-ready.md
```diff
@@ -0,0 +1,52 @@
...
+This process will install the latest NVIDIA GPU Drivers, and Docker with the NVIDIA container runtime.
+
+```sh
+# Install NGC-Ready software stack
+# where:
+#   <ssh-user>: SSH username to reach NGC-Ready server
+#   <ip-of-host>: IP of NGC-Ready server, or localhost. The trailing comma is required
+# If SSH requires a password, add: -k
+# If sudo requires a password, add: -K
+ansible-playbook -u <ssh-user> -i <ip-of-host>, playbooks/ngc-ready.yml
```
michael-balint: `-i` should refer to inventory file, not ip. Might want to include `-l` if specifying limitation by host.

docs/ngc-ready.md
```diff
@@ -0,0 +1,52 @@
...
+This process will test the functionality of the NGC-Ready server by running a functional test and two deep learning framework container tests from the NVIDIA Container Registry.
+
+```sh
+# Run NGC-Ready tests
+# where:
+#   <ssh-user>: SSH username to reach NGC-Ready server
+#   <ip-of-host>: IP of NGC-Ready server, or localhost. The trailing comma is required
+# If SSH requires a password, add: -k
+# If sudo requires a password, add: -K
+ansible-playbook -u <ssh-user> -i <ip-of-host>, playbooks/ngc-ready.yml --tags test
```
michael-balint: Same as above.

playbooks/ngc-ready.yml
```diff
@@ -0,0 +1,137 @@
...
+        pull: yes
+      tags:
+        - test
+        - never
+    - name: test pytorch
+      docker_container:
+        name: pytorch
+        image: "nvcr.io/nvidia/pytorch:{{ ngc_container_tag }}"
+        working_dir: /opt/pytorch/examples/upstream/mnist
+        detach: no
```
michael-balint: Would it hurt to add the following? Otherwise it won't work on virtual clusters, for example.

```
network_mode: host
```

michael-balint:
Overall comment: the tests are useful, but isn't this just doing what we do with some of our other ansible playbooks? What is the value add?

docs/ngc-ready.md
dholt: My impression was that this playbook would be used for one-off machines, so this was an attempt to simplify things and not force people to edit the inventory file. This is how I always use the standalone playbook as well.

playbooks/ngc-ready.yml
dholt: The tests don't require any networking so it's probably not necessary

michael-balint: They do for some - for example the pytorch one fails if you can't download the MNIST dataset.

dholt: ok, done

PR #294, opened 2019-06-20
Add NGC containers to Kubeflow

supertetelman:
Still be tested/cleaned up.

This adds a script that will update the Kubeflow config.yaml file and other templates files to include NGC containers, etc.

Future updates will include NGC style changes.

### DIFF
--- a/scripts/k8s_deploy_kubeflow.sh
+++ b/scripts/k8s_deploy_kubeflow.sh
@@ -60,6 +60,9 @@ if [ ! -d ${KUBEFLOW_SRC} ] ; then
     sudo mv ${tempd} ${KUBEFLOW_SRC}
 fi
 
+# Update the Kubeflow Jupyter UI
+./scripts/update_kubeflow_config.py
+
 # Get master ip
 master_ip=$(kubectl get nodes -l node-role.kubernetes.io/master= --no-headers -o custom-columns=IP:.status.addresses.*.address | cut -f1 -d, | head -1)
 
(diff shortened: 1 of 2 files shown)


scripts/k8s_deploy_kubeflow.sh
```diff
@@ -76,6 +79,10 @@ fi
 pushd ${HOME}
 ${KUBEFLOW_SRC}/scripts/kfctl.sh init ${KFAPP} --platform none
 cd ${KFAPP}
+
+# Update the Kubeflow Jupyter UI
+./scripts/update_kubeflow_config.py
```
michael-balint: Why do we do this twice?

supertetelman:
I added a missing line that was causing some failures and removed the now unnecessary extra script execution. I tested this a few times and it is now populating with NGC containers.

There are still two improvements we should probably make to this. These shouldn't be blocking for this PR and instead are future improvements, but let me know what you think.

1. Only show the first 3-10 tags for each container (currently the list is very long
2. Pre-populate the extra-resources with nvidia.com/gpu values instead of giving an example in the help message

supertetelman:
I added a missing line that was causing some failures and removed the now unnecessary extra script execution. I tested this a few times and it is now populating with NGC containers.

There are still two improvements we should probably make to this. These shouldn't be blocking for this PR and instead are future improvements, but let me know what you think.

1. Only show the first 3-10 tags for each container (currently the list is very long
2. Pre-populate the extra-resources with nvidia.com/gpu values instead of giving an example in the help message

supertetelman:
Does thee `kfctl.sh apply k8s` command not update the config map when it runs `kubectl apply --validate=false -f default.yaml`? I guess I assumed that running that would properly update Kubeflow. I'll dig into that a bit more and see if that is the issue.

I'll also push a few additional fixes/enhancements in here now based on other feedback.

michael-balint:
@supertetelman - No, running `/opt/kubeflow/scripts/kfctl.sh apply k8s` doesn't appear to update. Still get the old images.

PR #295, opened 2019-06-24
Add playbook to install pre-reqs for Rook

supertetelman:
- Addressing a bug that was causing Rook installation to fail on some systems.
- Created a new bootstrap-software playbook for us to add misc. library installs that don't fit under existing bootstrap playbooks.


playbooks/bootstrap-software.yml
```diff
@@ -0,0 +1,12 @@
...
+  become: true
+  tasks:
+    - name: install xfsprogs for ceph/rook
+      package:
+        name: xfsprogs
+        state: present
+    - name: install python-setuptools for ceph/rook
+      package:
+        name: python-setuptools
+        state: present
```
ryanolson: present or latest?

ryanolson: same with pip.  often times we get warnings when pip is too out of date.  perhaps we can have a task that ensures both `pip` and `setuptools` are up-to-date.

supertetelman: I wanted to leave this as present. A few of the other playbooks (I'm thinking nvidia-dgx specifically) install a tested/validated version of python-setuptools. We don't need latest for ceph/rook to work, we just need the library present. I'd rather not upgrade/risk breaking something else if it isn't necessary.

supertetelman:
Well, what are your thoughts on creating a general pre-req software playbook? I originally had it called boot-rook.yml, but I thought it would be more useful to create a more general file for packages that typically should be present, but are missing on some more bare-bones systems.

dholt:
It makes more sense to me to have the prerequisites installed with their respective software. If someone doesn't want Rook they shouldn't have to install the prerequisites.

playbooks/bootstrap-software.yml
dholt: +1 'present' is best-practice for Ansible anyway, I think the idea is that you don't want to unknowingly update user's packages

supertetelman:
Okay, I moved files into the bootstrap-rook.yml and updated the deployment script to hint at running this playbook or properly removing previous installs if it is failing.

michael-balint:
@supertetelman - I don't think moving the check/install for these dependencies in the k8s-cluster playbook is what we want. Can we move it under the rook bash script? I know it's gross, but the rook bash script should probably become a playbook as well, eventually.

Also (this doesn't pertain to this PR, but is relevant): the `scripts/rmrook.sh` script doesn't ever work for me. I constantly need to change the last line. Shouldn't we be running the ansible command on "all" hosts, not just "management"? Management doesn't exist in our default inventories. Should we update this in this PR?

PR #304, opened 2019-07-15
Add playbook to manage ad-hoc startup commands via rc.local

dholt:


playbooks/rc-local.yml
```diff
@@ -0,0 +1,30 @@
...
+        mode: 0755
+        create: yes
+      when: ansible_distribution == "Ubuntu"
+    - name: manage /etc/rc.d/rc.local permissions on RHEL/CentOS
+      file:
+        path: /etc/rc.d/rc.local
+        mode: 0755
+      when: ansible_os_family == "RedHat"
+    - name: manage /etc/rc.local contents
+      blockinfile:
```
ajdecon: Are you doing it this way to try to preserve pre-existing rc.local contents?

Otherwise it would seem like a template would be an easier way to manage this.

dholt: Yes exactly. We could probably still preserve file contents with a template but this was easy and admittedly I needed a quick/easy solution :)

ajdecon:
Tests pass, merging.

PR #306, opened 2019-07-16
Add Enroot role/playbook

dholt:
Ansible role and playbook to install [Enroot](https://github.com/NVIDIA/enroot)

### DIFF
--- a/playbooks/enroot.yml
+++ b/playbooks/enroot.yml
@@ -1,0 +1,9 @@
+---
+- import_playbook: nvidia-docker.yml
+
+- hosts: all
+  become: yes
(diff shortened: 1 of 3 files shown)


ajdecon:
Because `enroot` is a fairly non-obvious name, can we either add a comment to the playbook, or a README to the role? (Or both! ;-) ) So that it's clear from the code what this tool is and why you would want it.

michael-balint:
@ajdecon - That's a great suggestion. We'll likely have this in the docs since this is part of an upcoming deployment stack.

dholt:
Depricated in favor of #337

PR #311, opened 2019-07-19
Support for MAAS automation

dholt:


michael-balint:
Tested, works well, but merge conflicts must be resolved.

Suggest adding a line to the config.example/inventory for `maas` (can be empty).

Issue #312, opened 2019-07-22
Failed at  "TASK [slurm : create users]"


jxdn:
Hi Team,

I face this issue when i run :
ansible-playbook -l slurm-cluster playbooks/slurm-cluster.yml 

TASK [slurm : create users] ****************************************************
fatal: [login01]: FAILED! => 
  msg: '''dict object'' has no attribute ''SUDO_USER'''

Any idea why this issue happen?

supertetelman:
I'll take a look at this.

For now you can get past this by adding the '-u ubuntu' flag to the ansible command.

PR #315, opened 2019-07-24
Update user params in slurm files

supertetelman:
This resolves the issue where a -u flag was required when install slurm.

### DIFF
--- a/playbooks/slurm.yml
+++ b/playbooks/slurm.yml
@@ -9,7 +9,7 @@
   roles:
     - name: slurm
       vars:
-        user: "{{ ansible_user | default(ansible_env.SUDO_USER) }}"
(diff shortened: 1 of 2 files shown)


supertetelman:
As far as I can tell this was always failing and the ansible_ssh_user variable has been deprecated. Let me know if there were other reason we used that value here.

dholt:
Which part was always failing? I use this playbook with/without `-u` all the time.

supertetelman:
Maybe you have a less vanilla configuration. I've seen a handful of different people, including a currently open issue. I've experienced it myself on a clean Ubuntu 18 install.

dholt:
I’m asking which part you are referring to in your second comment

dholt:
#312

PR #319, opened 2019-07-30
Split Jenkinsfile into function-specific scripts

ajdecon:
Take most shell commands out of the Jenkinsfile and split into function-specific scripts.

This has the dual purpose of making the Jenkinsfile more readable, and eventually making it easier to write tests in nicer languages.

Note: this is a WIP because it can only be tested by opening a PR. ;-P Will ping reviewers directly once it passes tests and is ready to go.

### DIFF
--- a/.jenkins-scripts/munge-files.sh
+++ b/.jenkins-scripts/munge-files.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+if [ -z "${GPUDATA+x}" ]; then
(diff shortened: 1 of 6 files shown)


ajdecon:
Jenkins tests pass, and looking at the results it looks like everything actually ran. Since this PR is all about Jenkins tests, that seems right to me!

Ping @dholt @michael-balint for review, this should be good to go.

PR #338, opened 2019-09-06
Update main.yml

nvhans:
URL for dholt's SLURM RPM fiile is incorrect.


michael-balint:
Verified that the new link works. We have RPM assets under https://github.com/DeepOps/slurm/releases/tag/1.4, but not docker-1.4

PR #344, opened 2019-09-26
Enabling proxy use for installation - initial commit.

phogan-nvidia:
WORK IN PROGRESS - Do NOT merge

working on enabling proxies and need some final testing.


config.example/group_vars/all.yml
```diff
@@ -1,6 +1,17 @@
...
+
+# Proxy Variables needed for ansible etc.
+#HTTP_PROXY
+#HTTPS_PROXY
+#NO_PROXY
+
+#PROXY_ENV
+    #HTTP_PROXY
+    #HTTPS_PROXY
+    #NO_PROXY
```
ajdecon: Instead of using these comments as a target for the `sed` in `scripts/setup.sh`, let's just have the user duplicate the configuration directly. I.e., replace this text with:

```
# Configure these variables to use an HTTP proxy
# no_proxy should be used for internal Kubernetes addresses that should not be proxied.
# Values below show an example config.

# proxy_env:
#     http_proxy: "http://10.0.2.5:3128"
#     https_proxy: "http://10.0.2.5:3128"
#     no_proxy: "localhost,cluster.local,127.0.0.1,::1,10.0.2.10,10.0.2.20,10.0.2.30" 

```

It's a little extra work when for the user when configuring DeepOps, but it's less fragile against the user making other configuration changes or using an alternate path for their configuration directory.

playbooks/bootstrap-python.yml
```diff
@@ -4,7 +4,14 @@
...
     # Fortunately, python is a dependency of yum. So RedHat systems always have it.
-    - name: install python
+    - name: install python (no proxy)
       raw: test -e /usr/bin/python || (apt -y update && apt install -y python-minimal)
       register: output
       changed_when: output.stdout != ""
+      when: proxy_env is not defined
+
+    - name: install python (using proxy)
+      raw: test -e /usr/bin/python || (https_proxy="{{ https_proxy }}" http_proxy="{{ http_proxy }}" apt -y update && https_proxy="{{ https_proxy }}" http_proxy="{{ http_proxy }}" apt install -y python-minimal)
```
ajdecon: Can we use `proxy_env['http_proxy']` and `proxy_env['https_proxy']` here instead of the named variables on their own? This will avoid the duplication in the group_vars file.

playbooks/k8s-cluster.yml
```diff
@@ -31,6 +31,8 @@
 
 # Install Kubernetes
 # for configuration, see: config/group_vars/k8s-cluster.yml
+## 
+## PHogan
```
ajdecon: Not clear why this comment is here, can we remove it?

requirements.yml
```diff
@@ -16,11 +16,19 @@
...
 
-- src: nvidia.nvidia_docker
-  version: "v1.1.0"
+- src: https://github.com/phogan-nvidia/ansible-role-nvidia-driver
+  name: "nvidia.nvidia_driver"
+  version: "proxy"
+
+- src: https://github.com/phogan-nvidia/ansible-role-nvidia-docker
+  name: "nvidia.nvidia_docker"
+  version: "proxy"
```
ajdecon: Before merging this into DeepOps we should merge in the PRs for these roles. I'll review these as well.

roles/docker/tasks/main.yml
```diff
@@ -11,6 +11,9 @@
     state: "{{ docker_package_state }}"
   notify: restart docker
 
+## - import_tasks: docker-proxy.yml
+##   when: proxy_env is defined
+
```
ajdecon: Clean up commented code

roles/k8s-gpu-plugin/tasks/main.yml
```diff
@@ -19,24 +20,37 @@
...
+    url: "{{ nvidia_k8s_device_plugin_def }}"
+    dest: "{{ nvidia_k8s_device_plugin_file }}"
+    mode: 0755
+  environment: "{{proxy_env if proxy_env is defined else {}}}"
 
 - name: create GPU device plugin
   k8s:
     state: present
-    definition: "{{ lookup('url', nvidia_k8s_device_plugin_def, split_lines=False) }}"
+    definition: "{{ lookup('file', nvidia_k8s_device_plugin_file, split_lines=False) }}"
```
ajdecon: This task failed in my test because the `lookup` runs on the Ansible control machine, not on the remote machine it's configuring. If performing the `get_url` on the remote machine, you need to separately register the contents of the file you downloaded, rather than using a lookup.

I was able to run successfully in my test using the following change:

```
--- a/roles/k8s-gpu-plugin/tasks/main.yml
+++ b/roles/k8s-gpu-plugin/tasks/main.yml
@@ -35,10 +35,14 @@
     mode: 0755
   environment: "{{proxy_env if proxy_env is defined else {}}}"

+- name: get contents of device plugin definition
+  command: "cat {{ nvidia_k8s_device_plugin_file }}"
+  register: nvidia_k8s_device_plugin_contents
+
 - name: create GPU device plugin
   k8s:
     state: present
-    definition: "{{ lookup('file', nvidia_k8s_device_plugin_file, split_lines=False) }}"
+    definition: "{{ nvidia_k8s_device_plugin_contents.stdout }}"
     ## definition: "{{ lookup('url', nvidia_k8s_device_plugin_def, split_lines=False) }}"
   run_once: true
   environment:
```

scripts/proxy.sh
```diff
@@ -0,0 +1,5 @@
+# edit the proxy details
+#
+# export http_proxy="http://10.0.2.5:3128"
+# export https_proxy="http://10.0.2.5:3128"
+# export no_proxy="localhost,cluster.local,127.0.0.1,::1,10.0.2.10,10.0.2.20,10.0.2.30"
```
ajdecon: This is configuration rather than a script we want to execute to deploy some software. Let's move it into `config.example/` rather than `scripts/`.

scripts/setup.sh
```diff
@@ -14,27 +14,35 @@ cd "${SCRIPT_DIR}/.." || echo "Could not cd to repository root"
 # Pinned Ansible version
 ANSIBLE_OK="2.7.8"
 ANSIBLE_VERSION="2.7.11"
+ANSIBLE_REPO_VER="2.7"
+
+as_sudo(){
+   cmd="sudo bash -c '. ${SCRIPT_DIR}/proxy.sh && $1'"
+   ## echo $cmd
+   eval $cmd
+}
```
ajdecon: Why use this rather than just sourcing the proxy variables globally in the environment?

scripts/setup.sh
```diff
@@ -190,3 +202,15 @@ if [ ! -d "${CONFIG_DIR}" ] ; then
...
+# update the config/group_vars/all.yml file for proxy config
+if [ `grep -v ^# ${SCRIPT_DIR}/proxy.sh | grep -v ^$ | wc -l` -gt 0 ]; then
+    o=config.example/group_vars/all.yml
+    t=config/group_vars/all.yml 
+    echo "Updating [$t] to use proxies"
+    . ${SCRIPT_DIR}/proxy.sh && cat $o | sed -e "s|#PROXY_ENV|proxy_env:|g" \
+	                                    -e "s|#NO_PROXY|no_proxy: $no_proxy|g" \
+	                                    -e "s|#HTTP_PROXY|http_proxy: $http_proxy|g" \
+	                                    -e "s|#HTTPS_PROXY|https_proxy: $https_proxy|g" > $t
+fi
```
ajdecon: As in my comment for `config.example/group_vars/all.yml`: This seems pretty fragile, and will break if the user accidentally changes the comments in the config file or uses a different config path. The config directory is intended to be user-editable, so this seems likely to happen not infrequently.

Instead, we should just ask the user to configure proxies in both the environment shell script and in the Ansible group vars. It represents some duplication, but I think it's less likely to break, especially since proxies are usually configured once and then left alone.

requirements.yml
ajdecon: Ok, changes in the role repositories are merged in! Instead of pointing to your github, we can point to the v1.2.0 releases for each of these roles.

config.example/group_vars/all.yml
phogan-nvidia: The premise of using the proxy.sh script was to minimize the places for users to make edits and therefore make mistakes. setup.sh needs drives the config file changes like the other and unfortunately not all ansible commands respect the group_var's variable. so the config file is setup for the user but they're free to make changes as necessary.

playbooks/bootstrap-python.yml
phogan-nvidia: unfortunately, the ansible command does not respect the group_vars and this was the only workaround I was able to find.

playbooks/k8s-cluster.yml
phogan-nvidia: Its not needed

requirements.yml
phogan-nvidia: Great.

scripts/proxy.sh
phogan-nvidia: It can't be moved since setup.sh requires the use of proxyies as well to install software.

scripts/setup.sh
phogan-nvidia: global variables did not work for all commands executed on local host as well as remote - best option found.

scripts/setup.sh
phogan-nvidia: best to keep its as simple as possible and the reason for keeping edits to a single file thus using proxy.sh. the setup.sh script installs software and creates the initial contents of config dir, asking users to edits in multiple locations leads to more confusion rather than less. it's the same as having to edit the inventory file to make the config applicable to their env. In this case if they want difference proxies for setup.sh and software install from setup.sh and the ansible scripts they'll need to update group_vars. The likelihood of that in an enterprise is extremely low.

michael-balint:
Closing in favor of a new PR, which has merged changes from deepops master.

PR #346, opened 2019-09-27
Kubeflow 6 2

supertetelman:
* Add Kubeflow Readme
* Bump Kubeflow up to version v0.6.2
* Cleanup Kubeflow deployment script with cmd line arguments for delete, install, config override, and print_url
* Add deployment option for "community version" and vendor version with Dex pre-installed
* Add minimal testing harness
* Fix for nvidia-dgx role

**Note:** Upgrading to this version of Kubeflow breaks NGC compatability. Until we have revisity the custom jupyter-web-app and kubeflow_update python script NGC images will not display in the image selection drop down, there will be no custom run commands to support NGC containers, and the default custom resource field will not have the nvidia gpu config added by default.

**Note:** The -D delete flag may leave around a few configmaps or persistant volumes created through Kubeflow use.

### DIFF
--- a/docs/kubeflow.md
+++ b/docs/kubeflow.md
@@ -1,0 +1,30 @@
+# Kubeflow
+
+[Kubeflow](https://www.kubeflow.org/docs/) is a K8S native tool that eases the Deep Learning and Machine learning lifecycle.
+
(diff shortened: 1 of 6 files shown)


supertetelman:
To avoid upsetting people who were depending on NGC/Kubeflow integration and to make the people who want v0.6.2 happy I have added the old scripts back, marked them as deprecated, and added them to the docs.

I think with these changes we should be good to pull this through after some re-testing of the old scripts.

Issue #305, opened 2019-07-15
"default-shm-size": "1G" conflicts with DGX BaseOS setting


bulldude:
On the DGX BaseOS (tested on 4.1.0) default-shm-size": "1G" set in /etc/systemd/system/docker.service.d/docker-override.conf conflicts with the DeepOps value set in /etc/docker/daemon.json.  

The conflict stops the node from participating in GPU jobs (CPU/MEM still work).   Removing from either and restarting kubelet and docker.service resolves the issue.

Without the change you can run a docker GPU job locally on the DGX, but cannot launch any jobs with kubectl.

dholt:
Should be fixed in https://github.com/NVIDIA/deepops/pull/351

PR #351, opened 2019-10-03
Don't run docker or nvidia-docker roles on DGX

dholt:
DGX OS has its own docker and nvidia-docker setup and we can run into problems if we install over them.

Addresses https://github.com/NVIDIA/deepops/issues/305

### DIFF
--- a/playbooks/docker.yml
+++ b/playbooks/docker.yml
@@ -26,7 +26,9 @@
         # matches: kubespray/roles/container-engine/docker/vars/ubuntu-amd64.yml
         #docker_package: 'docker-ce=18.06.1~ce~3-0~ubuntu'
         docker_package: 'docker-ce=5:18.09.5~3-0~ubuntu-{{ ansible_distribution_release }}'
-      when: ansible_distribution == "Ubuntu"
+      when:
+        - ansible_distribution == "Ubuntu"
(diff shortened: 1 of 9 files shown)


dholt:
I don't know if this is the best solution. Open to suggestions

dholt:
Tested and confirmed this works with:

```
ansible-playbook -l boss-mule -e ansible_product_name=DGX-1 playbooks/nvidia-dgx.yml
ansible-playbook -l boss-mule -e ansible_product_name=DGX-1 playbooks/nvidia-driver.yml
ansible-playbook -l boss-mule -e ansible_product_name=DGX-1 playbooks/nvidia-docker.yml
ansible-playbook -l boss-mule -e ansible_product_name=DGX-1 playbooks/docker.yml
```

lukeyeager:
I've run into this, too. The problem is that DGX configures docker options using systemd files instead of `/etc/docker/daemon.json`:
```console
$ grep dockerd /etc/systemd/system/docker.service.d/docker-override.conf 
ExecStart=/usr/bin/dockerd -H fd:// -s overlay2 --default-shm-size=1G
```
... and Docker (reasonably?) crashes when you try to specify options both there and in the JSON file.

Neither the systemd file approach or the single JSON file approach are very good, IMO. Docker should be using some sort of `parts.d/` dropin file setup to allow multiple packages to configure the daemon.

PR #356, opened 2019-10-22
Add Chrony NTP client playbook, role, and vars

supertetelman:


playbooks/chrony-client.yml
```diff
@@ -0,0 +1,10 @@
+---
```
michael-balint: Add a comment explaining what this does.

config.example/group_vars/all.yml
```diff
@@ -9,6 +9,9 @@
 #nis_domain: example.com
 #nis_server: 10.0.0.1
 
+# NTP configuration
+#chrony_timezone: "Asia/Dubai"
```
michael-balint: `America/Los_Angeles` as default

supertetelman:
Go ahead and merge.

PR #357, opened 2019-10-22
EasyBuild and lmod playbooks for better HPC support

supertetelman:
* Add lmod playbook/role
* Add Easybuild playbook/role


roles/easy-build-packages/defaults/main.yml
```diff
@@ -0,0 +1,15 @@
+is_head: no
+is_compute: no
+
+eb_packages_path: "{{ eb_prefix_path}}/ebfiles_repo"
+eb_packages:
```
dholt: Should use an empty list here `[]` and move the defaults to `config/group_vars/`

roles/easy-build/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+eb_tmp_dir: /tmp/easybuild
+eb_bootstrap_url: 'https://raw.githubusercontent.com/easybuilders/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py'
```
dholt: Should use some version of this install script vs whatever is the latest

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+#     force: yes
+#     recurse: yes
+#     mode: 'u=rwx,g=rwx,o=rx' 
+#   when: item.stat.exists == false
+#   with_items:
+#     - "{{ folder_stats.results }}"
+---
+- name: "install needed packages"
+  become: yes
+  action: apt name=python-setuptools state=latest
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+  when: ansible_os_family == "Debian"
+
+- name: "install needed packages"
+  become: yes
+  action: yum name=python-setuptools state=latest
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+  when: ansible_os_family == "RedHat"
+
+- name: "rm previous tmp folders"    
+  file: 
+    path: "{{ eb_tmp_dir }}"
+    state: absent
+    force: yes
+
+- name: "rm EasyBuild software"
+  file: path="{{ eb_software_path }}/EasyBuild" state=absent force=yes
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+
+- name: "rm EasyBuild profile file"
+  become: yes
+  file: path="/etc/profile.d/z01_eb.sh" state=absent force=yes
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+  when: ansible_os_family == "Debian"
+
+- name: "rm EasyBuild profile file"
+  become: yes
+  file: path="/etc/profile.d/z01_eb.csh" state=absent force=yes
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+  when: ansible_os_family == "RedHat"
+
+- name: "rm EasyBuild module"
+  file: path={{ eb_modulepath }}/EasyBuild state=absent force=yes
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+
+- name: "mkdir required directories"
+  file: path={{ item }} state=directory recurse=yes #mode='u=rwx,g=rwx,o=rx' 
+  with_items:
+    - "{{ eb_modulepath }}"
+    - "{{ eb_software_path }}"
+  #delegate_to: 127.0.0.1
+
+- name: "download bootstrap script"
+  get_url: url="{{ item.url }}" dest="{{ item.dest }}" mode='u=rwx,g=rwx,o=rx' force=yes
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+# Ansible doesn't run task in a login shell, so we have to source the
+# EasyBuild config manually. We might want to create some abstraction for this.
+# 
+- name: "install latest EasyBuild version"
+  shell: . /etc/profile && python /tmp/bootstrap_eb.py {{ eb_tmp_dir }}
+  args:
+    executable: /bin/bash
+
+- name: "stat EasyBuild software"
+  stat: path="{{ eb_tmp_dir }}/software/EasyBuild"
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+  with_items:
+    - regexp: '^local root .*$'
+      line: 'local root = "{{ eb_software_path }}/EasyBuild/{{ ebsw_version.stdout }}"'
+
+# - name: "check EasyBuild software"
+#   stat: path="{{ eb_software_path }}/EasyBuild/{{ ebsw_version.stdout }}"
+#   register: swpath_stat
+
+- name: "cp EasyBuild software"
+  command: cp -R "{{ eb_tmp_dir }}/software/EasyBuild" "{{ eb_software_path }}"
```
dholt: Should use copy or synchronize built-ins

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+#  when: ebsw_stat.stat.exists
+
+# - name: "check EasyBuild module"
+#   stat: path="{{ eb_modulepath }}/{{ ebsw_version.stdout }}.lua"
+#   register: mdpath_stat
+
+- name: "cp EasyBuild module"
+  command: cp -R "{{ eb_tmp_dir }}/modules/all/EasyBuild" "{{ eb_modulepath }}"
```
dholt: Should use copy or synchronize built-ins

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,145 @@
...
+#  when: ebsw_stat.stat.exists
+
+- name: "rm EasyBuild tmp directories"
+  file: path={{ eb_tmp_dir }} state=absent force=yes
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/lmod/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
+#
+# ansible role to install lmod:
+#
+---
+- name: "install packages"
+  become: yes
+  action: apt name=lmod state=latest
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/lmod/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+  when: ansible_os_family == "Debian"
+
+- name: "install packages"
+  become: yes
+  action: yum name=lmod state=latest
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/lmod/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+  when: ansible_os_family == "RedHat"
+
+- name: "mkdir software path"
+  file: path={{ eb_software_path }} state=directory
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

roles/lmod/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+
+- name: "mkdir module path"
+  file: path={{ eb_modulepath }} state=directory
```
dholt: Should use the extended format to be consistent with our other roles and the other tasks in this role

config.example/group_vars/all.yml
```diff
@@ -50,6 +50,13 @@
 
 # Lmod modules
 #lmod_modulepath: /shared/modules/all
+#
+
+# EasyBuild path configuration
+eb_prefix_path: /mnt/software
```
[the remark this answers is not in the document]
supertetelman: `/mnt` should exist on all systems.

The purpose of this playbook is to set up EasyBuild modules that can be put into a shared NFS directory and used by Slurm/etc. across a cluster. I think it makes more sense to default this to a `/mnt` directory but I'll let @issamsaid  comment on this as most of the playbook is his design.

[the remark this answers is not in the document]
dholt: `/mnt` is for mounted filesystems; for standards, see: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

We set `deepops_dir: /opt/deepops` in `config.example/group_vars/all.yml`. It seems like a good idea to re-use that variable/location by default, or you could pick something like `/shared`, `/export`, etc. It should probably also use a similar default to `lmod_modulepath` above.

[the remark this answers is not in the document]
supertetelman: My understanding here is that this path was supposed to be a mounted file system. The idea here was to build to modules once and then they can be loaded identically on all nodes in the slurm cluster. We've used an NFS mount everytime we have used this playbook.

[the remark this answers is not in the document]
dholt: That's fine, that's how you would normally handle modules, but you can't assume `/mnt/software` exists and nothing in the PR creates it or checks for it. If I run this out-of-the-box it fails.

playbooks/lmod.yml
```diff
@@ -0,0 +1,11 @@
+#
+# playbook to install the lmod tool to manage software packages/modules: 
+#
+---
+- hosts: slurm-master
```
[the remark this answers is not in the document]
supertetelman: @issamsaid, was there a reason we did this? I recall that there was some ordering/timing issue that we hit without the seperation here.

supertetelman:
See https://github.com/NVIDIA/deepops/pull/441

issamsaid:
@supertetelman  is this merged?

PR #358, opened 2019-10-22
Playbook to set/reset gpu clock speed

supertetelman:
Create a playbook that will set or reset the GPU clock speeds.

After pulling this through and doing a playbook re-org we will probably want to cleanup how the variables are used here and put this into a bigger/more robust GPU configuration playbook.

### DIFF
--- a/playbooks/set-gpu-clocks.yml
+++ b/playbooks/set-gpu-clocks.yml
@@ -1,6 +1,6 @@
 ---
 # set gpu clocks on all worker nodes
-- hosts: slurm-node
+- hosts: compute-node
   gather_facts: no
   become: yes
   vars:
--- a/playbooks/set-gpu-clocks.yml
+++ b/playbooks/set-gpu-clocks.yml
@@ -1,0 +1,16 @@
+---
+# set gpu clocks on all worker nodes
+- hosts: slurm-node
+  gather_facts: no
+  become: yes
+  vars:
+    gpu_clock_reset: yes
+    gpu_clock: "1507,1507"
+  tasks:
+    - name: set the gpu clock to a specified amount
+      shell: "nvidia-smi -lgc {{ gpu_clock }}"
+      when: not gpu_clock_reset
+
+    - name: reset the gpu clock to the default
+      shell: "nvidia-smi -rgc"
+      when: gpu_clock_reset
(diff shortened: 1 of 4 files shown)


playbooks/set-gpu-clocks.yml
```diff
@@ -0,0 +1,16 @@
+---
+# set gpu clocks on all worker nodes
+- hosts: compute-node
```
dholt: Can we instead use `slurm-node:kube-node` instead of creating another inventory group? I think the inventory file is getting needlessly long which makes it more difficult to use.

playbooks/set-gpu-clocks.yml
```diff
@@ -0,0 +1,16 @@
...
+  gather_facts: no
+  become: yes
+  vars:
```
dholt: These variables should go in `group_vars` somewhere instead of the playbook. If you want to set defaults, we should create a role.

dholt: Some kind of "gpu management" role might be useful where we could set clocks, reset GPUs, etc.

playbooks/set-gpu-clocks.yml
supertetelman: I've encountered a few cases where people are trying to manage their GPU clusters through deepops _without_  Slurm or K8S. If this is just a use case we won't support or recommend user customization I can make this change.

playbooks/set-gpu-clocks.yml
supertetelman: I agree on the gpu management role and this being rolled in eventually. This was just a quick and dirty playbook people have used for some benchmarking use cases.

We can make this a role, but there's really only 1 variable people will ever change. Max clock or clock reset.

playbooks/set-gpu-clocks.yml
dholt: In that case just use `all`. There’s no reason to create another host group unless you need to differentiate between different sets of nodes (like in a client/server case).

playbooks/set-gpu-clocks.yml
dholt: There are tons of other tunable GPU parameters we could set via a role, I think it would be really useful.

PR #363, opened 2019-10-23
WIP: initial skeleton python package

ryanolson:


src/python/Makefile
```diff
@@ -0,0 +1,88 @@
+.PHONY: clean clean-test clean-pyc clean-build docs help
```
ajdecon: This file adds implicit dependencies on several other tools, including `flake8`, `sphinx-adoc`, `pytest`, `coverage`, and `tox`.

If we want to use this Makefile, we should either add these tools as project dependencies, or provide a setup script just for this cli package to install these.

dholt:
Trying to clean up old PRs, so closing until the `python_cli` branch is more ready as a whole.

PR #382, opened 2019-12-04
clean out the playbook garage ;-)

ajdecon:


ajdecon:
After some discussion offline, many of these that I thought were stale have actually been used recently! This calls more for some reorganization than removal.

Closing PR.

PR #383, opened 2019-12-04
Zabbix playbook

dholt:
TODO:

- [ ] add flag to enable/disable install; disable by default

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -236,3 +236,30 @@ deepops_venv: '{{ deepops_dir }}/venv'
 # OpenMPI
 # Playbook: openmpi
 openmpi_version: 4.0.3
+
+# Zabbix
+# Web User: Admin, Pass: zabbix
+postgresql_users:
+  - name: zabbix-server
+    password: zabbix-server
+zabbix_url: 'http://{{ hostvars[groups["slurm-master"][0]]["ansible_fqdn"] }}:{{ zabbix_apache_vhost_port }}'
+zabbix_server_name: "{{ ansible_fqdn }}"
+zabbix_server_hostname: "{{ ansible_fqdn }}"
+zabbix_version: 4.4
+zabbix_apache_vhost_port: 8080
+zabbix_apache_vhost_tls_port: 8443
+zabbix_timezone: America/Los_Angeles
+zabbix_server_pgsql_login_host: localhost
+zabbix_server_pgsql_login_user: zabbix-server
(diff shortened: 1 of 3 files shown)


config.example/group_vars/all.yml
```diff
@@ -96,3 +96,29 @@ users:
     groups: ['sudo']
     uid: 10007
     home: /home/nvidia
+
+# Zabbix
+# Web User: Admin, Pass: zabbix
+postgresql_users:
+  - name: zabbix-server
+    password: zabbix-server
```
ajdecon: This defaults to creating `zabbix-server` as an unprivileged user, so it can't create the Zabbix DB when it needs to. In order to finish the Ansible run, I manually added privileges to the user via `psql`.

You probably want this to look like:

```
postgresql_users:
    - name: zabbix-server
      password: zabbix-server
      priv: 'ALL'
```

But I haven't tested that. :) Docs here: https://docs.ansible.com/ansible/latest/modules/postgresql_user_module.html

config.example/group_vars/all.yml
```diff
@@ -96,3 +96,29 @@ users:
...
+zabbix_url: 'http://{{ hostvars[groups["slurm-master"][0]]["ansible_fqdn"] }}:{{ zabbix_apache_vhost_port }}'
+zabbix_server_name: "{{ ansible_fqdn }}"
+zabbix_server_hostname: "{{ ansible_fqdn }}"
+zabbix_version: 4.4
+zabbix_apache_vhost_port: 8080
+zabbix_apache_vhost_tls_port: 8443
+zabbix_timezone: America/Los_Angeles
+zabbix_server_pgsql_login_host: localhost
+zabbix_server_pgsql_login_user: zabbix-server
+zabbix_server_pgsql_login_password: zabbix-server
```
ajdecon: Why not use MySQL? Since it's already installed on the `slurm-master` node anyway. :)

dholt: Using MySQL was actually a major disaster, it uninstalled MariaDB without prompting and installed MySQL, which broke Slurm. I agree it would be ideal to use a single DB, so I can keep trying.

ajdecon:
Oh that’s good to know! You can tell I didn’t try. :) If it’s feasible it would be nice to have just one db server but not a huge deal.

—
Adam DeConinck - NVIDIA
Email sent on mobile, apologies for any typos!
________________________________

dholt:
Closing for now, may re-open later

Issue #393, opened 2019-12-12
msg: No package matching 'python2-openshift' found available, installed or updated


k8s-junkie:
Getting this error when trying to create a k8s-cluster on centos nodes..

PLAY [kube-master] **************************************************************************************************************************************************************************
META: ran handlers

TASK [install k8s GPU plugin] ***************************************************************************************************************************************************************
task path: /home/centos/cr8-mec-deployment/cr8-k8s-cluster/playbooks/k8s-gpu-plugin.yml:7

TASK [k8s-gpu-plugin : install virtualenv] **************************************************************************************************************************************************
task path: /home/centos/cr8-mec-deployment/cr8-k8s-cluster/roles/k8s-gpu-plugin/tasks/main.yml:2
skipping: [cpu-node0] => (item=[])  => changed=false 
  ansible_loop_var: item
  item: []
  skip_reason: Conditional result was False

TASK [k8s-gpu-plugin : create location for python virtual env] ******************************************************************************************************************************
task path: /home/centos/cr8-mec-deployment/cr8-k8s-cluster/roles/k8s-gpu-plugin/tasks/main.yml:11
skipping: [cpu-node0] => changed=false 
  skip_reason: Conditional result was False

TASK [k8s-gpu-plugin : install openshift python client for k8s_raw module] ******************************************************************************************************************
task path: /home/centos/cr8-mec-deployment/cr8-k8s-cluster/roles/k8s-gpu-plugin/tasks/main.yml:17
skipping: [cpu-node0] => changed=false 
  skip_reason: Conditional result was False

TASK [k8s-gpu-plugin : install openshift python client for k8s_raw module] ******************************************************************************************************************
task path: /home/centos/cr8-mec-deployment/cr8-k8s-cluster/roles/k8s-gpu-plugin/tasks/main.yml:23
Using module file /usr/lib/python2.7/site-packages/ansible/modules/packaging/os/yum.py
Pipelining is enabled.
<192.168.1.141> ESTABLISH SSH CONNECTION FOR USER: centos
[...]

PR #402, opened 2019-12-30
Ensure EPEL repo is enabled for k8s-gpu-plugin, plus some refactoring

ajdecon:
Addresses #393, where the `k8s-gpu-plugin` role is failing on CentOS because the EPEL repo is not yet installed when we try to install `python2-openshift`.

Actual changes:

1. Create a new role `epel` for installing EPEL repo
1. Make the `k8s-gpu-plugin` role dependent on the `epel` role
1. (Refactor) Remove EPEL installation from other roles, instead making them dependent on `epel` role

The refactor is optional, but I think it's helpful to keep the EPEL code in one place.


michael-balint:
This should be addressed by #227

michael-balint:
Accepted #227 which addresses the same issue.

PR #414, opened 2020-01-22
Minor update to inventory file (comment and ssh key)

supertetelman:
A lot of people have been surprised that DeepOps will update hostnames, so I added a note in there.

I've also some people adding a different ssh-key path, so I added a commented out field as well.


config.example/inventory
```diff
@@ -58,5 +59,7 @@ slurm-node
 [all:vars]
 # SSH User
 #ansible_user=ubuntu
+#ansible_ssh_private_key_file='~/.ssh/id_rsa'
+# Configure SSH bastion/jumpbox for the cluster
```
dholt: Sorry to nit, i think this comment line is from the old inventory missed in the merge. Otherwise looks good to me

supertetelman: Ugh, you're right. There was some odd merge conflict. One sec.

PR #420, opened 2020-01-28
DeepOps post-installation GPU verification

yangatgithub:
This script checks GPU plug-in status after Kubernetes installation.


supertetelman:
Looks good to me.

We should also update the documentation to include this, but that should probably be another PR with some other doc cleanup involved.

PR #428, opened 2020-02-07
Include NGC containers in Kubeflow Notebooks

supertetelman:
This PR should go in after https://github.com/NVIDIA/deepops/pull/427

This update the old NGC functionality. Before deploying Kubeflow we pull down a list of NGC containers and then update the config-map.yaml file to use NGC containers instead of GCP containers.


michael-balint:
@supertetelman please fix merge conflict

PR #431, opened 2020-02-07
Added instructions for increasing virtual vm disk size

michael-balint:
This is handy when testing storage-intensive services like kubeflow on virtual. The default disk size per VM of ~30 GB is not sufficient otherwise and the pods start to fail. ~100 GB per VM is recommended.


roles/nvidia-dgx/vars/ubuntu.yml
```diff
@@ -180,7 +180,7 @@ PKGS_DGX1_ALL:
   - "{{ PKGS_NONVM }}"
   - "{{ PKGS_COMMON }}"
   - "{{ PKGS_DRIVER }}"
-    #- "{{ PKGS_DGX1 }}"
+# - "{{ PKGS_DGX1 }}"
```
supertetelman: Why not just remove this if it is commented out? Do we expect people to use this?

virtual/README.md
```diff
@@ -164,6 +166,26 @@ In the Vagrantfile of choice (Vagrantfile-<os_type>), make the following modific
 
 NOTE: The amount of CPUs and memory on the host system will vary. Change the amounts above accordingly to values that make sense.
 
+### Increase Disk Space
+
+1. Add v.machine_virtual_size = 100 to the Vagrantfile (Vagrantfile-<os_type>). This parameter should go under each libvirt section per node. The units are GBs, so in this case 100 GB are allocated per node.
+2. Vagrant ssh to each machine (ex: `vagrant ssh virtual-gpu01`)  and do the following...
```
supertetelman: Vagrant ssh -> `vagrant ssh`

roles/nvidia-dgx/vars/ubuntu.yml
michael-balint: This is already merged into master. Was working off of a previous branch.

PR #436, opened 2020-02-13
Merged old/roles/nvidia-dgx-firmware and roles/nvidia-dgx-diag

michael-balint:
Adds firmware update to the existing nvidia-dgx-diag role.

TODO: Consolidate this further. It is just at a working state.


michael-balint:
@timuster should also comment (since I pulled his fw update tasks from `old`).

playbooks/nvidia-dgx-fw-update.yml
```diff
@@ -0,0 +1,14 @@
+---
+# This playbook updates the firwmware on DGX nodes
+# Refer to the README in roles/nvidia-dgx-firmware for more info
+
+- hosts: all
+  become: yes
+  connection: paramiko_ssh
```
dholt: probably should remove this

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
+# nvidia-dgx-firmware role
+
+This role makes use of the NVIDIA DGX firmware container and can be used to:
+1. Collect health & configuration diagnostics
+2. Update the firmware
+
+## Setup
+
+1) Download the latest DGX firmware container and put it in `./deepops/config/containers/dgx-firmware`. Keep the original file name. Update the role variables to reflect the version being used.
```
dholt: The default value is `{{ inventory_dir }}/containers/dgx-firmware`, so instead of `./deepops/config/containers/dgx-firmware` we should say `config/containers/dgx-firmware` (which isn't great either since they may not be using `config/`).

Maybe we should just default to the current working directory, which would usually be the repo directory? Then we could document that if you downloaded to `/tmp` or somewhere else, just pass i.e. `-e fw_dir=/tmp`

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+2) Change the `nv_mgmt_interface` variable to reflect the systems being collected from.
+
+```yml
+# The OS/mgmt interface on the server
+nv_mgmt_interface: enp1s0f0 # DGX-1
+# nv_mgmt_interface: enp134s0f0 # DGX-2
+# nv_mgmt_interface: enp2s0f1 # DGX-Station
+```
```
dholt: Can we remove this? I can't find where it's used

roles/nvidia-dgx-firmware/vars/main.yml
```diff
@@ -13,10 +17,21 @@ firmware_container_src_dir: "{{ inventory_dir }}/containers/dgx-firmware"
...
-# Remote directory to save misc. temporary logs to
-nv_diag_dir: "/tmp/nv-diag"
+# Remote directory to save misc. temporary files to
+fw_dir: "/tmp/nvfw"
 
 # The OS/mgmt interface on the server
 nv_mgmt_interface: enp1s0f0 # DGX-1
 # nv_mgmt_interface: enp134s0f0 # DGX-2
 # nv_mgmt_interface: enp2s0f1 # DGX-Station
+
```
dholt: Ditto- Can we remove this? I can't find where it's used

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The default behavior is to collect diagnostics. To disable, change the `run_diagnostics` variable...
+
+```yml
+run_diagnostics: false
+```
+
+### Collected Information
+
+All logs will be stored locally in `./deepops/config/logs`. Logs are stored by hostname with timestamps. To change where logs are stored change the `local_log_directory` variable.
```
dholt: Should also change this path to be relative to the usual working directory, i.e. `config/logs`

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Logs will temporarily be stored in `fw_dir` on the remote machines and will be cleaned up at the end of the playbook. The default remote log dir is `/tmp/nvfw`.
+
+The following information is collected on all `kube-node` and `slurm-node` hosts:
```
dholt: Maybe should say "on all hosts" vs "on all `kube-node` and `slurm-node` hosts"?

playbooks/nvidia-dgx-fw-update.yml
michael-balint: The new firmware container requires running with `-it` with the `update_fw` option. When using the default ssh connection plugin (set to `auto`), it complains about not having a tty. This was the only workaround that seemed to work.

michael-balint: I can leave a comment explaining this in the playbook - would that help?

roles/nvidia-dgx-firmware/README.md
michael-balint: Both the `old` version and the `diag` that was committed recently had the firmware container in the config (aka inventory directory as it was called in `old`). Agree that putting it elsewhere (such as the working dir, might make more sense. Then add an exception to .gitignore in case the working dir is the repo directory?

So `fw_dir` is something else - it is the remote firmware directory. Perhaps it is better to name this `remote_fw_dir` for clarity.

roles/nvidia-dgx-firmware/README.md
michael-balint: Used in [roles/nvidia-dgx-firmware/tasks/get-data.yml](https://github.com/NVIDIA/deepops/blob/11663d96ce6b838e116fabaeb8568ffbadbc71b0/roles/nvidia-dgx-firmware/tasks/get-data.yml), but I guess `get-data` is never used? @supertetelman - do you know why this is here?

roles/nvidia-dgx-firmware/README.md
michael-balint: Yes! I modified this behavior so that it would be all. Need to reflect that in the README.

roles/nvidia-dgx-firmware/tasks/update-firmware.yml
```diff
@@ -2,15 +2,15 @@
...
   set_fact:
     force_parameter: '-f'
-  when: "force_update == 'true'"
+  when: force_update
 
 - name: Update firmware if required
   block:
     - name: Update firmware if required
-      command: "docker run --rm --privileged -v /:/hostfs -t -e auto=1 -e COLUMNS=200 -e LINES=300 {{ firmware_update_container }} update_fw {{ force_parameter }} {{ target_fw }}"
+      command: "docker run --rm -ti --privileged -v /:/hostfs -e auto=1 -e COLUMNS=200 -e LINES=300 {{ firmware_update_repo }}:{{ firmware_update_tag }} set_flags AUTO=1 update_fw {{ force_parameter }} {{ target_fw }}"
```
supertetelman: Does this work on the latest firmware? I recall that some of these flags changed at one point.

roles/nvidia-dgx-firmware/vars/main.yml
supertetelman: This is used to generate a CSV that maps (hostnames, macs, IPs)L

https://github.com/NVIDIA/deepops/blob/master/roles/nvidia-dgx-diag/tasks/get-data.yml

roles/nvidia-dgx-firmware/README.md
supertetelman: I'm actually in favor of keeping it in config. That solidifies the idea that it is not ephemeral and that it is a specific configuration file used for something. If we default to somewhere outside the repo, we might hit permissions issues or it could easily get lost. If we default to working dir, same thing or they might accidentally check it in.

I'm in favor of having a seperate variable to change this directory.

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
...
+1. Collect health & configuration diagnostics
+2. Update the firmware
+
+## Setup
+
+1) Download the latest DGX firmware container and put it in `./deepops/config/containers/dgx-firmware`. Keep the original file name. Update the role variables to reflect the version being used. 
+
+```yml
+# The Docker repo name
+firmware_update_repo: nvfw-dgx1
```
supertetelman: I'm not the biggest fan of the splitup between repo and tag, but it is what it is.

roles/nvidia-dgx-firmware/README.md
supertetelman: This was a special task that isn't included in the role by default. It's excluded because it is datacenter configuration specific and requires minor tweaking based on the networking layout.

This was to map hostnames, IPs, and MACs.

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
...
+* InfiniBand configuration
+* Date
+* DCGM health checks
+* NVSM health checks
+* MAC/IP/hostname
+* Misc. basic information
+
+## Update Firmware
+
+This role can also be used to update the firmware on DGX nodes. To achieve this, set `update_firmware: true`. Running diagnostics and updating firmware can both be achieved in the same run.
```
supertetelman: Has this been tested and does order matter?

I know we've seen issues in the past running firmware checks and firmware updates back to back. Especially when the firmware update involves a reboot.

roles/nvidia-dgx-firmware/README.md
```diff
@@ -0,0 +1,61 @@
...
+* Current firmware
+* Out of date firmware
+* InfiniBand configuration
+* Date
+* DCGM health checks
+* NVSM health checks
+* MAC/IP/hostname
+* Misc. basic information
+
+## Update Firmware
```
supertetelman: Should we mention something about required reboots here?

roles/nvidia-dgx-firmware/tasks/update-firmware.yml
michael-balint: Yes, tested `BMC` update and `all`. Also tested running an update and forced update on already updated nodes. Test subjects were all DGX-1s at a variety of different fw versions.

supertetelman: I have a replacement for this that uses the docker_container module instead of the command. I'll open up a seperate PR for that.

roles/nvidia-dgx-firmware/README.md
michael-balint: Updated in README. Keeping `config` location for now and we can always update later + add a variable (for location) if necessary.

roles/nvidia-dgx-firmware/README.md
michael-balint: Yes - and order doesn't matter. But we put the update first, diagnostics second.

roles/nvidia-dgx-firmware/tasks/update-firmware.yml
michael-balint: Sounds good!

PR #441, opened 2020-02-17
add a playbook and roles to install software modules using EasyBuild

issamsaid:
add a complete setup to install user defined software modules using [easybuild](https://easybuild.readthedocs.io/en/latest/) :

- create a playbook software-modules to encapsulate the whole process.
- create a role to install lmod separately
- create a role to install easybuild binaries and tools
- create a role to checkout from a given repository a set of defaul easy-configs to bootstrap the cluster

Need more testing on Redhat and need to move the easy-configs repo to a DeepOps namespace.


playbooks/software-modules.yml
```diff
@@ -0,0 +1,12 @@
+#
+# install software modules using EasyBuild.
+#
+---
+- hosts: all
+  roles:
+    - name: lmod
+
+- hosts: easy-build-node
```
supertetelman: Can we just set this to slurm-master and get rid of the easy-build-node group as a default?

config.example/inventory
```diff
@@ -53,6 +53,9 @@ kube-node
 slurm-master
 slurm-node
 
+[easy-build-node]
```
supertetelman: Can we just set this to slurm-master and get rid of the easy-build-node group as a default?

roles/easy-build/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+eb_tmp_dir: /tmp/easybuild
+eb_bootstrap_url: 'https://raw.githubusercontent.com/easybuilders/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py'
```
supertetelman: Should we peg this to a specific commit hash rather than just the develop branch?

roles/easy-build-packages/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
+#
+# install software modules using EasyBuild:
+#
+---
+- name: "install depending software"
+  become: true
+  apt:
```
supertetelman: Should we have a check in here to throw an error on non-Ubuntu systems?

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,109 @@
...
+    dest: "{{ eb_tmp_dir }}/modules/all/EasyBuild/{{ ebsw_version.stdout }}.lua"
+    regexp: "{{ item.regexp }}"
+    line: "{{ item.line }}"
+  with_items:
+    - regexp: '^local root .*$'
+      line: 'local root = "{{ sm_software_path }}/EasyBuild/{{ ebsw_version.stdout }}"'
+  when: not ebsw_exe.matched
+
+- name: "cp EasyBuild software"
+  command: cp -R "{{ eb_tmp_dir }}/software/EasyBuild" "{{ sm_software_path }}"
```
supertetelman: We could probably use copy here

roles/easy-build-packages/tasks/main.yml
supertetelman: @issamsaid : Do we need to add the RHEL equivalent here?

supertetelman:
Looks good to me pending an answer to https://github.com/NVIDIA/deepops/pull/441/files/54b2636fde779ac07512241984682b55a6a95e68#r380952587

roles/easy-build-packages/tasks/main.yml
ajdecon: We should split this to do the right pre-install tasks for each distro type, similar to the pattern used here: https://github.com/NVIDIA/deepops/tree/master/roles/nvidia-ml/tasks

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,109 @@
+#
+# an ansible role to install easy_build:
+#
+---
+- name: "install needed packages"
```
supertetelman: @ajdecon , These are both one liners, do you think they need to be split up into rhel/ubuntu yml files?

roles/lmod/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
+#
+# ansible role to install lmod:
+#
+---
+- name: "install packages"
```
supertetelman: @ajdecon , These are both one liners, do you think they need to be split up into rhel/ubuntu yml files?

roles/easy-build/tasks/main.yml
ajdecon: Not necessarily. Probably fine to just have them as single tasks with a conditional. The file split just makes it cleaner to add to later.

roles/lmod/tasks/main.yml
ajdecon: Not necessarily. Probably fine to just have them as single tasks with a conditional. The file split just makes it cleaner to add to later.

roles/easy-build/defaults/main.yml
supertetelman: @issamsaid , can you change this to : https://raw.githubusercontent.com/easybuilders/easybuild-framework/develop/easybuild/scripts/bootstrap_eb.py

supertetelman:
@issamsaid , I made all the necessary changes here: https://github.com/issamsaid/deepops/pull/1 Just merge that through and we can close this PR.

playbooks/software-modules.yml
issamsaid: No, sometimes you want to build the modules on a GPU node. The slurm master may or may not have GPUs. that is why I think we should keep it.

roles/easy-build/defaults/main.yml
issamsaid: Yes.

config.example/inventory
issamsaid: No. Sometimes we need to build the module on a GPU node. The slurm master may or may not have GPUs. I think we should give the option to have a separate node for eb. Worst case scenario we default it slurm-master.

roles/easy-build/tasks/main.yml
issamsaid: Good point.

roles/lmod/tasks/main.yml
issamsaid: I am missing the RH counter part of those.

config.example/inventory
supertetelman: We dn't need to give the option.

If this is a reasonable default, let's leave it. If a user needs to run it on a different box they can just add a new group to the inventory file then update the group in the playbook.

This is what we are now doing in the nfs playbook as well.

issamsaid:
@supertetelman  done it is merged.

michael-balint:
lgtm - @dholt, any comments?

config.example/inventory
dholt: Agree with Adam, we do not need a new group for this, it just adds complexity. The 'hosts' parameter in the playbook should be set to 'all', and the user can run the playbook on whatever node they want using the `-l` flag with `ansible-playbook`

roles/easy-build/tasks/main.yml
```diff
@@ -0,0 +1,109 @@
...
+    - regexp: '^local root .*$'
+      line: 'local root = "{{ sm_software_path }}/EasyBuild/{{ ebsw_version.stdout }}"'
+  when: not ebsw_exe.matched
+
+- name: "cp EasyBuild software"
+  command: cp -R "{{ eb_tmp_dir }}/software/EasyBuild" "{{ sm_software_path }}"
+  when: not ebsw_exe.matched
+
+- name: "cp EasyBuild module"
+  command: cp -R "{{ eb_tmp_dir }}/modules/all/EasyBuild" "{{ sm_module_path }}"
```
dholt: Ditto, use built-in module vs command

roles/lmod/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+    group: root
+    mode: 0777
+  when: ansible_os_family == "RedHat"
+  tags:
+    - configuration
+
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=891541
+- name: "bugfix for lomd posix_c"
+  become: yes
+  shell: |
```
dholt: Should use the built-in ansible module for this vs shell

dholt:
There are still some unresolved comments that I'd like to see fixed, especially the one regarding the extra group.

There are also several places where we use the condensed task format that should be expanded to match the rest of the tasks, i.e. (`file: path={{ eb_tmp_dir }} state=absent force=yes`)

config.example/group_vars/all.yml
```diff
@@ -179,6 +179,17 @@ maas_single_node_install: true
...
 
+################################################################################
+# SOFTWARE MODULES (SM)                                                        #
+################################################################################
+sm_prefix: /sw                                 # has to be in a shared location
+sm_module_root: "{{ sm_prefix }}/modules"
+sm_module_path: "{{ sm_module_root }}/all"
+sm_software_path: "{{ sm_prefix }}/software"
+sm_files_path: "{{ sm_prefix }}/easybuild_files"
+# need to put this in github/DeepOps
```
dholt: We should do this before this PR is merged, can you file a PR against: https://github.com/DeepOps/easybuild_files

issamsaid: can you add me to that one, or push something so that I can issue a pull request

issamsaid: done.

issamsaid:
@michael-balint  can you please review this to unblock the merge?

PR #440, opened 2020-02-15
DeepOps air-gap repo variables update

nvhans:
set variables for offline-mirror/offline-cache installation based on DeepOps tag 19.10 & Kubespray 2.11.0
- leverage port configurations from roles/offline-repos-mirrors/default/main.yml for config/offline_repo_vars.yml
- update SLURM to latest release from dholt
- update CentOS to 7.7.1908 release
- add clarification to yum repo mirrors: base, extras, updates
- add nvidia-dgx-7 & updates to yum repo mirrors
- remove docker-engine yum repo pull: out of date
- add variables to misc-downloads
- add variable section, used by offline-mirror/offline install
- lock python setuptools to <45

PR #446, opened 2020-02-24
Release 20.02

dholt:


michael-balint:
@dholt - can we rebase off of master so we merge #440 into this?

PR #456, opened 2020-03-04
Use Docker role from Kubespray

dholt:
The Docker role from Kubespray is much more robust, and we don't have to maintain it.


config.example/group_vars/all.yml
```diff
@@ -112,12 +112,13 @@ gpu_clock_lock: "1507,1507"
...
-  # Docker's default subnet is 172.17.0.0/16
-  # Pick a smaller subnet, which is also less likely to conflict with corporate addresses.
-  bip: 192.168.99.1/24
+# Docker configuration
+# Playbook: docker, nvidia-docker, k8s-cluster
+# For supported Docker versions, see: kubespray/roles/container-engine/docker/vars/*
+docker_version: '18.09'
+docker_dns_servers_strict: no
+docker_storage_options: -s overlay2
+docker_options: "--bip=192.168.99.1/24"
```
ajdecon: When I tested this locally, Docker failed to start correctly on my GPU node. When I tracked down the error it was the following:

```
unable to configure the Docker daemon with file /etc/docker/daemon.json: the following directives are specified both as a flag and in the configuration file: bip: (from flag: 192.168.99.1/24, from file: 192.168.99.1/24)
```

I think we're seeing this because Kubespray is configuring with options, but the nvidia.nvidia_docker role is configuring with the daemon.json file. For the bip option at least, it seems to be one or the other.

ansible.cfg
```diff
@@ -1,5 +1,6 @@
 [defaults]
-roles_path = ./galaxy-roles:./roles
+roles_path = ./galaxy-roles:./roles:kubespray/roles/
```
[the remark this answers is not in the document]
dholt: Will do, I copied from the docs and it bugged me to, but ignored it in case there was some sort of reason they did it that way.

config.example/group_vars/all.yml
dholt: We updated the nvidia-docker role to not set "bip". You'll have to force update your Galaxy roles (this was added in another PR to the setup script). Also make sure you update your local config with the new changes in the example dir. There weren't a lot of ways I could think of to guarantee a clean update, so we may have to deal with a bit of upgrade pain.

dholt: On second thought I'll just comment this line out, hopefully it'll cut down on upgrade pain

PR #457, opened 2020-03-04
Update Kubespray to v2.12.2 (k8s 1.16.8)

dholt:


supertetelman:
I have verified that this PR will not break Kubeflow v7.1+. I tested this against both the istio_dex.yaml and istio.yaml configurations.

PR #462, opened 2020-03-08
GPU Operator 1.0.0

michael-balint:
+ Bumped version up to 1.0.0
+ Added a note on how to use in kubernetes-cluster.md
+ Increased virtual specs for CPU node (needed for GPU Operator)


docs/kubernetes-cluster.md
```diff
@@ -36,6 +36,15 @@ Instructions for deploying a GPU cluster with Kubernetes
    
    # (optional) Modify `config/group_vars/*.yml` to set configuration parameters
    ```
+
+   Note that as part of the kubernetes deployment process, the default behavior is to also deploy the [NVIDIA k8s-device-plugin](https://github.com/NVIDIA/k8s-device-plugin) for GPU support. The [GPU Operator](https://github.com/NVIDIA/gpu-operator) is an alternative deployment method, which will deploy the device plugin and leverage driver containers within kubernetes. To enable the GPU Operator in DeepOps...
+
+   ```sh
+   vi config/group_vars/all.yml
```
ajdecon: This should be changed to `config/group_vars/k8s-cluster.yml` (or you should move the variable to `config/group_vars/all.yml`)

dholt: the var should stay in `config/group_vars/k8s-cluster.yml` since it's k8s specific

michael-balint: Done.

PR #466, opened 2020-03-10
Upgrade to Kubeflow v1.0.1

supertetelman:
This PR bumps the Kubeflow versions to v1.0.1.

It also encapsulates the hacky uninstall mechanism in a hidden flag. The built-in kubeflow delete is still unreliable when attempting to uninstall istio, etc. Uninstalling and re-installing only the kubeflow ns is now stable.

One other big concern here comes from the Kubeflow [documentation](https://www.kubeflow.org/docs/started/k8s/kfctl-istio-dex/):

> Istio configuration for trustworthy JWTs - This configuration uses Istio version 1.3.1 with SDS enabled, which requires you to use Kubernetes 1.13 or later. You may need to add extra configurations to your Kubernetes as mentioned in Istio’s blog.

I recommend that we make the default installation method be the Istio without Dex method. If we do this we can document or script additional steps to support the Istio+Dex installation.

This version bump includes:

- [x] kfctl binary release v1.0.1

- [x] Kubeflow project release v1.0.1

- [x] Kubeflow Manifests release v1.0.1

This PR will remain in a WIP state until:

- [x] Kubeflow has officially released all packages

- [x] We determine what to do about the new Istio configuration


supertetelman:
It looks like Kubeflow never released manifests for v1.0.0 and skipped to v1.0.1. The on-prem files are identical (aside from the version). I've just bumped these to use the up-to-date files and am re-testing.

supertetelman:
This PR is ready. I've tested both the Dex and non-dex installs on a clean K8S deployment. I have also tested delete/re-install.

supertetelman:
This is currently broken due to :https://github.com/kubeflow/manifests/issues/991 (and it is likely that older versions currently in `master` are also broken.

PR #473, opened 2020-03-12
Fix localusers for Slurm role SSH access

ajdecon:
Ensure that the Ansible SSH user always has access to all hosts. This avoids situations where a non-root Ansible user loses access to the compute hosts after a Slurm job runs.

(I run into this often when troubleshooting virtual installs! :P)


ajdecon:
Ah, this is wrong -- see #315 . I need to find a better fix for this.

PR #476, opened 2020-03-17
Remove a hardcoded domain from monitoring script

supertetelman:


scripts/k8s_deploy_monitoring.sh
```diff
@@ -71,10 +72,10 @@ fi
 NGINX_INGRESS_APP_NAME="${ingress_name}" ./scripts/k8s_deploy_ingress.sh
 
 # Determine correct IP for ingress URL
-ingress_ip_string="$(echo ${master_ip} | tr '.' '-').nip.io"
+ingress_ip_string="$(echo ${master_ip} | tr '.' '-').${MONITORING_DOMAIN}"
 if kubectl describe service -l "app=${ingress_name},component=controller" | grep 'LoadBalancer Ingress' >/dev/null 2>&1; then
     lb_ip="$(kubectl describe service -l "app=${ingress_name},component=controller" | grep 'LoadBalancer Ingress' | awk '{print $3}')"
-    ingress_ip_string="$(echo ${lb_ip} | tr '.' '-').nip.io"
+    ingress_ip_string="$(echo ${lb_ip} | tr '.' '-').${MONITORING_DOMAIN}"
```
ajdecon: nip.io is a service that provides a “magic” domain that translates a formatted domain name to an IP. For example, http://10-0-0-1.nip.io resolves to 10.0.0.1.

I agree this isn’t ideal, especially for offline installs that can’t resolve the magic domain. But I don’t think it can be replaced by an arbitrary custom domain with the same format as nip.io. To have a custom domain, we would need to the user instructions to set up DNS for the IP and naming they want.

Unfortunately [k8s ingress requires a valid dns name](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource) so I don’t think we could use an IP directly with an ingress and one shared load balancer.

I suppose if we wanted to use IPs directly we could give the MetalLB load balancer a range of IPs to use, and just have it assign an IP per service. But I think that would be a larger change.

michael-balint:
Closing, sticking with nip.io as the convention.

PR #488, opened 2020-03-30
When building Slurm cluster, restrict /etc/hosts generation to Slurm

ajdecon:
The template for building `/etc/hosts` depends on getting network interface information for all hosts in the cluster, so that it can provide IPs for the different interfaces in the cluster. Unfortunately, this causes issues if we run Ansible across a subset of hosts on the cluster, because facts are not collected for all hosts. 

Specifically, this interacts poorly with our process for building Slurm clusters if you have Kubernetes hosts in the inventory file.

I'm thinking about a longer-term fix, but for the moment we can resolve this by restricting the `/etc/hosts` template to building hosts for the Slurm cluster. We do this by setting `hosts_add_ansible_managed_hosts_groups: ["slurm-cluster"]`.


ryanolson:
Just a question, are the ansible managed hosts in `/etc/hosts` in a `blockinfile` set?

PR #493, opened 2020-04-03
Fix handling of services when downgrading Slurm

dholt:
Test Plan:

Test downgrading the Slurm version. Run the `slurm-cluster.yml` playbook with `-e slurm_version=20.02.4` and then a second time with `-e slurm_version=19.05.7` and make sure everything is working correctly.

### DIFF
--- a/roles/slurm/tasks/controller.yml
+++ b/roles/slurm/tasks/controller.yml
@@ -79,10 +79,10 @@
   tags:
   - config
 
-- name: start slurmdbd
+- name: restart slurmdbd
   systemd:
     name: slurmdbd
-    state: started
+    state: restarted
     enabled: yes
     daemon-reload: yes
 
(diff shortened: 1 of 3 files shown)


roles/slurm/tasks/controller.yml
```diff
@@ -117,9 +117,23 @@
   environment:
     PATH: '{{ slurm_install_prefix }}/bin:{{ ansible_env.PATH }}'
 
-- name: start slurmctld
+- name: clear slurm state when downgrading
+  command: /usr/sbin/slurmctld -i
```
ajdecon, line 121: Looks like this should be:

```
/usr/local/sbin/slurmctld -i
```

dholt: thanks, should be fixed

PR #497, opened 2020-04-13
Minor fixes to Slurm monitoring

ajdecon:
* Run Prometheus node exporter on the compute nodes
* ~Change the default Prometheus listen port on the head node to avoid conflict with Open OnDemand~
* Change the default OOD port to avoid conflict with Prometheus


ajdecon:
Swapped this to change the OOD port instead. 

It turns out Prometheus will only listen on a different port with a command line flag change, and this is a pain with our current Docker-based setup. I'd have to do port forwarding or override the entrypoint or other such contortions, and it's easier in this context to change the OOD port.

PR #519, opened 2020-05-22
Temporary work-around for Kubespray issue 6160

ajdecon:
https://github.com/kubernetes-sigs/kubespray/issues/6160

Kubespray pins the version of `docker-ce`, but does not pin the version
of `docker-ce-cli` by default.

Unfortunately, Docker has published a version of `docker-ce-cli` (19.03.9)
which is no longer compatible with the pinned `docker-ce` version
(18.09.7). When the two are installed together, any docker cli commands
like `docker images` result in an error like:

```
Error response from daemon: client version 1.40 is too new. Maximum
supported API version is 1.39
```

This PR adds configuration for the Kubespray docker role to pin the
`docker-ce-cli` version as well as the `docker-ce` version.

We will need to track the Kubespray issue so that we can remove this
WAR when it's fixed upstream.

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -226,3 +226,25 @@ deepops_dir: /opt/deepops
 # Directory for python virtualenv
 # Roles: K8s GPU operator, GPU plugin, OpenShift/K8s
 deepops_venv: '{{ deepops_dir }}/venv'
+
+
+################################################################################
+# Bugfix - https://github.com/kubernetes-sigs/kubespray/issues/6160            #
+################################################################################
+ubuntu_docker_pkg_info:
+  pkg_mgr: apt
+  pkgs:
+  - name: "docker-ce-cli=5:19.03.8~3-0~ubuntu-{{ ansible_distribution_release|lower }}"
+    force: yes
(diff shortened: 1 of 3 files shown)


ajdecon:
Note that Kubespray has now fixed this upstream: https://github.com/kubernetes-sigs/kubespray/pull/6163

However the fix is in [Kubespray release v2.13.1](https://github.com/kubernetes-sigs/kubespray/releases/tag/v2.13.1), DeepOps is currently pinned to 1.16.7.

I'm open to feedback on whether we should merge this work-around or attempt to upgrade Kubespray. This PR is the smaller change, but we'll have to update to the newer Kubespray eventually anyway.

dholt:
My inclination would be to just make this PR update kubespray to the latest tag. I think you listed our k8s version, not our kubespray version; I don't think we're too far behind the v2.13.1 tag so it shouldn't be a huge update.

ajdecon:
@dholt : Oops, you're absolutely right. :) Based on dates, it looks like our Kubespray release would be closer to v2.12.x.

I'll open a separate branch to try updating to latest tag and submit it as a PR for CI testing. If that works then we can close this PR.

michael-balint:
Fixed with #521

PR #523, opened 2020-05-29
Update DCGM exporter role to work with new version of container

dholt:
The DCGM exporter now exports its own metrics via port 9400 vs spitting out a flat file that is consumed by the node exporter. This PR modifies the service to expose port 9400 in the container, adds the new endpoints to the Prometheus config on the master node, and updates the GPU dashboard.


roles/grafana/files/gpu-dashboard.json
```diff
@@ -1,2175 +1,2463 @@
...
+          "to": "null"
+        }
+      ],
+      "sparkline": {
+        "fillColor": "rgba(31, 118, 189, 0.18)",
+        "full": false,
+        "lineColor": "rgb(31, 120, 193)",
+        "show": false
+      },
+      "tableColumn": "{instance=\"rl-dgx-d21-u17:9100\", job=\"cluster\", module=\"node-exporter\"}",
```
supertetelman, line 1011: Lots of references to rl-dgx-ds1-u17 in here that we should probably remove.

supertetelman:
This relates to https://github.com/NVIDIA/deepops/pull/563 where I used a cleaned-up version of the same dashboard.

roles/grafana/files/gpu-dashboard.json
dholt: updated to share the dashboard between k8s and slurm

PR #524, opened 2020-05-29
Ceph kubeflow

yangatgithub:
Fix Ceph provisioning issues in AWS for StorageClass creation, which is required by Kubeflow.

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -226,3 +226,25 @@ deepops_dir: /opt/deepops
 # Directory for python virtualenv
 # Roles: K8s GPU operator, GPU plugin, OpenShift/K8s
 deepops_venv: '{{ deepops_dir }}/venv'
+
+
+################################################################################
+# Bugfix - https://github.com/kubernetes-sigs/kubespray/issues/6160            #
+################################################################################
+ubuntu_docker_pkg_info:
+  pkg_mgr: apt
+  pkgs:
+  - name: "docker-ce-cli=5:19.03.8~3-0~ubuntu-{{ ansible_distribution_release|lower }}"
+    force: yes
+  - name: "docker-ce=5:18.09.7~3-0~ubuntu-{{ ansible_distribution_release|lower }}"
+    force: yes
+
+docker_yum_conf: "/etc/yum_docker.conf"
+
+redhat_docker_pkg_info:
+  pkg_mgr: yum
+  pkgs:
+  - name: "docker-ce-cli-19.03.8-3.el7"
+    yum_conf: "{{ docker_yum_conf }}"
+  - name: "docker-ce-18.09.7-3.el7"
+    yum_conf: "{{ docker_yum_conf }}"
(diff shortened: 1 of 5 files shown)


supertetelman:
It looks like you accidentally pulled in a bunch of the Docker changes from Adam's other branch here. Can we pull those out to avoid conflicts?

scripts/k8s_deploy_rook.sh
```diff
@@ -6,7 +6,9 @@
 # `helm upgrade --namespace rook-ceph rook-ceph rook-release/rook-ceph --version v0.9.0-174.g3b14e51`
 
 HELM_ROOK_CHART_REPO="${HELM_ROOK_CHART_REPO:-https://charts.rook.io/release}"
-HELM_ROOK_CHART_VERSION="${HELM_ROOK_CHART_VERSION:-v1.1.1}"
+
+# Always use the latest unless any specific reason for a particular release
+#HELM_ROOK_CHART_VERSION="${HELM_ROOK_CHART_VERSION:-v1.1.1}"
```
supertetelman, line 11: This is bad practice, we want to always peg to a specific version.

Can you update this with whatever the current version is?

scripts/k8s_deploy_rook.sh
```diff
@@ -16,7 +18,7 @@ helm repo add rook-release "${HELM_ROOK_CHART_REPO}"
 # We need to dynamically set up Helm args, so let's use an array
 helm_install_args=("--namespace" "rook-ceph"
                    "--name" "rook-ceph"
-		   "--version" "${HELM_ROOK_CHART_VERSION}"
+#		   "--version" "${HELM_ROOK_CHART_VERSION}"
```
supertetelman, line 21: Same here, we should remove this and just use the current version for now.

services/rook-cluster.yml
```diff
@@ -16,7 +16,7 @@ spec:
     enabled: true
   storage:
     useAllNodes: true
-    useAllDevices: false
+    useAllDevices: true
```
supertetelman, line 19: What does this do? Would this be an issue for larger clusters?

playbooks/k8s-cluster.yml
```diff
@@ -49,6 +49,17 @@
   tags:
     - bootstrap
 
+- hosts: all
```
supertetelman, line 52: We need to pull this commit out.

config.example/group_vars/all.yml
```diff
@@ -226,3 +226,25 @@ deepops_dir: /opt/deepops
 # Directory for python virtualenv
 # Roles: K8s GPU operator, GPU plugin, OpenShift/K8s
 deepops_venv: '{{ deepops_dir }}/venv'
+
```
supertetelman, line 229: We need to pull this commit out.

yangatgithub:
Later on after additional testing I found this works without the “deviceFilter” statement I tried earlier, this approach is more flexible and should work in large cluster.  If users want more granular control, they probably can specifically define the “device” they’d like to use.

yangatgithub:
Accidentally add PR519 branch, will take it out and re-commit with other changes

PR #525, opened 2020-05-29
Fix topology fact gathering for new system type

dholt:
On some newer ASUS servers, the PCIe addressing format changed from `1d:00.0` to `0000:1d:00.0`.

### DIFF
--- a/roles/facts/files/topology.fact
+++ b/roles/facts/files/topology.fact
@@ -10,7 +10,7 @@ echo "    \"threads_per_core\": \"$THREADSPERCORE\"",
 echo "    \"logical_cpus\": \"$(expr $SOCKETS \* $CORESPERSOCKET \* $THREADSPERCORE)\""
 echo "  },"
 echo "  \"gpu_topology\": ["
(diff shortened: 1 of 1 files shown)


roles/facts/files/topology.fact
```diff
@@ -10,7 +10,7 @@ echo "    \"threads_per_core\": \"$THREADSPERCORE\"",
 echo "    \"logical_cpus\": \"$(expr $SOCKETS \* $CORESPERSOCKET \* $THREADSPERCORE)\""
 echo "  },"
 echo "  \"gpu_topology\": ["
-gpus=$(for i in `lspci | grep -E "(3D|VGA compatible) controller: NVIDIA" | awk '{print $1}' | cut -d : -f 1`; do \
+gpus=$(for i in `lspci | grep -E "(3D|VGA compatible) controller: NVIDIA" | awk '{print $1}' | sed 's/^0000://g' | cut -d : -f 1`; do \
```
ajdecon, line 13: This might be a silly question, but do we think this will consistently be 4 leading zeros? Should we just strip any number of leading zeros in the first section of the PCIe address?

E.g., `sed 's/^0+://g'`

PR #536, opened 2020-06-10
Install CUDA Toolkit on DGX by default

dholt:
We were avoiding installing the CUDA toolkit on DGX servers in the nvidia-cuda playbook assuming it came from the DGX OS by default, but it does not.

Also fix some linting.


michael-balint:
Re-ran tests, this PR is passing.

PR #538, opened 2020-06-11
Revert recent Rook upgrade

supertetelman:
We needed to set useAllDevices to true to automatically use the NVME drives in AWS md5 instances.

However this is making the nightly tests fail and may have undesirable consequences in certain clusters, so we are re-disabling this.

### DIFF
--- a/scripts/k8s_deploy_rook.sh
+++ b/scripts/k8s_deploy_rook.sh
@@ -6,7 +6,7 @@
 # `helm upgrade --namespace rook-ceph rook-ceph rook-release/rook-ceph --version v0.9.0-174.g3b14e51`
 
 HELM_ROOK_CHART_REPO="${HELM_ROOK_CHART_REPO:-https://charts.rook.io/release}"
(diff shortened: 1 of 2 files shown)


supertetelman:
Let's not merge this one through yet. I kicked off some more comprehensive tests in Jenkins and the test_ceph.sh script is still failing with this change (due to timeout).

supertetelman:
It looks like with the Rook upgrade from v1.1 to v1.3 directory based (`/var/lib/rook`) OSD support was dropped. 

This feature is intended for "testing" and within the DeepOps project we use it for testing and for POC clusters that need Rook for a dynamic storage class but don't need performance.

We rely on this for testing and for quickly deploying clusters, so we need to revert back to the version where this was supported to avoid complications.

For clusters being deployed outside of this scope I would recommend modifying the default Rook/Ceph configuration (and bumping the version) or looking into NFS integration such as with the NetApp Trident playbooks.

For reference https://github.com/rook/rook/issues/3379.

PR #539, opened 2020-06-14
add nfs-clients and nfs-server hosts 

issamsaid:
This PR amends the following:
- change the hosts of the nfs-client playbook
- change the hosts of the nfs-server playbook
- add nfs hosts to the inventory


dholt:
We've discussed this before and don't want to add more unnecessary host groups. If you need to limit a playbook to a specific host, use the limit flag, `-l`

issamsaid:
@dholt good point. However, can we at least change the hosts in those 2 playbooks to `"{{ hostlist | default('all') }}"`  so that we can integrate  those playbook  into cluster building playbooks with custom hosts?

dholt:
Sure, seems reasonable. See here for how we've integrated them already: https://github.com/NVIDIA/deepops/blob/master/playbooks/open-ondemand.yml

PR #541, opened 2020-06-15
disable ood_portal_generator on Ubuntu

SpontaneousDuck:
ood_portal_generator does not properly find the folder to place the generated portal config file. I disabled it on Ubuntu only so the templated file can be used instead.

### DIFF
--- a/roles/ood-wrapper/vars/ubuntu.yml
+++ b/roles/ood-wrapper/vars/ubuntu.yml
@@ -3,6 +3,8 @@ install_from_src: true
 ood_apache_service_name: apache2
 ood_htpasswd_file: /etc/apache2/.htpasswd
 
+ood_portal_generator: false
+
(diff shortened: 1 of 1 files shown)


dholt:
This should already be the default:

```
galaxy-roles/ood-ansible/defaults/main.yml:109:ood_portal_generator: false
```

I'm curious how you ended up with it evaluating to true or why there was a problem with the default setting, I can't replicate the failure

SpontaneousDuck:
Ah, so the issue come from me updating ood-ansible. They merged some commits to fix errors with v1.7 so I updated the commit used. The commit changing the default to true was made before the one referenced by this repo (https://github.com/OSC/ood-ansible/blob/5d80a62aa2eb15407366464714c3027591e84246/defaults/main.yml) but was not merged until after. Might need to just track this until the version referenced is updated then? They do say in the PR (https://github.com/OSC/ood-ansible/pull/48) for that that setting it to true will be the default going forward for them so this will need to be set to false for any Ubuntu installs if that repo's reference commit is ever updated.

PR #546, opened 2020-06-17
Remove reference to RHEL support.

michael-balint:


supertetelman:
Looks good, but do you also want to update the RHEL references in the dgx-ready and dgx-pod docs?

michael-balint:
> Looks good, but do you also want to update the RHEL references in the dgx-ready and dgx-pod docs?

Done

PR #547, opened 2020-06-17
Modified slurm-cluster.yml to add the ability to change GPU clocks

FattyRichness:
Modified slurm-cluster.yml to add the ability to change GPU clock frequencies via nvidia-smi

### DIFF
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -61,8 +61,8 @@ node_uri: /node
 rnode_uri: /rnode
 
 ################################################################################
-# Set GPU Clocks Adjustment Permission                                                                #
+# Allow the User Permision to Set GPU Clocks                                    #
 ################################################################################
-set_gpu_clocks: no
+allow_user_set_gpu_clocks: no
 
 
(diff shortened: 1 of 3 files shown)


playbooks/gpu-clocks.yml
```diff
@@ -0,0 +1,15 @@
+---
+- hosts: slurm-node
+  become: true
+  tasks:
+   # - name: install custom facts
+   #   include_role:
+   #     name: facts
```
dholt, line 7: I think you want to leave this un-commented out so that you can run it standalone, the facts role is what finds the GPUs on the system.

config.example/group_vars/slurm-cluster.yml
```diff
@@ -59,3 +59,10 @@ httpd_listen_addr_port:
 httpd_use_rewrites: false
 node_uri: /node
 rnode_uri: /rnode
+
+################################################################################
+# Set GPU Clocks Adjustment Permission                                                                #
+################################################################################
+set_gpu_clocks: no
```
dholt, line 66: I think a more descriptive variable would be helpful, I thought this had something to do with actually setting the clock speed. Maybe `allow_user_set_gpu_clocks`?

playbooks/gpu-clocks.yml
FattyRichness: Ah ok i didn't know that's what the facts role did:)

config.example/group_vars/slurm-cluster.yml
FattyRichness: Yes I agree, we can change it to that.

PR #555, opened 2020-07-01
Enable docker networking by default.

michael-balint:
See https://github.com/kubernetes-sigs/kubespray/issues/2002

If this isn't set, we can't reach the outside net from within containers. This affects downloading data within containers, pip installing within jupyter, etc.

### DIFF
--- a/playbooks/docker.yml
+++ b/playbooks/docker.yml
@@ -2,9 +2,11 @@
 - hosts: "{{ hostlist | default('all') }}"
   become: true
   become_method: sudo
+  vars_files:
+    # include kubespray-defaults here so that we can set the facts using the
+    # kubespray 0040-set_facts.yml tasks
+    - ../kubespray/roles/kubespray-defaults/defaults/main.yaml
   tasks:
-    - name: include default kubespray vars required for set_facts tasks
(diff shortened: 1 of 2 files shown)


playbooks/docker.yml
```diff
@@ -2,9 +2,11 @@
 - hosts: "{{ hostlist | default('all') }}"
   become: true
   become_method: sudo
+  vars_files:
+    # include kubespray-defaults here so that we can set the facts using the
+    # kubespray 0040-set_facts.yml tasks
+    - ../kubespray/roles/kubespray-defaults/defaults/main.yaml
   tasks:
-    - name: include default kubespray vars required for set_facts tasks
-      include_vars: ../kubespray/roles/kubespray-defaults/defaults/main.yaml
```
michael-balint, line 7: @dholt - Please see changes here. @issamsaid and I discovered a weird edge case when running this role by itself where our variables in config/group_vars/all.yml are not getting picked up if they already exist in the kubespray-defaults file above. If we use `include_vars`, then the vars are imported for all subsequent parts of the playbook. If we use the `vars_file` convention, then the kubespray_defaults are only used for this block of tasks. This is relevant because otherwise the `docker_iptables_enabled` var does not get set correctly from config/group_vars/all.yml. Only passing it through from the command line works.

Issue #558, opened 2020-07-07
DeepOps monitoring does not use default storage by default


supertetelman:
It looks like after the recent monitoring update persistent storage is no longer working for prometheus.

This can easily be duplicated by:

1. Deploy cluster
2. Deploy/poll ceph
3. Deploy monitoring
4. Verify GPU Node metrics
5. Delete all pods in monitoring namespace
6. Look at GPU Node metrics, all previous metrics will be lost

supertetelman:
I've got a nice and pretty PR here: https://github.com/NVIDIA/deepops/pull/608

But running with this change, I still wasnt seeing my metrics persist. I think something in the config is still off.

supertetelman:
Now that we are integrating nfs-client-provisioner this is a bit easier. Made the necessary updates in this branch: https://github.com/supertetelman/deepops/tree/nfs-client-provisioner-prom.

Despite using a NFS volume to persit `/prometheus` the metrics are lost upon reboot. I need to dig through the prometheus docs more.

supertetelman:
Implemented.

PR #560, opened 2020-07-08
Update Prometheus to use a 10Gi PVC.

tcarland:
Remove storageClassName value to have PVC use the default storageclass

### DIFF
--- a/services/prometheus-monitor.yml
+++ b/services/prometheus-monitor.yml
@@ -585,9 +585,9 @@ metadata:
   annotations:
     volume.alpha.kubernetes.io/storage-class: default
 spec:
-  storageClassName: rook-ceph-block
+  storageClassName: ""
   accessModes:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 2Gi
+      storage: 10Gi
--- a/config.example/helm/monitoring.yml
+++ b/config.example/helm/monitoring.yml
@@ -9,7 +9,17 @@ prometheus:
       kubernetes.io/ingress.class: "nginx"
       nginx.ingress.kubernetes.io/ssl-redirect: "false"
       nginx.ingress.kubernetes.io/rewrite-target: /
+  server:
+    persistentVolume:
+      enabled: true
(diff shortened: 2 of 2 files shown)


supertetelman:
With this change deployment of the monitoring service will fail if Ceph (or another storageclass) has not been deployed first.

Perhaps it would be best to make this an option of the script with a `--persist` flag or something similar.

supertetelman:
This addressess: https://github.com/NVIDIA/deepops/issues/558

services/prometheus-monitor.yml
```diff
@@ -585,9 +585,9 @@ metadata:
   annotations:
     volume.alpha.kubernetes.io/storage-class: default
 spec:
-  storageClassName: rook-ceph-block
+  storageClassName: ""
```
supertetelman, line 588: I'm not sure this is what we want. I think we should be removing this line:

"PVCs don't necessarily have to request a class. A PVC with its storageClassName set equal to "" is always interpreted to be requesting a PV with no class, so it can only be bound to PVs with no class (no annotation or one set equal to ""). A PVC with no storageClassName is not quite the same and is treated differently by the cluster, depending on whether the DefaultStorageClass admission plugin is turned on."

https://kubernetes.io/docs/concepts/storage/persistent-volumes/

supertetelman:
Before merging this PR through we need to address the impact on people who do not have a DefaultStorageClass (or Ceph).

I'm thinking we should do the following:

1. Use this change as the default, users of DeepOps will expect metrics data to persist
2. Do an initial check of DefaultStorageClass like we do in other scripts, print a warning and links to Docs if they don't
3. Add a flag to deploy without persistance and use a modified version of the yaml without the volume added.

supertetelman:
I'm closing this PR, cherry-picking these commits, and opening a new PR later this week with some tweaks to the deployment script to keep some level of backwards compatability.

supertetelman:
https://github.com/NVIDIA/deepops/pull/608

PR #573, opened 2020-07-21
Fix Kubeflow deletion logic and other related workarounds

supertetelman:
Kubeflow has changed the deletion logic and doing a `kfctl delete -f $CONFIG_FILE` now deletes webcert, istio, custom crds, etc. As such I have removed the workaround code. In doing so the `-D` flag is now deprecated and will do the same thing as `-d`.

In addition to this, there is a minor bug that has been causing errors during uninstal (see https://github.com/kubeflow/kubeflow/issues/4815). I have implemented a workaround to this. We are basically just adding the `ClusterName` field to the downloaded Kubeflow manifest before running apply. Without this fix Kubeflow has been throwing an error when you run a `kfctl delete -f $CONFIG_FILE`

As part of this I have bumped up the `kfctl` binary version to the v1.1-rc-1 release candidate.

I also commented out the broken NGC update script in the process.

**Update**
Included in this PR is now the following:

- Install the MPI Operator when install Kubeflow. Update the docs appropriately.
- Update the polling function to wait for Deployments to complete, not just HTTP to respond.

### DIFF
--- a/scripts/k8s_deploy_kubeflow.sh
+++ b/scripts/k8s_deploy_kubeflow.sh
@@ -142,7 +142,7 @@ function stand_up() {
 
   # Make cleanup scripts first in case deployment fails
   # TODO: This kfctl delete seems to be failing due to a Kubeflow config bug
-  echo "cd ${KF_DIR} && ${KFCTL} delete -V -f ${CONFIG_FILE} --delete_storage; cd && sudo rm -rf ${KF_DIR}" > ${KUBEFLOW_DEL_SCRIPT}
+  echo "cd ${KF_DIR} && ${KFCTL} delete -V -f ${CONFIG_FILE} --force-deletion --delete_storage; cd && sudo rm -rf ${KF_DIR}" > ${KUBEFLOW_DEL_SCRIPT}
   chmod +x ${KUBEFLOW_DEL_SCRIPT}
 
   # Initialize and apply the Kubeflow project using the specified config. We do this in two steps to allow a chance to customize the config
--- a/scripts/k8s_deploy_kubeflow.sh
+++ b/scripts/k8s_deploy_kubeflow.sh
@@ -149,7 +149,8 @@ function stand_up() {
   ${KFCTL} build -V -f ${CONFIG_URI}
 
   # Update Kubeflow with the NGC containers and NVIDIA configurations
-  ${SCRIPT_DIR}/update_kubeflow_config.py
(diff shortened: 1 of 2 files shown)


scripts/k8s_deploy_kubeflow.sh
```diff
@@ -19,8 +19,8 @@ export KFCTL="${KFCTL:-${CONFIG_DIR}/kfctl}"
 export KUBEFLOW_DEL_SCRIPT="${KF_DIR}/deepops-delete-kubeflow.sh"
 
 # Download URLs and versions # XXX: kfctl introcuded a version mismatch, this is naming only
-export KFCTL_FILE=kfctl_v1.0.2-0-ga476281_linux.tar.gz # https://github.com/kubeflow/kfctl/releases/tag/v1.0.2
-export KFCTL_URL="https://github.com/kubeflow/kfctl/releases/download/v1.0.2/${KFCTL_FILE}"
+export KFCTL_FILE=kfctl_v1.1-rc.1-0-g3e61b81_linux.tar.gz # https://github.com/kubeflow/kfctl/releases/tag/v1.1-rc.1
```
michael-balint, line 22: v1.1.0 is out - can we upgrade to that instead of the RC? [kf releases](https://github.com/kubeflow/kfctl/releases)

scripts/k8s_deploy_kubeflow.sh
```diff
@@ -36,7 +36,7 @@ function help_me() {
   echo "-h    This message."
   echo "-p    Print out the connection info for Kubeflow"
   echo "-d    Delete Kubeflow from your system (skipping the CRDs and istio-system namespace that may have been installed with Kubeflow"
-  echo "-D    Full Delete Kubeflow from your system along with all Kubeflow CRDs the istio-system namespace. WARNING, do not use this option if other components depend on istio."
+  echo "-D    Deprecatedm same as -d. Previouesly 'Full Delete Kubeflow from your system along with all Kubeflow CRDs the istio-system namespace. WARNING, do not use this option if other components depend on istio.'"
```
michael-balint, line 39: fix spelling

supertetelman: fixed

scripts/k8s_deploy_kubeflow.sh
supertetelman: Done.

PR #576, opened 2020-07-29
Playbook to install Spack package manager

ajdecon:
Adds a new role playbook for installing the Spack package manager.

Actually installing Spack is pretty simple, as it just consists of doing a `git clone`. ;-) But this will also get the dependencies and environment set up correctly, as well as including an optional build step for installing some packages by default. Plus, it's just nice to have all our automation in one place.

## Test plan

1. Turn up a Slurm cluster using `playbooks/slurm-cluster.yml`
1. Run `ansible-playbook playbooks/spack-modules.yml` 
1. Log into the login node and run `spack`

If you want to include the build step, set `spack_build_packages: true` in your config or as an extra-var.

## TODO (WIP)

- [x] Validate that the scheduler detection is correctly picking up the installed Slurm
- [x] Fully validate that the default package list installs correctly
- [x] Enable Spack to work with existing Lmod playbook
- [x] Test to ensure EasyBuild still works correctly after these changes
- [x] Write a docs file
- [x] Should the Spack install directory go in `/sw` by default (as it is now), or should we put it in a location like `/home/.spack` that will be NFS-shared by default?
- [x] Fix NFS configuration so that Spack works correctly on shared filesystem
- [ ] The current release version of Spack doesn't include CUDA 11 yet. Is it better to use the `develop` branch for now to get the later CUDA, or just accept the older CUDA?


ajdecon:
Confirmed that when we build OpenMPI with:

```
spack install openmpi@3.1.6 +cuda +pmi schedulers=auto
```

OpenMPI is picking up the Slurm schedule correctly:

```
vagrant@virtual-login01:~$ spack load openmpi
vagrant@virtual-login01:~$ ompi_info | grep slurm
                 MCA ess: slurm (MCA v2.1.0, API v3.0.0, Component v3.1.6)
                 MCA plm: slurm (MCA v2.1.0, API v2.0.0, Component v3.1.6)
                 MCA ras: slurm (MCA v2.1.0, API v2.0.0, Component v3.1.6)
              MCA schizo: slurm (MCA v2.1.0, API v1.0.0, Component v3.1.6)
```

docs/software-modules.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## Installing Spack
+
+[Spack](https://spack.io/) is a package manager for supercomputers, Linux, and MacOS which makes installing scientific software easy.
+
+To install Spack using DeepOps, you can run:
+
+```bash
+ansible-playbook -l slurm-cluster playbooks/spack-modules.yml
```
dholt, line 23: Great documentation! I think we either need to call out running nfs-server/nfs-client playbooks since they're required and we don't currently run them automatically, or run those playbooks by default somewhere in the slurm-cluster process.

ajdecon: Agree! I think I'll probably put the NFS playbooks by default in slurm-cluster, but that's going to be a separate PR because I want to also add some NFS docs at the same time.

supertetelman: When you do that you might want to clean up the nfs variables. They're currently in all.yml and in slurm-cluster.yml. I ran into some situations where this caused a bit of confusion because people were updating the all.yml and it was not clear that those values are only used for stand-alone or kubernetes (not slurm-cluster).

dholt: That's a pretty fundamental Ansible concept though. I think it makes sense that we have NFS config which is not Slurm related, but also default NFS config that is required for Slurm. Maybe just make sure we have clear comments in the respective files?

PR #581, opened 2020-07-30
Set up NFS shares on Slurm cluster by default

ajdecon:
* Enable NFS shares on Slurm cluster explicitly in the `slurm-cluster.yml` playbook
* Adds variables to the config for turning this on and off
* Adds documentation on using and configuring NFS shares

### DIFF
--- a/config.example/inventory
+++ b/config.example/inventory
@@ -53,6 +53,12 @@ kube-node
 slurm-master
 slurm-node
 
+[nfs-server:children]
+slurm-master
+
+[nfs-clients:children]
+slurm-node
+
 ######
 # SSH connection configuration
 ######
--- a/config.example/inventory
+++ b/config.example/inventory
@@ -53,12 +53,6 @@ kube-node
 slurm-master
 slurm-node
 
-[nfs-server:children]
-slurm-master
-
-[nfs-clients:children]
-slurm-node
-
 ######
 # SSH connection configuration
 ######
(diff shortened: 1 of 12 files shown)


playbooks/slurm-cluster.yml
```diff
@@ -23,6 +23,14 @@
 # Install software
 - include: software.yml
 
+# Set up NFS filesystem
+- include: nfs-server.yml hostlist=slurm-master
```
dholt, line 27: might want to use the hostlist 'hack' here in case someone wants to set up the NFS server on a different system

ajdecon: On reflection, I actually think the right way to do this is in the inventory file. I just pushed an update to refactor for new inventory groups.

config.example/inventory
```diff
@@ -53,6 +53,12 @@ kube-node
 slurm-master
 slurm-node
 
+[nfs-server:children]
```
dholt, line 56: I've really been trying to avoid this, it's just another group people have to configure and add nodes to, and now you can't arbitrarily run against nodes not defined here.

ajdecon: Hmm, fair enough -- I didn't recall the earlier change. Let me think about a better way to handle this.

ajdecon: Ok, let me know what you think of the new solution

Issue #570, opened 2020-07-18
Issue installing Rook/Ceph


almansoorim:
Hi,

I am trying to test deepops in VMs (vmware workstation player), and tried to do  a simple setup:
1 VM for provisioning
1 VM is the master/worker node

I am not trying to run any heavy computations, just wanted to make sure deepops installs correctly.
I cant seem to get Rook/Ceph to work.

Used this command on deploy: ./scripts/k8s_deploy_rook.sh
and this command to monitor: ./scripts/ceph_poll.sh

I keep getting this error:
[errno 5] RADOS I/O error (error connecting to the cluster)
command terminated with exit code 1
command terminated with exit code 1

and on master/worker machine:
IPVS: rr: TCP some_IP:3300 - no destination available

Also, tried new installs couple of time making sure bootstrap-rook.yml also is satisfied. 

Is it possible to test the whole deepops system on VMs? or I am missing something else?

SpontaneousDuck:
I am also seeing this same exact issue!

guilhermehartmann:
Same behaviour in my current installation. Removing the namespace fails on cephclusters.ceph.rook.io that holds the state of the namespace alive.

shadww:
Still looking into this issue. I feel like I am missing something tbh.
If I run "ceph status" command inside "rook-ceph-tools-*", I get the same error: "[errno 5] RADOS I/O error (error connecting to the cluster)"

commands:
> kubectl -n rook-ceph exec -ti rook-ceph-tools-856c5bc6b4-vjq72 bash
> ceph status

almansoorim:
Still looking into this issue. I feel like I am missing something tbh.
If I run "ceph status" command inside "rook-ceph-tools-*", I get the same error: "[errno 5] RADOS I/O error (error connecting to the cluster)"

commands:
> kubectl -n rook-ceph exec -ti rook-ceph-tools-856c5bc6b4-vjq72 bash
> ceph status

SpontaneousDuck:
I believe the issue @guilhermehartmann is seeing is related to [https://github.com/rook/rook/issues/5783](rook/rook#5783) which is being solved in the next major release of rook (v1.4). Basically the cephfs is never created and thus errors out on trying to delete. 

Have you tried an older version of rook yet @almansoorim ? I have not gotten around to trying that but was able to set this up successfully with an older version of deepops. I don't know which version increment is causing the problem for me.

almansoorim:
[...]

PR #582, opened 2020-07-30
peg rook/ceph container to a version not master

supertetelman:
Fix as mentioned in https://github.com/NVIDIA/deepops/issues/570.

In the rook yml we were using `rook/ceph:master` for a container. This caused some issues as rook/ceph moved forward and the rest of our config was tied to an only version.


supertetelman:
This change unblocks the failures we have been seeing in the nightly tests.

PR #585, opened 2020-07-31
Add Jenkins tests for Spack install

ajdecon:
* Minimal test in the per-PR tests (just test that we can install Spack itself)
* Larger test in the nightly test (include the package build)


jenkins/Jenkinsfile-nightly
```diff
@@ -159,6 +159,11 @@ pipeline {
           sh '''
             timeout 60  bash -x ./.jenkins-scripts/test-slurm-job.sh
           '''
+
+          echo "Test Spack (including build)"
```
supertetelman, line 163: We might want to add this to the multi-node test, but it shouldn't matter much for this test so I think it is okay.

PR #588, opened 2020-07-31
slurm playbook should build slurm all at once on all nodes

ajdecon:
## Summary

Today, the `slurm` playbook runs in two stages: first on the controller node, and then on the compute nodes. This makes perfect sense, but because we build Slurm and its dependencies from source, the full execution of the playbook takes a *very* long time. This mostly impacts development, as it makes the CI tests take longer and impacts full-cluster-rebuild speed when iterating on some changes.

This playbook refactors the `slurm` role and playbook to instead run in *three* parts:

- Run on all nodes to build slurm and its dependencies
- Run on the controller node to configure slurm
- Run on the compute nodes to configure slurm

This is absolutely a performance hack for development, but I think it has value given that DeepOps is frequently used to stand up short-lived development clusters.

I also indulged in a little refactoring of the role flow while I was in there, but I think it breaks it up into more manageable chunks and makes the flow a little clearer.

## Test plan

Run the slurm-cluster playbook on a fresh cluster. Slurm should be turned up successfully, but be built on all nodes at the same time.


ajdecon:
As a data point for time saved, using the `virtual/` cluster on my dev setup:

* `./cluster_up.sh` on master took `22m51s` (with CUDA and OpenMPI install disabled)
* `./cluster_up.sh` on this PR took `16m34s` (with CUDA and OpenMPI install disabled)

dholt:
I'm getting a ton of these, I think when it includes `playbooks/nvidia-cuda.yml`:

`[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{
slurm_cluster_install_cuda | default(true) }}`

ajdecon:
@dholt : oops! I frequently forget that particular issue in Ansible conditionals. Have updated with what should be the right syntax.

dholt:
When I pull the latest change and re-run, I'm failing with this:

```
TASK [slurm : copy service files] **************************************************************************************************
Tuesday 04 August 2020  08:47:53 -0700 (0:00:00.056)       0:01:03.225 ********
failed: [chief-liger] (item=slurmd.service) => changed=false
  ansible_loop_var: item
  item: slurmd.service
  msg: Source /tmp/slurm-build/etc/slurmd.service not found
```

Not sure why yet

ajdecon:
@dholt : Got it, that behavior makes sense.

To account for machines where `/tmp` may be wiped, I can think of two possible solutions:

* Move the service file copy to the build step. However, in this PR we don't know which nodes are controllers vs compute during the build, so we would have to copy all service files on all nodes. I think this is probably fine because we don't start services until the later plays with `is_controller` or `is_node` enabled, but there might be scenarios where we accidentally start the wrong services.

* Alternatively, move the build location to a directory which isn't auto-cleaned, such as `/opt/deepops/build-slurm` or something. We could then add an optional step at the end of the role to explicitly clean this up.

I sort of favor the second option, as I can see scenarios where you might want to keep the build directory... But I'm interested in what you think on this before I make that change.

dholt:
I like option 2 also, since part of the build logic is to use the old build directory to clean/uninstall, and that can't happen if it's removed. In that case I agree about an optional cleanup step, which could also have two levels: doing a `make clean` to free up space, and just removing the build directory entirely.

dholt:
Should we also modify the hwloc and pmix build directories? 

```
hwloc_build_dir: /tmp/hwloc-build
pmix_build_dir: /tmp/pmix-build
```

If so, it might make sense to do:

```
/opt/deepops
  /opt/deepops/build
    /opt/deepops/slurm
    /opt/deepops/hwloc
    /opt/deepops/pmix
```

PR #592, opened 2020-08-03
Fixes for Slurm cluster on CentOS 7

ajdecon:
Fix syntax error in redhat tasks for `nvidia-cuda` playbook

### DIFF
--- a/roles/nvidia-cuda/tasks/install-redhat.yml
+++ b/roles/nvidia-cuda/tasks/install-redhat.yml
@@ -1,4 +1,4 @@
---
+---
 - name: RedHat | add epel repo
   yum_repository:
     name: epel


config.example/group_vars/slurm-cluster.yml
```diff
@@ -99,8 +99,8 @@ allow_user_set_gpu_clocks: no
 ################################################################################
 # Enroot & Pyxis                                                               #
 ################################################################################
-slurm_install_enroot: true
-slurm_install_pyxis: true
+slurm_install_enroot: false
```
dholt, line 102: I think we want these by default... we should check and not run the playbook/role elsewhere or add it to the role

ajdecon: Yeah agreed. I opened a PR on the enroot role to add CentOS support: https://github.com/NVIDIA/ansible-role-enroot/pull/6

ajdecon:
@dholt : OK, I had to repair some git antics, but now this is a one-character change 😅

Issue #593, opened 2020-08-04
[request] setting not to set up chrony


ozota1:
deepops 20.06 installs and setups chrony .
But prefferd NTP solution depends on user's environment.
And if there is existing configuration of chrony, the configuration will be overrided and it might be needed to be repaired. 

So I would like a setting not to set up chrony.

for example, by the below setting in group_vars/all.yml, deepops will do nothing about chrony.
```
chrony_setup: "no"
# if chrony_setup is "no", the below settings can be omitted
chrony_service_state: "started"
chrony_service_enabled: "yes"
chrony_timezone: "Etc/UTC"
chrony_config_server:
  - 0.pool.ntp.org
  - 1.pool.ntp.org
  - 2.pool.ntp.org
  - 3.pool.ntp.org
```

ajdecon:
Sure, this is a good idea. I'll take a look at adding this.

ajdecon:
Resolved by #594

ozota1:
Thank you!

PR #594, opened 2020-08-04
Add a flag to disable running the chrony playbook in cluster turnup

ajdecon:
Addresses feature request in #593 . Chrony is still enabled by default.

## Test plan

Set `install_chrony: false` in configuration before running either the `slurm-cluster.yml` or `k8s-cluster.yml` playbook. Observe that the `chrony-client.yml` playbook is skipped:

```
PLAY [all] *********************************************************************************************************************************************************************************************************

TASK [Configure Chrony client] *************************************************************************************************************************************************************************************skipping: [virtual-login01]
skipping: [virtual-gpu01]

TASK [Set timezone] ************************************************************************************************************************************************************************************************skipping: [virtual-login01]
skipping: [virtual-gpu01]
```

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -30,7 +30,7 @@
 
 # NTP configuration
 # Playbook: chrony-client
-install_chrony: yes
+chrony_install: yes
 chrony_service_state: "started"
 chrony_service_enabled: "yes"
 chrony_timezone: "Etc/UTC"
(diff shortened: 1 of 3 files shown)


config.example/group_vars/all.yml
```diff
@@ -30,6 +30,7 @@
 
 # NTP configuration
 # Playbook: chrony-client
+install_chrony: yes
```
dholt, line 33: can we call this something like `chrony_install` so it matches the other vars?

ajdecon: Yeah good call

PR #597, opened 2020-08-05
Slurm role enhance, cleanup, re-org

dholt:
This was initially a change for improving Enroot performance, but it turned into modifying some files to templates, converting static paths to variables, and cleaning up some things to more closely align with the old slurm-perf role.

TODO:

- [x] add a Slurm epilog script that cleans up ENROOT_RUNTIME_PATH and ENROOT_DATA_PATH after a job completes, to make sure the storage used by all containers is reclaimed.

Test Plan:

Example Slurm job to test Enroot/Pyxis functionality:

`srun --pty -G1 --container-image ubuntu --container-name foo hostname`


roles/slurm/templates/etc/slurm/prolog.d/50-all-enroot-dirs
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+set -ex
+
+command -v enroot >/dev/null || exit 0  # enroot not installed
+
+{% if enroot_runtime_path | default(none) %}
+runtime_path="$(sudo -u "$SLURM_JOB_USER" sh -c 'echo "{{ enroot_runtime_path }}"')"
+mkdir -p "$runtime_path"
+chown "$SLURM_JOB_UID:$(id -g "$SLURM_JOB_UID")" "$runtime_path"
+chmod 0700 "$runtime_path"
```
lukeyeager, line 10: Beware that `/run/enroot/` might not be readable+executable by non-root users at this point. For our clusters, we have an additional chmod here (which is super hard-coded and lame):
```
chmod 0755 /run/enroot
chmod 0700 "$runtime_path"
```
I think it depends on what the umask is for root?

dholt: Good call. It did work when I tested but I'll be more explicit. I'm assuming it makes more sense to do something like `$(dirname $runtime_path)` vs hard-coding `/run/enroot`?

lukeyeager: Oh now I remember. The reason we need that additional chmod is that sometimes ill-behaved cluster admins (me) will SSH into a compute node and run enroot as root. In that scenario, the directories aren't created by the slurm prolog - they're created by pyxis. And sometimes they get created in such a way that the parent directory isn't readable by the world. So depending on whether or not you want to deal with cowboy admins, you might be comfortable dropping this additional chmod command.

dholt: better belt and suspenders it, never know who's going to get root privs

PR #605, opened 2020-08-06
Helm 3 upgrade

michael-balint:


ajdecon:
Currently failing for me on `metallb`:

```
+ helm status metallb
+ helm install metallb stable/metallb --values /home/adeconinck/src/github/deepops/virtual/config/helm/metallb.yml
Error: failed to download "stable/metallb" (hint: running `helm repo update` may help)
```

I think we need to explicitly add the stable repo in scripts where we use it.

PR #608, opened 2020-08-07
Persist Prometheus data to PV with option to use legacy code-path

supertetelman:
Based off this PR: https://github.com/NVIDIA/deepops/pull/560

**New** features:

- The default behavior has changed, it now relies on a default storage class
- Default behavior now persists prom data to a PV
- New `-x` flag uses old `yml` which does not persist data to a PV
- Fixed some logic in cmd line parsing
- Added a new Jenkins test that verify delete logic, and PV deployment (does not verify data persists)

**Test plan:**
1. Deploy Kubernetes without Ceph
2. Attempt  to run script. [it should fail]
3. Attempt to run script with -x flag [Verify GPU Node dashboard comes up]
4. Wait 5 minutes and reboot all pods in monitoring namespace. [Verify data has left GPU Node dashboard]
5. Attempt to run script with -d flag [Verify it is deleted
6. Install rook/ceph
7. Attempt to run script [Verify GPU Node dashboard comes up]
8. Wait 5 minutes and reboot all pods in monitoring namespace. [Verify data has persisted in GPU Node dashboard]
9. Attempt to run script with "delete" option [Verify it is deleted]

### DIFF
--- a/.jenkins-scripts/test-monitoring.sh
+++ b/.jenkins-scripts/test-monitoring.sh
@@ -21,6 +21,25 @@ while [ ${time} -lt ${timeout} ]; do
   sleep 15
 done
 
+# Delete Monitoring (this should take ~30 seconds)
+source ./scripts/k8s_deploy_monitoring.sh -d
+
+# Deploy Monitoring without persistent data (this should be faster because containers have already been downloaded)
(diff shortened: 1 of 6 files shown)


supertetelman:
Will re-open later.

PR #615, opened 2020-08-11
Add flags to make NV driver and Singularity optional on Slurm clusters

ajdecon:
In general we want DeepOps to set up a full-featured cluster, but sometimes we don't want that!

* The user may want to configure some components outside of DeepOps
* The user might not need some functionality, like containers
* Or the user might be a DeepOps developer trying to iterate quickly on some piece of code! 😉 

This PR adds new flags to make the NVIDIA driver install and Singularity install optional in Slurm clusters. With this change, all major software installs should now have optional flags except for Slurm itself, plus the minor administrative tasks for Ansible or Slurm to work.

## Test plan

Turned up a virtual cluster with `cluster_up.sh`, with Kubernetes disabled, Slurm enabled, and a vars file including the following:

```
slurm_cluster_install_cuda: no
slurm_cluster_install_openmpi: no
slurm_cluster_install_nvidia_driver: no
slurm_cluster_install_singularity: no
slurm_install_enroot: no
slurm_install_pyxis: no
slurm_enable_monitoring: no
slurm_enable_nfs_server: no
slurm_enable_nfs_client_nodes: no
chrony_install: no
allow_user_set_gpu_clocks: no
```

This gets me a minimally-functional Slurm cluster!

```
$ vagrant ssh virtual-login01
Last login: Tue Aug 11 19:36:44 2020 from 10.0.0.1
vagrant@virtual-login01:~$ srun -N1 hostname
virtual-gpu01
```

And the total turnup time is 13m10s, vs usually taking >25m.

### DIFF
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -36,6 +36,8 @@ hosts_add_ansible_managed_hosts_groups: ["slurm-cluster"]
 ################################################################################
 slurm_cluster_install_cuda: yes
 slurm_cluster_install_openmpi: yes
+slurm_cluster_install_nvidia_driver: yes
+slurm_cluster_install_singularity: yes
 
 ################################################################################
 # NFS                                                                          #
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -37,7 +37,7 @@ hosts_add_ansible_managed_hosts_groups: ["slurm-cluster"]
 slurm_cluster_install_cuda: yes
 slurm_cluster_install_openmpi: yes
 slurm_cluster_install_nvidia_driver: yes
-slurm_cluster_install_singularity: yes
+slurm_cluster_install_singularity: no
 
 ################################################################################
 # NFS                                                                          #
--- a/playbooks/slurm-cluster.yml
+++ b/playbooks/slurm-cluster.yml
@@ -17,6 +17,7 @@
 
 # Install NVIDIA driver
 - include: nvidia-driver.yml
+  when: slurm_cluster_install_nvidia_driver|default(true)
 
 # Install NVIDIA CUDA Toolkit
 - include: nvidia-cuda.yml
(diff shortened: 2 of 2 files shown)


config.example/group_vars/slurm-cluster.yml
```diff
@@ -36,6 +36,8 @@ hosts_add_ansible_managed_hosts_groups: ["slurm-cluster"]
 ################################################################################
 slurm_cluster_install_cuda: yes
 slurm_cluster_install_openmpi: yes
+slurm_cluster_install_nvidia_driver: yes
+slurm_cluster_install_singularity: yes
```
ajdecon, line 40: @dholt : I go back and forth on whether Singularity should be enabled by default, given that we also enable enroot/pyxis by default... Any thoughts? Leaving it as is for now to minimize scope of this PR.

dholt: Great question... based on some other PRs, it should probably be installed via Spack vs manually built, and I'm ok with setting it to not being built by default

ajdecon: Ok, flipping this to "no".

supertetelman:
Did you do anything with the nvidia_driver tag I added a few weeks ago? We should probably consolidate that with https://www.linkedin.com/in/michael-mombourquette-itil-v4-0953463/.

ajdecon:
@supertetelman : Did you intend the LinkedIn link or something else? :)

I didn't touch the tag you added, and that's a fair point. I go back and forth about tags vs vars for turning off functionality, and we're not really coherent in how use use them. 

FWIW, I tend to prefer vars in most cases because they can be saved as part of configuration. I tend to think of tags as a tool to use to skip tasks when you're sitting there running Ansible to accomplish an ad-hoc management task. For example, "lets reconfigure Slurm but skip the nvidia_driver tag because I know nothing changed there".

I'm not sure how to make those two things mesh better, though. 🤷‍♂️

dholt:
Agreed about tags vs. vars. Vars are great to set up your config, tags are great to just run parts when you're deploying. I think the direction we've been going with vars to enable/disable features is great, and we can still add tags to those playbooks to skip for when you're actually doing a deployment. It's way to easy to forget to skip a tag for us to use that as a way to enable/disable features.

PR #627, opened 2020-08-13
 use ansible_ssh_private_key_file + '.pub' in slurm.yml, don't assume id_rsa

supertetelman:
The slurm.yml was always using /home/$USER/.ssh/id_rsa.pub.

Now it uses {{ ansible_ssh_private_key_file  }}.pub if ansible_ssh_private_key_file  is set.

### DIFF
--- a/playbooks/slurm.yml
+++ b/playbooks/slurm.yml
@@ -12,7 +12,7 @@
       authorized_key:
         user: root
         state: present
-        key: "{{ lookup('file', lookup('env','HOME') + '/.ssh/id_rsa.pub') }}"
+        key: "{{ lookup('file', ansible_ssh_private_key_file | default(lookup('env','HOME') + '/.ssh/id_rsa') + '.pub') }}"
 
 # Build slurm first on all nodes
 - hosts: slurm-cluster
--- a/config.example/group_vars/k8s-cluster.yml
+++ b/config.example/group_vars/k8s-cluster.yml
@@ -34,3 +34,6 @@ dashboard_image_repo: "kubernetesui/dashboard"
 
 # kubespray v2.13.1 deploys helm v3.1.2
 helm_version: "v3.1.2"
+
+# Ensure hosts file generation only runs across k8s cluster
+hosts_add_ansible_managed_hosts_groups: ["k8s-cluster"]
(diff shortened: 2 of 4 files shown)


dholt:
Do you mind fixing to just merge the changes you want? Looks like you have a couple other PRs included

playbooks/slurm.yml
```diff
@@ -12,7 +12,7 @@
       authorized_key:
         user: root
         state: present
-        key: "{{ lookup('file', lookup('env','HOME') + '/.ssh/id_rsa.pub') }}"
+        key: "{{ lookup('file', ansible_ssh_private_key_file | default(lookup('env','HOME') + '/.ssh/id_rsa') + '.pub') }}"
```
dholt, line 15: The SSH key value in the `authorized_keys` file is the public key, not the private key. So I'd expect the variable to be named `ansible_ssh_public_key_file` and included the `.pub` suffix.

supertetelman: Oh shoot, yeah. I had that in my test and somehow lost it when I cleaned things up.

supertetelman: Oh no, this is right.

It does `ansible_ssh_private_key_file ` or `default(lookup('env','HOME') + '/.ssh/id_rsa')` and then adds '.pub' to the end of either of those files.

Not sure how I got these extra commits in here though...

dholt: But that doesn't make sense, we should re-name the variable to `ansible_ssh_public_key_file` and the value should include `.pub` vs tacking it on to the value.

`key: "{{ lookup('file', ansible_ssh_public_key_file | default(lookup('env','HOME') + '/.ssh/id_rsa.pub')) }}"`

dholt: It does look like we use `ansible_ssh_private_key_file` elsewhere, are you just trying to re-use it? Maybe we could then do something like this by default in the config?

`ansible_ssh_public_key_file: "{{ ansible_ssh_private_key_file }}.pub"`

supertetelman: Sure, that makes sense. `ansible_ssh_private_key_file ` is just the standard for the private key and we set that in the inventory file.  What you're recommending is the same thing, just a little bit cleaner.

PR #630, opened 2020-08-14
Fix Slurm lib dir permissions

dholt:


ajdecon:
@dholt : what kind of issue did you see that required this permission being set? Wondering if we need to add a new test to Jenkins

dholt:
> @dholt : what kind of issue did you see that required this permission being set? Wondering if we need to add a new test to Jenkins

Updating the hackathon cluster, which was deployed on an older DeepOps version using slurm-perf. Not sure if it's a umask issue or what, but I ran into several areas with bad permissions.

Issue #631, opened 2020-08-17
Kubernetes Dashboard needs a NodePort configuration or else "anonymous user is forbidden"


supertetelman:
It looks like there were some changes in the way the k8s dashboard is accessed. Running the dashboard script will now give the below error as-is in the browser.

The official method of access is to use the kube-proxy or to configure the dashboard with a NodePort (less secure, for dev environments only).

I'd propose that we resolve this by including a patch command to the kubernetes-dashboard service as part of the dashboard user script.

**Error:**

```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {
    
  },
  "status": "Failure",
  "message": "services \"https:kubernetes-dashboard:\" is forbidden: User \"system:anonymous\" cannot get resource \"services/proxy\" in API group \"\" in the namespace \"kube-system\"",
  "reason": "Forbidden",
  "details": {
    "name": "https:kubernetes-dashboard:",
    "kind": "services"
  },
  "code": 403
}
```

PR #632, opened 2020-08-17
K8s dashboard version updates and access fixes

supertetelman:
Addressess https://github.com/NVIDIA/deepops/issues/631

* Bump version of kubernetes dashboard from v2.0.0-rc5 to v2.0.3 along with metrics-scraper
* Update Jenkins test to ignore ssl warning and validate the response of the dashboard
* Update dashboard deployment script to create a NodePort at `31443`, the existing access method no longer worked.

### DIFF
--- a/scripts/k8s_deploy_dashboard_user.sh
+++ b/scripts/k8s_deploy_dashboard_user.sh
@@ -9,12 +9,13 @@ if [ $? -eq 0 ] ; then
 fi
 
 # Get IP of first master
+dashboard_port=$(kubectl -n kube-system get svc kubernetes-dashboard --no-headers -o custom-columns=PORT:.spec.ports.*.nodePort)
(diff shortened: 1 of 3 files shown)


scripts/k8s_deploy_dashboard_user.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# Make the dashboard a NodePort
+kubectl patch svc -n kube-system kubernetes-dashboard  -p '{"spec": {"type": "NodePort", "ports": [{"nodePort": 31443, "port": 443}] }}'
```
supertetelman, line 4: Debated whether this should go here or live in a yaml file in config. I didn't necessarily want to add another config yaml for something that is not used very often and would have had to add more complex port discovery for url printing. This seemed simpler and good enough.

supertetelman:
The dashboard test is passing in the nightlies. And we should probably consider some better Jenkins integration there.

PR #635, opened 2020-08-18
 Fix Slurm install in CentOS due to Lmod changes; eb_modulepath->sm_module_path in centos

supertetelman:
It looks like we renamed this variable and missed updating the name in the .csh file. This was failing the nightly CentOS tests.


ajdecon:
Same change is in #629 with a few other fixes

supertetelman:
Okay great. Caught this in the nightly and didn't see the other PR. I'll close this.

PR #636, opened 2020-08-18
Add additional services/validation in Slurm Cluster Deployment Guide

supertetelman:
Include pyxis/enroot/singularity/open ondemand/monitoring/validation

### DIFF
--- a/docs/deepops/dgx-pod.md
+++ b/docs/deepops/dgx-pod.md
@@ -373,16 +373,16 @@ Deploy Prometheus and Grafana to monitor Kubernetes and cluster nodes:
 ```
 
 The services can be reached from the following addresses:
-* Grafana: http://mgmt:30200
-* Prometheus: http://mgmt:30500
-* Alertmanager: http://mgmt:30400
+* Grafana: http://\<kube-master\>:30200
+* Prometheus: http://\<kube-master\>:30500
+* Alertmanager: http://\<kube-master\>:30400
 
 #### Logging
 
 Follow the [ELK Guide](elk.md) to setup logging in the cluster.
 
 The service can be reached from the following address:
-* Kibana: http://mgmt:30700
+* Kibana: http://\<kube-master\>:30700
 
 
 #### Kubeflow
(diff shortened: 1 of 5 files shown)


docs/slurm-cluster.md
```diff
@@ -54,11 +54,26 @@ Instructions for deploying a GPU cluster with Slurm
...
 
 Now that Slurm is installed, try a ["Hello World" example using MPI](../examples/slurm-mpi-hello/README.md).
 
 
+## Monitoring Slurm
+
+As part of the Slurm installation, Grafana and Prometheus are both deployed.
+
+The services can be reached from the following addresses:
+* Grafana: http://mgmt:3000
```
ajdecon, line 73: Instead of `http://mgmt:3000`, can you restate it to make it clear that `mgmt` is the Slurm controller node? This looks like it should be obvious, but I have seen this kind of description confuse people before. :( 

Either by replacing the URLs directly with something like `http://<slurm-controller>:3000`, or by adding a sentence following the bullets like "where `mgmt` is the name of the Slurm controller node".

supertetelman: Yeah I agree that it needs clarity. I was just copy/pasting from the k8s doc for now. I think both of them should be updated here a bit. I would probably want to call it out using the groups in the inventory file though.

michael-balint:
@supertetelman - as part of the reorg, `docs/slurm-cluster.md` --> `docs/slurm-cluster/README.md`. Can you roll your changes on top of the new file fix the merge conflict above?

docs/slurm-cluster.md
supertetelman: Okay, I made this change and also updated all the other docs. the IPs now reference http://<slurm-master>:PORT and  http://<kube-master>:PORT univerally. This should make it more clear that the IP or hostname matches what is in the inventory file (as long as they haven't set up some sort of loadbalancer to hit their mgmt services).

Thoughts?

ajdecon (approved):
LGTM

PR #658, opened 2020-09-15
Bump nvidia.nvidia_driver role to v1.2.2

ajdecon:
- Addresses a bug with driver install on CentOS/RHEL where there is an
old kernel with no `kernel-headers` package
- Updated documentation in the README


ajdecon:
Rebased after repo reorg

dholt (approved):

PR #660, opened 2020-09-23
Set up node health check for Slurm cluster

ajdecon:
Installs [Node Health Check](https://github.com/mej/nhc) when setting up a Slurm cluster, and configures a very basic set of health checks by default.

NHC is a widely-used tool for performing basic health checks on HPC clusters, and it would be a good tool to include here. This should make it easier for DeepOps users to detect when compute nodes have problems and prevent runs on broken nodes.

## Test plan

1. Turn up a Slurm cluster with this PR
2. On a compute node, check that NHC is installed.
    ```
    root@virtual-gpu01:~# which nhc
    /usr/sbin/nhc
    ```
3. Check that NHC runs cleanly without failures.
    ```
    root@virtual-gpu01:~# nhc -v
    root@virtual-gpu01:~# echo $?
    0
    ```
4. Edit `/etc/nhc/nhc.conf` to list the wrong number of GPUs.
    ```
    root@virtual-gpu01:~# diff /etc/nhc/nhc.conf.orig /etc/nhc/nhc.conf
    27c27
    <  virtual-gpu01 || check_nv_gpu_count 1
    ---
    >  virtual-gpu01 || check_nv_gpu_count 2
    ```
5. Check that NHC fails when we run it again
    ```
    root@virtual-gpu01:~# nhc -v
    ERROR:  nhc:  Health check failed:  Check check_nv_gpu_count 2 returned 1
    ```
6. Check that the node is now drained in Slurm
    ```
    root@virtual-gpu01:~# sinfo -Ra
    REASON               USER      TIMESTAMP           NODELIST
    NHC: Check check_nv_ vagrant   2020-09-24T18:00:23 virtual-gpu01
    ```


ajdecon:
Removed TODOs from summary as they aren't part of the "minimum viable NHC PR".

ajdecon:
Did a bunch of git surgery for rebasing on master

dholt (approved):

PR #663, opened 2020-09-25
Add playbook to disable cloud init

ajdecon:
This PR could be alternatively titled, "cloud-init destroyed my `/etc/hosts` on a MaaS-provisioned instance and it made me sad".

`cloud-init` is a really handy tool for automatically configuring newly-provisioned machines. It's used in a variety of cloud providers, and also used by local provisioning systems like Canonical's [MaaS](https://maas.io/). Unfortunately, `cloud-init` doesn't just run when a machine is first provisioned. It also runs on every reboot, and in its default configuration it manages `/etc/hosts` and a bunch of other files.

This can interact poorly with DeepOps, especially if you're building a Slurm cluster... because we set up our custom `/etc/hosts` file, then reboot the nodes when we install the GPU driver or make changes to Grub. Then after the reboot, `cloud-init` will reset the hosts file to its desired version. Now our `/etc/hosts` is gone and services like Slurm and NFS may no longer function. 😭 

(`cloud-init` also manages [a bunch of other settings](https://cloudinit.readthedocs.io/en/latest/topics/modules.html) with respect to Grub, SSH, Apt, and other tools that DeepOps may want to make changes to. The hosts file was just the one that frustrated me enough to write a playbook!)

This PR adds a playbook which disables `cloud-init` according to [the instructions in its docs](https://cloudinit.readthedocs.io/en/latest/topics/boot.html). It also enables this playbook by default for Slurm and Kubernetes clusters.

## Test plan

1. Provision a new machine that uses `cloud-init`, for example by using MaaS
1. Make a change to `/etc/hosts`
1. Reboot your machine -- your change will be gone...
1. Run `ansible-playbook playbooks/generic/disable-cloud-init.yml`
1. Make a change to `/etc/hosts` and reboot again
1. Observe that your change is still present! 🎉

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -239,3 +239,6 @@ deepops_venv: '{{ deepops_dir }}/venv'
 # OpenMPI
 # Playbook: openmpi
 openmpi_version: 4.0.3
+
+# Disable cloud-init
+deepops_disable_cloud_init: true
--- a/playbooks/generic/disable-cloud-init.yml
+++ b/playbooks/generic/disable-cloud-init.yml
@@ -1,0 +1,41 @@
+---
+# Many provisioning systems, especially for Ubuntu, will include cloud-init
+# as part of the initial setup process for networking and other configuration.
+#
+# However, DeepOps attempts to manage many of the same files, which can lead to
+# conflicts and some components breaking under certain circumstances. /etc/hosts
+# is a good example of a file that both systems try to manage.
+#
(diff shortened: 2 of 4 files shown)


ajdecon:
@supertetelman : I'd appreciate your review and thoughts on this, as you've done more with DeepOps on AWS than me!

@FattyRichness : I think this PR should address the issues we were seeing with MaaS-provisioned hosts in your testing.

dholt (approved):

PR #669, opened 2020-09-28
Fix PMIx unnecessary rebuild

dholt:
We were using the `pmix_info` binary to determine the PMIx version, but since we recently defaulted to an older PMIx version (2.2.4) which does not have this file, PMIx was being rebuilt every time the Slurm playbook was run. This PR uses a header file to grab the PMIx version to compare against the installed version.

### DIFF
--- a/roles/slurm/tasks/pmix.yml
+++ b/roles/slurm/tasks/pmix.yml
@@ -2,16 +2,17 @@
 - name: default to building pmix
   set_fact:
     pmix_build: yes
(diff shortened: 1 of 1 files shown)


ajdecon (approved):
Looks good to me!

Tested as follows:

1. Turned up a new DeepOps Slurm cluster with `virtual/`
1. After it was up, re-ran the `slurm-cluster.yml` playbook
1. Confirmed that the pmix install was skipped in the second run.

PR #672, opened 2020-09-28
Install DCGM via downloaded packages on non-DGX

ajdecon:
DCGM is extremely useful for health checking and diagnostics, but on non-DGX systems it must be explicitly downloaded and installed. We can't download this in DeepOps, because it requires an NVIDIA developer login, but we can provide a pointer and install the package once it's downloaded.

On a DGX, this role will just attempt to install the package from the repositories.

## Test plan

1. Follow the instructions in `docs/slurm-cluster/dcgm.md` to install DCGM and install it.
1. Check that the `datacenter-gpu-manager` package is installed and the `dcgm` service is running.

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -221,6 +221,22 @@ maas_single_node_install: true
 
 maas_kvm: false
 
+
+################################################################################
+# NVIDIA Datacenter GPU Manager                                                #
+################################################################################
+# DCGM is available through package repositories for DGX, but for other host
+# types you must download from the NVIDIA Developer portal at
+# https://developer.nvidia.com/dcgm
+# To install DCGM, please download the correct package type(s) and set the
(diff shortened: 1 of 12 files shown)


supertetelman:
Should this be added to the `ngc-ready-server.yml` playbook? It seems like we don't have a `standalone.yml` playbook anymore, but that is the next best thing with all the other packages.

Also either way, at one point adding a new Jenkins test for this would be useful. Something outside the scope of the slurm or kubernetes tests.

ajdecon:
@supertetelman : Adding it to the `ngc-ready-server.yml` playbook makes sense, though off by default. I'll add that in.

I'm not sure how I'd write a good Jenkins test for it, as the download url for DCGM is only accessible when you're signed into an NVIDIA developer account. The user has to manually download the file(s) and specify a path in their Ansible config to use this playbook.

dholt (approved):

PR #673, opened 2020-09-29
Add GPU Feature Discovery to k8s-cluster for improved MIG support

supertetelman:
Quick update to install the GPU Feature Discovery project the same way we install the GPU plugin.

- [x] Add GPU Feature Discovery

- [x] Bump versions to support official MIG

- [x] Update plugin/feature discovery to take a values.yml (for easy mig configuration)

- [x] Stage GPU Operatore role to support migStrategy


supertetelman:
In this PR I set the default migStrategy to "mixed", but in the packages it is actually defaulted to "none". I'm checked with the dev, and in "mixed" mode non-A100 GPUs will act as-if this is set to "none" and any non-MIG nodes will also act as if this is "none". So I believe "mixed" is the best default to help getting MIG-enabled clusters.

supertetelman:
Resolved the conflict with the other version bump PR.

michael-balint (commented):
Tested - this works for me on DGX V100. Haven't tested the A100/MIG functionality. Please address Kevin's points and then we can approve/merge.

michael-balint (approved):

PR #675, opened 2020-09-29
Fix incorrect paths in k8s deploy ingress script

dholt:
The script to deploy the ingress controller had a couple bad paths after the re-org PR


supertetelman (approved):

supertetelman:
Works for me, passes Jenkins. I'll address my comment in another PR.

PR #680, opened 2020-10-01
Update scripts to be consistant with use of DEEPOPS_CONFIG_DIR instead of config_dir

supertetelman:
- Change config_dir to DEEPOPS_CONFIG_DIR in all scripts
- Add some debug to helm install script based on recent Jenkins/GitHub issues


ajdecon (approved):
Configuration logic all looks correct and CI passed. LGTM.

PR #681, opened 2020-10-01
Fix another path issue in the deploy rook script

supertetelman:


ajdecon (approved):
In CI run, rook script succeeded and didn't show any errors. Logic all looks correct. LGTM.

PR #693, opened 2020-10-08
Add more tests to Slurm job test in Jenkins

ajdecon:
Seeing some failures in Slurm tests that aren't terribly clear. Adding more tests to that script so we can get more signal and information.

Opening as a draft PR to start so I can just see how this behaves. If I think this is useful long-term I'll mark it for review.


ajdecon:
Sigh, maybe Jenkins won't trigger on a draft PR? Marking for review but please don't accept.

ajdecon:
These tests work and likely add more signal to the Jenkins runs for Slurm. Requesting they be merged now.

supertetelman (approved):
LGTM

PR #697, opened 2020-10-09
Fix OpenMPI config flag on CentOS

ajdecon:
Building OpenMPI on CentOS with an external PMIx requires that we explicitly specify the location of libevent, even when it's in the
system paths.


ajdecon:
@supertetelman : Virtual cluster turnup on CentOS worked with this change for me. I'll let the Jenkins CI test this change on Ubuntu. :-)

supertetelman (approved):
LGTM and passing tests.

PR #698, opened 2020-10-09
Change Jenkins MPI test to use NVHPC

ajdecon:
Change the Jenkins MPI test to use NVHPC instead of our hand-built OpenMPI.

If this works well we can drop the OpenMPI build from our Jenkins tests, since we want to test the NVHPC install anyway.


ajdecon:
Worked in a local test, marking ready to review to trigger Jenkins testing.

dholt (approved):

PR #700, opened 2020-10-13
Set failOnInitError to false by default in device plugin

supertetelman:
This is a fix for the failing multi-node tests.

It appears that the device plugin has a "bug". The device plugin should fail if it cannot initialize by default. For whatever reason this is failing the multi-node tests, but not the single mgmt node test.

I'm setting this to false by default because we have other verificaiton in deepops and because many people have mixed architecture clusters supported by deepops.


supertetelman:
See https://github.com/NVIDIA/k8s-device-plugin/issues/199.

dholt:
I don't think this is a bug, we're supposed to be using node selectors when we install the plugin

supertetelman:
If we are supposed to be using NodeSelectors then why does it work with 1 master node, but fail with 3 master nodes? Do you think Kubespray is doing something differently with taints/tolerations in those two cases? That was the main concern I had.

In any case, this is a change in behavior from the last release. Previously this would "just work". And now it silently fails. That's why I mad this PR. If we don't want to push this through we'll probalby need to add some additional NodeSelector logic in the `k8s-cluster.yml` and/or modify our testing process.

michael-balint (approved):

klueska:
A little late to comment, but yes setting failOnInitError to false by default will revert the plugin semantics to the way they used to be. I haven't yet investigated why things seems to "work" with a single master node (when they shouldn't), but setting failOnInitError to false should revert to the old semantics.

PR #709, opened 2020-10-15
Update nvidia-docker playbook to set nvidia as default runtime in DGX

supertetelman:
Currently we still need to manually set `nvidia` to be the default container runtime in `/etc/docker/daemon.json`. This can be done as per the below commit.

Or an alternative approach to this would to just set the DGX override to true by default and allow the galaxy role to do the trick:
```sh
#conf/group_vars/all.yml
docker_override_dgx: true
```

I'd prefer this smaller change as it is safer and will not change docker/nvidia-docker versions, however it does somewhat defeate the purpose of the galaxy role.


dholt:
We explicitly do not do this because it conflicts with the DGX OS packages

PR #543, opened 2020-06-16
EasyBuild roles improvement 

issamsaid:
- removed all the hardcoded shell commands
- removed some unnecessary tasks
- added support for arm64 (jetson+EGX)

PR #711, opened 2020-10-15
EasyBuild roles improvement v2

supertetelman:
See https://github.com/NVIDIA/deepops/pull/543


config.example/group_vars/all.yml
```diff
@@ -220,6 +220,17 @@ maas_single_node_install: true
 
 maas_kvm: false
 
+################################################################################
+# SOFTWARE MODULES (SM)                                                        #
+################################################################################
```
supertetelman, line 225: Not sure if this stuff will conflict with any of the recent slurm changes.

dholt: These were all moved to the slurm-cluster playbook

dholt (changes requested):

michael-balint:
Closing this since we're supporting spack + lmod instead.

PR #712, opened 2020-10-15
NHC support for Ubuntu 20.04 sshd check

dholt:
On Ubuntu 20.04, the NHC script which checks for the `sshd` process fails to find the `sshd` service. It appears the `ssh` and `sshd` services are synonymous. This either has to do with the way the `sshd` service shows up when you run `ps`, or the fact that it's an alias for the `ssh` service (or vice versa). Either way, this fix works on both Ubuntu 18.04 and 20.04


ajdecon (changes requested):
As discussed in chat, needs to be `-d sshd` on 18.04 and `-d sshd:` on 20.04.

ajdecon (approved):
LGTM

PR #713, opened 2020-10-15
Fix Lmod install on Ubuntu 20.04

dholt:
The `lua-posix` package in Ubuntu 20.04 does not have the same packaging bug as the packge in 18.04. So we only make the symlink on 18.04.

This will conflict with #711


ajdecon (approved):
LGTM

PR #708, opened 2020-10-15
Ubuntu 20.04 updates

michael-balint:
* migrated from python2 to python3

Tested on Ubuntu 20.04:
- [x] nvidia-dgx
- [x] k8s-cluster
- [x] slurm-cluster

Tested on Ubuntu 18.04:
- [x] nvidia-dgx
- [x] k8s-cluster
- [x] slurm-cluster

PR #714, opened 2020-10-16
Added libswitch-perl

yangatgithub:


dholt:
Can you please give a description for what this solves? Is this Ubuntu 20.04 specific? If so, you need to make sure the role still works on Ubuntu 18.04, which likely means not changing the package names for every 'Ubuntu' OS.

If you need an example for setting per-distro version vars, see: https://github.com/NVIDIA/deepops/blob/eb699c04fa318dced37eac3da094c68bc571dccf/roles/nhc/tasks/main.yml#L2

dholt:
Closing this in favor of #708, please try that PR and leave comments if it doesn't work on Ubuntu 20.04

PR #729, opened 2020-10-22
Add build dependencies for NHC

dholt:
Several of our roles which build packages from source don't specify the 'build-essential' package on Ubuntu which is required to build. Since NHC is first in the slurm-cluster playlist, I hit the failure there first :) but we'll eventually need to add this same fix to the other roles, like OpenMPI.


ajdecon (approved):
LGTM

Issue #732, opened 2020-10-24
cuda baseurl causing apt update fails


itzsimpl:
There seems to be an issue with the `nvidia_driver_ubuntu_cuda_repo_baseurl` as `apt update` started failing with the following error:
```
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64  Release: The following signatures were invalid: BADSIG F60F4B3D7FA2AF80 cudatools <cudatools@nvidia.com>
W: Failed to fetch http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/Release.gpg  The following signatures were invalid: BADSIG F60F4B3D7FA2AF80 cudatools <cudatools@nvidia.com>
W: Some index files failed to download. They have been ignored, or old ones used instead.
```
changing the `baseurl` to `https` solves the issue; the offending configs are:
https://github.com/NVIDIA/deepops/blob/e91a0feff99d152af5dd71cc8d72b185e6190ea6/roles/nvidia-cuda/defaults/main.yml#L28
and 
https://github.com/NVIDIA/ansible-role-nvidia-driver/blob/5f58443c4b2b9f3e0ca09458cec04905648bb140/defaults/main.yml#L17

ajdecon:
Hi @itzsimpl : Would you like to [contribute a PR](https://github.com/NVIDIA/deepops/blob/master/CONTRIBUTING.md) for this change? As you appear to have identified the fix, I'd love to let you get the credit for the fix. If not I can look at rolling a PR for this.

PR #743, opened 2020-10-30
Fix for #732; apt update fails due to GPG error

itzsimpl:
Note that the galaxy scripts should be updated as well.


yangatgithub:
Hi, which exactly command did you run when you saw the GPG error? thanks.

yangatgithub (approved):

michael-balint (approved):

PR #745, opened 2020-11-02
Fix for all GPUs visible when ssh on slurm compute node

itzsimpl:
As mentioned in https://slurm.schedmd.com/pam_slurm_adopt.html, "The pam_systemd module will conflict with pam_slurm_adopt, so you need to disable it in all files that are included in sshd or system-auth (e.g. password-auth, common-session, etc.)." 

Without disabling it, when a user started a job, they could ssh to the compute node and while there see all GPUs that are installed on the system even though their job did not specifically request for them.

### DIFF
--- a/roles/slurm/tasks/compute.yml
+++ b/roles/slurm/tasks/compute.yml
@@ -145,3 +145,11 @@
     marker: "# {mark} ANSIBLE MANAGED BLOCK (ansible-role-slurm)"
   tags:
     - config
+
+- name: comment out pam_systemd.so as it conflicts with pam_slurm_adapt.so, see https://slurm.schedmd.com/pam_slurm_adopt.html
+  replace:
+    path: /etc/pam.d/common-session
+    regexp: '^([^#].*pam_systemd.so.*)'
+    replace: '# \1 # ANSIBLE MANAGED (ansible-role-slurm)'
+  tags:
+    - config
--- a/roles/slurm/tasks/compute.yml
+++ b/roles/slurm/tasks/compute.yml
@@ -151,5 +151,6 @@
     path: /etc/pam.d/common-session
     regexp: '^([^#].*pam_systemd.so.*)'
     replace: '# \1 # ANSIBLE MANAGED (ansible-role-slurm)'
+  when: ansible_distribution == 'Ubuntu'
   tags:
     - config


dholt:
Thanks for the PR!

This PAM file doesn't exist on CentOS. I'm not sure which is correct, maybe `system-auth` and/or `password-auth`. 

In the spirit of something being better than nothing, if you want to restrict this task to Ubuntu, I think we can merge it.

dholt:
Testing this on Ubuntu 18.04, I'm still able to SSH in to a node where I have a job running and see a GPU, even though the job requested no GPUs.

itzsimpl:
This puzzles me, as I tested on Ubuntu 18.04.5 on two systems, pre and post the patch. As soon as I remove the patch I'm able to see all GPUs, as soon as I add it I see no GPUs.

dholt:
> This puzzles me, as I tested on Ubuntu 18.04.5 on two systems, pre and post the patch. As soon as I remove the patch I'm able to see all GPUs, as soon as I add it I see no GPUs.

Yeah this was my mistake and i deleted my comment. I was able to log in because my user was in `/etc/localusers`.

itzsimpl:
Ok. I'll restrict it to Ubuntu then.

itzsimpl:
FYI; it seems that on CentOS password-auth should be patched; see https://hpcworks.wordpress.com/2017/05/29/setup-slurm-pam-plugin-on-centos-7/. Unfortunately, I don't have a CentOS setup, so I cannot check.

likueimo:
I do similar thing on CentOS 7. 
Below is my bash script.
Maybe it helps .
```bash
# for slurm security
logger -t xcatpost_sh_preboot "mask systemd-logind for pam_slurm_adopt.so"
systemctl mask systemd-logind
# disable pam_systemd.so
# search /etc/pam.d contain pam_systemd and mark-out
sed -i 's/.*pam_systemd.so/#&/' $(grep -r -l /etc/pam.d/ -e 'pam_systemd.so')
```

michael-balint (approved):

avolkov1:
To get pam_slurm_adopt to work In addition to these changes in my testing I had to change these lines:
https://github.com/NVIDIA/deepops/blob/332e7f22e8c77dcf90c37c652d8648b0d682e0f2/roles/slurm/tasks/compute.yml#L138-L147

to:
```
- name: setup /etc/pam.d/sshd
  blockinfile:
    path: /etc/pam.d/sshd
    block: |
      -account required {{ slurm_pam_lib_dir }}/pam_slurm_adopt.so action_no_jobs=ignore
      -account sufficient pam_listfile.so item=user sense=allow file=/etc/localusers
    insertafter: EOF
    marker: "# {mark} ANSIBLE MANAGED BLOCK (ansible-role-slurm)"
  tags:
    - config
```
Above I had to change the order with pam_slurm_adopt before pam_listfile modules. The option `action_no_jobs=ignore` option to pam_slurm_adopt is so that the  pam_listfile would run for local users.

PR #746, opened 2020-11-02
Increase security, limit access to slurmdbd.conf

itzsimpl:
According to https://www.hpcsec.com/2019/12/22/cve-2019-19727/, slurmdbd.conf should have strict access permissions as it contains clear text credentials for mysql.


ajdecon (approved):
LGTM

PR #747, opened 2020-11-03
Looks like with the move from python2 to python3 we no longer need sshpass

supertetelman:
We had a comment in `k8s-cluster.yml` indicating that sshpass was only needed due to a Python2 bug. With the move to Python3 I am removing this line.

We may want to remove this from `setup.sh` as well.

### DIFF
--- a/playbooks/k8s-cluster.yml
+++ b/playbooks/k8s-cluster.yml
@@ -42,18 +42,6 @@
   tags:
     - bootstrap
 
-# Install 'sshpass' program for: https://github.com/ansible/ansible/issues/56629
-- hosts: all
-  gather_facts: true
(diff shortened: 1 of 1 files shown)


dholt:
What does this do if you still use python2?

supertetelman:
I guess it doesn't matter, looks like it is still failing even on the python3 bits:
```TASK [download : prep_kubeadm_images | Copy kubeadm binary from download dir to system path] ***
fatal: [virtual-mgmt01-1 -> 10.0.0.21]: FAILED! => changed=false 
  cmd: sshpass
  msg: '[Errno 2] No such file or directory: b''sshpass'': b''sshpass'''
  rc: 2```

PR #748, opened 2020-11-03
Update bootstrap Python to support CentOS

supertetelman:
We recently bumped from Python2 to Python3. 

This caused our `bootstrap-python.yml` to start attempting an apt-get install on CentOS systems. This PR fixes that.

Because this is the bootstrap of python, it is possible that we are unable to gather_facts on some systems, so the `ansible_os_distribution` variables cannot be captured/used.

My strategy around this was nesting install commands, which is now passing internal/manual tests.

### DIFF
--- a/playbooks/bootstrap/bootstrap-python.yml
+++ b/playbooks/bootstrap/bootstrap-python.yml
@@ -1,18 +1,16 @@
 ---
 - hosts: all
   become: true
-  gather_facts: False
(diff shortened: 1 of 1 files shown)


dholt:
Verified this works against CentOS 7. Have not tried Ubuntu 18 or 20

dholt (approved):

PR #755, opened 2020-11-05
Implement a caching container registry for Slurm and k8s

ajdecon:
This PR sets up a local Docker registry, with caching enabled for pulls from Docker Hub, and configures the nodes to pull their containers via this registry.

In the case of Kubernetes, registry mirrors will be deployed to each Kubernetes master.

In the case of Slurm, registry mirrors will be deployed to each Slurm controller.

## Test plan

1. Turn up a new cluster
1. On the registry host, check that there is a container `deepops-registry` running
1. Attempt to pull a container such as `ubuntu:latest` from any host in the cluster
1. Check the logs on the registry container and validate that the container was pulled via the registry container as a proxy.


playbooks/k8s-cluster.yml
```diff
@@ -42,6 +42,16 @@
   tags:
     - bootstrap
 
+# Configure hostnames, /etc/hosts
+- include: generic/hosts.yml
+
+# Set up a local cluster container registry
+- include: container/standalone-container-registry.yml
+  vars:
+    standalone_container_registry_host: "{{ groups['kube-master'][0] }}"
```
supertetelman, line 51: Will this be overriden by whatever is in the `all.yml`?

roles/standalone-container-registry/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+      - docker
+    state: present
+  when: ansible_os_family == "RedHat"
+
+- name: run registry docker container
+  docker_container:
+    name: "{{ standalone_container_registry_name }}"
+    image: "{{ standalone_container_registry_image }}"
+    state: started
+    network_mode: host
```
michael-balint, line 48: By default, if docker restarts (as is the case when deploying K8s), this registry container will not be restarted. Suggest adding `restart: yes` as part of the `docker_container` parameters. Might also consider a `restart_policy: always` and `restart_retries` set to a reasonable number (5?). See https://docs.ansible.com/ansible/2.5/modules/docker_container_module.html

michael-balint (changes requested):
If docker restarts, we need to ensure that the registry container will also restart. See suggestion on `roles/standalone-container-registry/tasks/main.yml`.

roles/standalone-container-registry/tasks/main.yml
ajdecon: Added in d4a9d4445e547efdf6d84cbfa93b4822127471ec

michael-balint:
Reminder that the registry mirror tags are not getting picked up on dgx because of the following file being used:

```
$ cat /etc/systemd/system/docker.service.d/docker-override.conf
[Unit]
After=nvidia-fabricmanager.service

[Service]
ExecStart=
ExecStart=/usr/bin/dockerd -H fd:// -s overlay2 --default-shm-size=1G
LimitMEMLOCK=infinity
LimitSTACK=67108864
```

dholt (approved):

michael-balint (approved):
Works by default with #758 changes.

PR #758, opened 2020-11-10
Allow disabling docker/nvidia-docker install

dholt:
Adds a flag to disable installation of docker/nvidia docker. On DGX OS, this will erase the default DGX OS docker configuration and install the version from DeepOps.

Notes:
* This does not fix the issue where the DGX OS default docker runtime is not set to 'nvidia' by default.
* This does not affect a 'k8s-cluster' install, which still does its own thing via kubespray

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -121,12 +121,11 @@ gpu_clock_lock: "1507,1507"
 # Playbook: docker, nvidia-docker, k8s-cluster
 # 
 # For supported Docker versions, see: kubespray/roles/container-engine/docker/vars/*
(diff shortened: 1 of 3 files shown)


supertetelman:
Won't this break things for clusters that have dgx and non-dgx nodes?

michael-balint:
Since we're removing the DGX check as part of the when statement, perhaps we should just change the variable name. It isn't for DGX specifically, the variable is now to override nvidia-docker install regardless of what the host already has.

dholt:
Updated PR based on feedback and discussions

michael-balint (approved):

PR #766, opened 2020-11-20
Update nvidia-dgx-firmware role to work with new update container with more verifications

supertetelman:
**Updates:**
* Bump versions, new container introduces a proper automated version complete with json output
* Use proper docker run with automation flags instead of the hacky `-it` with formatting that was necessary
* Add md5sum checksums to container tarball before any updates
* Update readme
* Add json parsing tool to get status from the output of the new container
* Add some debug
* Collapse some of the playbooks and improve tagging/error ignoring

**TODOS:**
* In general this is complete, but there are still a few cases where we check firmware versions or attempt to update a few more times than necessary. This only wastes a little bit of time and I'd rather have more verification/debug steps than less when dealing with firmware upgrades.

* Some of the variables also have generic names and could use a rename to start with `nv_dgx_fw`

**Testing plan:**

- [x] Verify `playbooks/nvidia-dgx/nvidia-dgx-diag.yml` collects all logs
- [x] Verify `playbooks/nvidia-dgx/nvidia-dgx-fw-update.yml` works when no updates are needed (by running the same update container twice)
- [x] Verify `playbooks/nvidia-dgx/nvidia-dgx-fw-update.yml` works when updates are needed (by using a new container or downgrading, then upgrading)
- [x] Verify `playbooks/nvidia-dgx/nvidia-dgx-fw-update.yml` works when no updates areforced (by setting `force_firmware` with a single component)
- [x] Verify `playbooks/nvidia-dgx/nvidia-dgx-fw-update.yml` works when updating only a single compontent (by setting `target_fw`)

### DIFF
--- a/docs/deepops/dgx-diagnostic-firmware.md
+++ b/docs/deepops/dgx-diagnostic-firmware.md
@@ -1,1 +1,18 @@
-/home/atetelman/repo/sa/deepops/roles/nvidia-dgx-firmware/README.md
+# DGX firmware upgrade, diagnostics, and log collection
+
+The nvidia-dgx-firmware role has been built to perform several administrative tasks.
+
+For DGX-specific clusters it can be used to:
+
+* Gather manifest of currently installed DGX firmware
+* Update DGX firmware
+* Gather IB configuration information
+
+For all GPU clusters it can be used to:
+
+* Run NVSM health checks
+* Run DCGM stress checks
+* Gather log files from all nodes
+
+
+For full usage and details see the role (README)[../../roles/nvidia-dgx-firmware/README.md].
--- a/docs/deepops/dgx-diagnostic-firmware.md
+++ b/docs/deepops/dgx-diagnostic-firmware.md
@@ -1,0 +1,1 @@
+/home/atetelman/repo/sa/deepops/roles/nvidia-dgx-firmware/README.md
--- a/roles/nvidia-dgx-firmware/README.md
+++ b/roles/nvidia-dgx-firmware/README.md
@@ -6,7 +6,7 @@ This role makes use of the NVIDIA DGX firmware container and can be used to:
 
 ## Setup
 
-1) Download the latest DGX firmware container and put it in `src/containers/dgx-firmware`. Keep the original file name. Update the role variables to reflect the version being used. 
+1) Download the latest [DGX firmware container](https://docs.nvidia.com/dgx/dgxa100-fw-container-release-notes/index.html) and put it in `src/containers/dgx-firmware`. Keep the original file name. Update the role variables to reflect the version being used. 
 
 ```yml
 # The Docker repo name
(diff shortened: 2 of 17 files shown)


roles/nvidia-dgx-firmware/README.md
```diff
@@ -22,14 +22,17 @@ firmware_update_tag: 19.10.7
 # The OS/mgmt interface on the server
 nv_mgmt_interface: enp1s0f0 # DGX-1
 # nv_mgmt_interface: enp134s0f0 # DGX-2
+# nv_mgmt_interface: enp225s0f0 # DGX A100
```
michael-balint, line 25: Make the default DGX A100 since in the example above it is for A100.

roles/nvidia-dgx-firmware/README.md
```diff
@@ -71,3 +85,7 @@ ansible-playbook -l slurm-node playbooks/nvidia-dgx/nvidia-dgx-diag.yml
 # update all firmware
 ansible-playbook -l slurm-node playbooks/nvidia-dgx/nvidia-dgx-fw-update.yml
 ```
+
+> Note: This playbook is designed to only allow upgrading of a single component per run; the recommended best-practice is to run `update_fw all`, however when updating individual components it is best to perform some level of manual verificaiton over the logs.
```
michael-balint, line 89: Correct spelling (verification).

docs/deepops/dgx-diagnostic-firmware.md
```diff
@@ -0,0 +1 @@
+/home/atetelman/repo/sa/deepops/roles/nvidia-dgx-firmware/README.md
```
michael-balint, line 1: Please create a link or a more substantive file with a pointer to the role README.

roles/nvidia-dgx-firmware/defaults/main.yml
```diff
@@ -3,15 +3,22 @@
 # Which portions of the role to run
 run_diagnostics: true
 update_firmware: false
+load_firmware: false
```
michael-balint, line 6: If we're adding a var to load the firmware, shouldn't we also make changes to the `nvidia-dgx-fw-update.yml` playbook and any associated docs?

michael-balint (changes requested):

docs/deepops/dgx-diagnostic-firmware.md
supertetelman: You don't like just having a redirect/symlink?

roles/nvidia-dgx-firmware/defaults/main.yml
supertetelman: I'm changing the default behavior here to true and adding a comment.

This will simply copy the firmware container over and load it. We want to do this for both diagnostics and FW update, the only reason we would set this to false is if the container has already been loaded manually or we are re-running the playbook for some reason.

michael-balint (approved):

PR #768, opened 2020-11-23
Add a convenience playbook for docker login

ajdecon:
Providing a Docker login functionality will enable some key workflows, including:

- Using Docker Pro or other accounts that provide greater rate limits than anonymous access
- Accessing private containers that are not available anonymously

For test instructions, see `docs/docker-login.md`. Note that these instructions will walk you through how to put the login credentials in an Ansible Vault file for security.


config.example/group_vars/all.yml
```diff
@@ -168,6 +168,17 @@ bind_paths: []
...
+#   Note: example only! you should put these in an Ansible Vault file for security!
+#
+#docker_login_registries:
+#- registry: docker.io
+#  username: myuser
+#  password: mypassword
+#  email: docker@docker.io
+#- registry: nvcr.io
+#  username: '$oauthtoken'
+#  password: mypassword
```
dholt, line 182: Can we move these up near the other docker options?

dholt (dismissed):
A couple minor organizational comments.

Can we move the documentation to a subfolder in docs? Don't care where, but probably not root. Maybe make a `misc` folder and put this doc and the firewall doc in there?

config.example/group_vars/all.yml
ajdecon: @dholt : Made this change and the recommended move of the doc file.

michael-balint (commented):
Playbook runs successfully, but nothing gets saved to `~/.docker/config`. Doing `docker pull` on a container in one of the registries yields an error asking for authentication.

```
TASK [docker-login : log into docker registry] **************************************************
changed: [deepops-dgx1q-05] => (item={'registry': 'nvcr.io', 'username': '$oauthtoken', 'password': '<redacted>'})
changed: [deepops-dgx1q-05] => (item={'registry': 'docker.io', 'username': '<redacted>', 'password': '<redacted>'})
```

Also, if possible, is there a way to scrub the ansible logs above so that the passwords aren't being shown in plaintext?

ajdecon:
Added `no_log` flag to `docker_login` task.

Tested and confirmed that `/root/.docker/config.json` is being set in my tests.

michael-balint (approved):

PR #769, opened 2020-11-26
Add RHEL/CentOS 8 support

nvhans:
RHEL/CentOS 8 EPEL repository urls do not follow the same structure as RHEL/CentOS 7 EPEL repository. Mostly just modify existing yum_repository with static url & key methods with yum install epel-release method. This provides the proper create epel.repo functions on RHEL 7 and RHEL 8.

Some package names are different between 7 & 8 requiring additional ansible_distribution_major_version checks.


ajdecon (approved):
LGTM

Issue #788, opened 2020-12-11
Create a simple debug.sh script for K8s to generate debug bundles


supertetelman:
I regularly help people debug K8s deployment issues and the issue is almost always identified in one of the below files.

It would be handy to create a script that runs all this and puts it into a Tarabll that can be shared.

* run the verify_gpu.sh
* get inventory file
* get nvidia daemon log files
* kubectl describe nodes
* kubectl get pods -A

supertetelman:
Implemented.

PR #789, opened 2020-12-11
Add a simple debug script for k8s

supertetelman:
Addresses https://github.com/NVIDIA/deepops/issues/788

I am doing this in a shell script rather than a Ansible Playbook to make the log collection process more transparent for people who are less familiar with Ansible.

Basic tool to collect logs and basic debug information that solves 95% of the K8s problems users run into.

### DIFF
--- a/scripts/k8s/debug.sh
+++ b/scripts/k8s/debug.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+timestamp=$(date +%s)
+logdir=config/log_${timestamp}
+mkdir ${logdir}
+
+# Provisioner configuration
+cp config/inventory ${logdir}
+
+# Docker configuration
+ansible kube-node -ba "docker info" -vv > ${logdir}/docker-info.log
+ansible kube-node -ba "cat /etc/docker/daemon.json" -vv > ${logdir}/docker-daemon.log
+
+# Kubectl
+kubectl get pods -A > ${logdir}/get-pods.log
+kubectl get daemonset -A > ${logdir}/get-daemons.log
+kubectl get nodes > ${logdir}/get-nodes.log
+kubectl describe nodes > ${logdir}/describe-nodes.log
+for pod in $(kubectl get pods -n kube-system | grep nvidia-device-plugin | awk '{print $1}'); do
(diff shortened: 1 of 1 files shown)


scripts/k8s/debug.sh
```diff
@@ -0,0 +1,28 @@
...
+kubectl get nodes > ${logdir}/get-nodes.log
+kubectl describe nodes > ${logdir}/describe-nodes.log
+for pod in $(kubectl get pods -n kube-system | grep nvidia-device-plugin | awk '{print $1}'); do
+  kubectl -n kube-system logs ${pod} > ${logdir}/get-plugin-logs-${pod}.log
+done
+
+# Helm
+helm list -aA > ${logdir}/helm-list.log
+
+# Packagin
```
michael-balint, line 25: Packagin --> Packaging.

michael-balint:
This worked well for me and could be very useful when trying to determine the state of the k8s cluster.

@dholt - any thoughts/comments?

dholt (commented):
Looks good. You might think about using a prefix on the logdir, like `log_*`. Also you might want to redirect stderr to the logs as well.

supertetelman:
I made the requested documentations and also rebased the branch a bit so it doesn' pull in those other Docker config changes.

ajdecon (approved):
This looks great to me, and I think it will come in handy!

I'm approving because I don't want to block this, but here are two nice-to-haves that I'd like to see:

1. This script *implicitly* assumes it's running from the repository root. It would be good to explicitly ensure it runs correctly even if the current working dir isn't the repo root, by doing something like what [setup.sh does](https://github.com/NVIDIA/deepops/blob/master/scripts/setup.sh#L12).
1. This script assumes that the configuration directory is named `config` but DeepOps supports anything named `config*` in its gitignore. That's harder to detect automatically, but using a `CONFIG_DIR` [environment var with a default](https://github.com/NVIDIA/deepops/blob/e91a0feff99d152af5dd71cc8d72b185e6190ea6/scripts/setup.sh#L280) would also be nice.

That said, those aren't blockers! If you add these in, send me a ping and I'll approve again. If you don't have time right now, feel free to merge and I can add them in with a separate PR later. 😄

PR #795, opened 2020-12-15
Bumped version of nvidia-driver role to 2.0.0.

michael-balint:


ajdecon (changes requested):
Let's block this while we discuss versioning on the other repo.

ajdecon (approved):
LGTM

PR #798, opened 2020-12-15
Set explicit path to python interpreter

dholt:
On the latest Ubuntu 18.04.5 images from MAAS, the `/usr/bin/python` binary seems to be missing. Since we have updated to use python3, and Ubuntu explicitly points to the python2 or python3 binaries, this change updates ansible to point to the python3 binary.


ajdecon (approved):
LGTM

PR #801, opened 2020-12-16
Add default for docker_install

dholt:
Set a default for the new `docker_install` var in case one forgets to update existing config with new vars.

I'm open to NOT doing this since changing the default means changing it in all of these files in addition to the configuration. I think we already have several other variables that have this problem if one doesn't update configuration.


ajdecon (approved):
Ugh. I'm conflicted -- I don't want to have to litter playbooks with defaults everywhere, but this particular flag is a big change and I can see people forgetting to add to their configuration.

Approving but not merging yet in case others want to comment.

PR #802, opened 2020-12-16
Fixes for CentOS python configuration and required packages

ajdecon:
### Python fix for CentOS 7

We have been trying to use python3 everywhere, but this doesn't work on CentOS 7 because the package management libs only support Python 2.

This PR adds to the `bootstrap-python.yml` playbook to pre-emptively ensure the right libs are installed regardless of our auto-detected Python. This pre-empts unexpected Python errors in later playbooks.

The major downside of this approach is that on CentOS 7, we end up with an unnecessary Python 3 install. We can't easily detect the OS release ahead of time due to the way we're doing the bootstrap without assuming a Python is present. However, given that Python 3 doesn't take much disk space, and is the supported upstream Python anyway, I don't see this as a major issue.

### Misc fixes for CentOS 8

Misc fixes discovered during full testing of Python issues.

- Need build dependencies explicitly installed for NHC
- Need the powertools repo enabled for Slurm build dependencies and lmod install
- Need slightly different list of packages for slurmdbd
- Update to https://github.com/NVIDIA/ansible-role-enroot v0.3.2 for EL8 support
- Need to use python3-openshift instead of python2-openshift

## Test plan

Tested full cluster playbooks for k8s and slurm on each of CentOS 7 and CentOS 8:
- [x] k8s on centos 7
- [x] slurm on centos 7
- [x] k8s on centos 8 (but see the note below)
- [x] slurm on centos 8

Also validate that Ubuntu still works:

- [x] confirm that Jenkins passes for Ubuntu

### Kubernetes on CentOS 8.3

The CentOS 8.3 kernel `4.18.0-240.1.1.el8_3.x86_64` no longer works with Kubespray due to not being able to load a netfilter kernel module (`nf_conntrack_ipv4`):

```
TASK [kubernetes/node : Modprobe nf_conntrack_ipv4 for kernels < 4.19] ***********************************************************************************************************************************************************************
fatal: [virtual-mgmt01]: FAILED! => {"changed": false, "msg": "modprobe: FATAL: Module nf_conntrack_ipv4 not found in directory /lib/modules/4.18.0-240.1.1.el8_3.x86_64\n", "name": "nf_conntrack_ipv4", "params": "", "rc": 1, "state": "present", "stderr": "modprobe: FATAL: Module nf_conntrack_ipv4 not found in directory /lib/modules/4.18.0-240.1.1.el8_3.x86_64\n", "stderr_lines": ["modprobe: FATAL: Module nf_conntrack_ipv4 not found in directory /lib/modules/4.18.0-240.1.1.el8_3.x86_64"], "stdout": "", "stdout_lines": []}
```

This is fixed upstream but not yet in a Kubespray release: https://github.com/kubernetes-sigs/kubespray/pull/6988 . We should plan to update to the latest Kubespray when they release next in 2021.

I confirmed that DeepOps succeeds by explicitly running on CentOS 8.2 which does not have the problematic kernel.

### DIFF
--- a/playbooks/bootstrap/bootstrap-python.yml
+++ b/playbooks/bootstrap/bootstrap-python.yml
@@ -36,7 +36,7 @@
         state: present
       when: (ansible_python.version.major == 2) and (ansible_os_family == "RedHat") and (ansible_distribution_major_version == "7")
 
-    - name: install python 2 libraries on RHEL/CentOS 7
+    - name: install python 2 libraries on EL7
       package:
         name:
         - python-setuptools
--- a/playbooks/bootstrap/bootstrap-python.yml
+++ b/playbooks/bootstrap/bootstrap-python.yml
@@ -1,4 +1,5 @@
 ---
+# Initial bootstrap of Python 3 installation on all hosts
 - hosts: all
   become: true
   gather_facts: false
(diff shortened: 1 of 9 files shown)


ajdecon:
Note that nightly CentOS test completed necessary Ansible steps successfully. (See job deepops-nightly/672 in Jenkins server.)

supertetelman (approved):
LGTM and passing Jenkins

PR #807, opened 2020-12-18
Attempt to address different showmount path in centos/ubuntu for jenkins

ajdecon:
Our CentOS nightly run in Jenkins is failing on `test-slurm-nfs-mount.sh` because it can't find `showmount`:

```
debug1: Sending command: showmount -e | grep home
bash: showmount: command not found
```

I suspect the issue here is that `showmount` isn't in the default PATH for this SSH command. Unfortunately, Ubuntu and CentOS put `showmount` in different locations...

This change to the test adds a conditional for the CentOS case and swaps to use the absolute path `/usr/sbin/showmount` in the CentOS case.

Unfortunately this only shows up in our nightly Jenkins test so this is difficult to test 😢

### DIFF
--- a/workloads/jenkins/scripts/test-slurm-nfs-mount.sh
+++ b/workloads/jenkins/scripts/test-slurm-nfs-mount.sh
@@ -2,13 +2,26 @@
 set -e
 
 source workloads/jenkins/scripts/jenkins-common.sh
-ssh -v \
-	-o "StrictHostKeyChecking no" \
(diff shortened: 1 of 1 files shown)


ajdecon:
Please don't merge until we confirm this still passes the per-PR test on Ubuntu.

supertetelman (approved):
LGTM, let's merge through after `deepops-nightly/681/` completes.

Issue #809, opened 2020-12-25
registry-mirror: docker-options.conf is invalid


michaelfuckner:
When setting up a default config, the option registry-mirror in docker-options.conf is invalid, preventing docker from starting

 cat /etc/systemd/system/docker.service.d/docker-options.conf
[Service]
Environment="DOCKER_OPTS= --iptables=True \
--exec-opt native.cgroupdriver=systemd \
--insecure-registry=login01:5000:5000 --insecure-registry=registry.local:31500  \
--registry-mirror=http://login01:5000http://:5000  \
--data-root=/var/lib/docker \
--log-opt max-size=50m --log-opt max-file=5 \
"

ajdecon:
Unfortunately, I'm not able to duplicate this locally. When I run a default configuration on a new virtual development cluster, I get this:

```
vagrant@virtual-gpu01:~$ cat /etc/systemd/system/docker.service.d/docker-options.conf
[Service]
Environment="DOCKER_OPTS= --iptables=True \
--exec-opt native.cgroupdriver=systemd \
--insecure-registry=virtual-login01:5000 --insecure-registry=registry.local:31500  \
--registry-mirror=http://virtual-login01:5000  \
--data-root=/var/lib/docker \
--log-opt max-size=50m --log-opt max-file=5 \
"
```

The configuration for registry-mirror and insecure-registry are generated by default using the list of machines in the `slurm-master` group, i.e.

```
docker_insecure_registries: "{{ groups['slurm-master'] | map('regex_replace', '(.*)', '\\1:5000') | list + ['registry.local:31500'] }}"
docker_registry_mirrors: "{{ groups['slurm-master'] | map('regex_replace', '(.*)', 'http://\\1:5000') | list }}"
```

It looks like somehow an extra "empty" host is being included when these are being generated. I'm not sure how that could happen with an Ansible inventory 😄 but it might be worth checking your inventory to see if it includes any unexpected entries for that group.

If you still see issues, feel free to post your inventory file and I can try to help debug. You can also likely work around this by overriding the `docker_insecure_registries` and `docker_registry_mirrors` variables directly.

michaelfuckner:
I just did a very basic installation: DGX A100 (DGX OS 5) with login Server (login01, Ubuntu 20.04 LTS default installation). Doesn't even look like a CR-LF vs CR issue. Is my setup correct or is there any advice to use a DGX with multiple users?


Here is what I set up
/etc/hosts
192.168.1.1: login01
[...]

PR #818, opened 2021-01-06
Fix regex replace for docker-options.conf

andiariffin:
This PR is to address #809.

Signed-off-by: Andi Ariffin <andi@novaglobal.com.sg>


ajdecon (approved):
Passes CI, LGTM. Thanks for figuring this out!

PR #820, opened 2021-01-07
Add playbook for setting up slurm login on compute.

avolkov1:
A playbook to enable slurm login functionality on a compute node. A typical use case would be single node slurm deployment.
TODO: Add doc explaining when and how to run this playbook.


arnoldas500:
Has anyone gotten this to work? Would be interested in using the DGX as a single compute and login node all in one.

avolkov1:
Yes this works, but it's still a work in progress. I'll post a document in this PR showing the ansible commands to run and enable this functionality.

This is meant for multiple users sharing a single GPU server such as a DGX server. It could also be used with multiple nodes where you for some reason do not have another non-GPU server to act as a slurm-manager and login node.

Are you currently using DeepOps?

arnoldas500:
> @arnoldas500
> 
> Yes this works, but it's still a work in progress. I'll post a document in this PR showing the ansible commands to run and enable this functionality.
> 
> This is meant for multiple users sharing a single GPU server such as a DGX server. It could also be used with multiple nodes where you for some reason do not have another non-GPU server to act as a slurm-manager and login node.
> 
> Are you currently using DeepOps?

Yes I am. I was just able to get slurm working with cuda 11.0 with dgx server running 20.04 and using the storage node as the login node.

roles/nvidia-hpc-sdk/tasks/main.yml
```diff
@@ -30,6 +30,7 @@
     url: "{{ hpcsdk_download_url }}"
     dest: "{{ hpcsdk_dest_download_path }}"
     mode: "0444"
+    validate_certs: no
```
ajdecon, line 33: What problem did you run into here? It's generally a bad idea to not validate certs, and the HPC SDK download site should have valid certs.

roles/nvidia-dcgm-exporter/templates/docker.dcgm-exporter.service.j2
```diff
@@ -9,7 +9,7 @@ Restart=always
 ExecStartPre=-/usr/bin/docker stop %n
 ExecStartPre=-/usr/bin/docker rm %n
 ExecStartPre=/usr/bin/docker pull {{ nvidia_dcgm_container }}
-ExecStart=/usr/bin/docker run --rm --gpus all --cpus="{{ nvidia_dcgm_max_cpu }}" --name %n -p 9400:9400 {{ nvidia_dcgm_container }}
+ExecStart=/usr/bin/docker run --rm --gpus all --cpus={{ nvidia_dcgm_max_cpu }} --name %n -p 9400:9400 {{ nvidia_dcgm_container }}
```
ajdecon, line 12: Why drop the quotation marks here?

roles/prometheus-node-exporter/templates/docker.node-exporter.service.j2
```diff
@@ -9,7 +9,7 @@ Restart=always
 ExecStartPre=-/usr/bin/docker stop %n
 ExecStartPre=-/usr/bin/docker rm %n
 ExecStartPre=/usr/bin/docker pull {{ node_exporter_container }}
-ExecStart=/usr/bin/docker run --rm --network host --cpus="{{ node_exporter_max_cpu }}"  --pid="host" --name %n -v {{ node_exporter_prom_dir }}:/run/prometheus {{ node_exporter_container }} --collector.textfile.directory="{{ node_exporter_prom_dir }}"
+ExecStart=/usr/bin/docker run --rm --network host --cpus={{ node_exporter_max_cpu }}  --pid=host --name %n -v {{ node_exporter_prom_dir }}:/run/prometheus {{ node_exporter_container }} --collector.textfile.directory="{{ node_exporter_prom_dir }}"
```
ajdecon, line 12: Similar, why drop the quotation marks?

roles/slurm/tasks/controller.yml
```diff
@@ -39,6 +39,7 @@
     target: '/lib64(/.*)?'
     setype: mysqld_db_t
     state: present
+    ignore_selinux_state: yes
```
ajdecon, line 42: Ignoring the selinux state seems like a bad idea. The Ansible docs indicate this is supposed to be used in scenarios like a `chroot` environment where we can't get the state in the first place. Why are we doing this here?

roles/slurm/tasks/login-compute-setup.yml
```diff
@@ -0,0 +1,32 @@
...
+    fi
+  args:
+    creates: "{{ '/etc/systemd/system.control/sshd.service.d/50-DeviceAllow.conf' \
+                 if ansible_os_family == 'RedHat' else \
+                 '/etc/systemd/system.control/ssh.service.d/50-DeviceAllow.conf' }}"
+  when: is_controller
+  tags:
+    - config
+
+- name: Patch lastuserjob epilog.
```
ajdecon, line 18: It seems to me that the users in `/etc/slurm/localusers.backup` should always be allowed to log in anyway. 😄 Why not just modify the prolog script provided by DeepOps, instead of patching it on the fly for this use case?

ajdecon (changes requested):
Overall this looks really good, and resulted in a working single-node setup for me. Nice work!

I'm requesting changes (with comments inline) only because there seem to be some changes not related to the "login on compute" setup. For example, disabling cert validation when downloading the HPC SDK. It would be good to know what problems you ran into that led to these changes; if they're real issues, they probably belong in a separate PR.

<hr/>

### A note on reboots

When running with a local Ansible connection, how do you handle the compute node reboot after the grub update in `roles/slurm/tasks/compute.yml`, or the reboot after installing the NVIDIA driver? 

If the answer is "reboot the node manually and then re-run Ansible", then that should be specified in the doc. I can definitely see this issue confusing folks new to Ansible.

<hr/>

### Problem outside the scope of the PR

I did run into some problems with python packages when running with `ansible_connection=local`, but I think these are related to the new setup script introduced in #912 which uses a virtual environment for DeepOps. Noting this here as I had to manually install a few python packages outside the scope of the virtual environment, but that should probably be fixed in a separate PR.

roles/nvidia-hpc-sdk/tasks/main.yml
avolkov1: I was helping a client run these scripts and on their system it refused to download the HPC SDK until the validate certs setting was set to no. I can remove this.

roles/nvidia-dcgm-exporter/templates/docker.dcgm-exporter.service.j2
avolkov1: In my testing on RHEL8 for some reason systemd did not like quotes around those parameters. It would error out on that docker command complaining about the `--cpus` option and in another case about the `--pid` option. I finally got it working when I removed quotes. The docker commands themselves worked fine, just within systemd running these docker commands RHEL8 would choke on these particular quote parameters.

The values don't need quotes since there's not spaces or conventional delimiters in the parameters so it works fine without quotes on Ubuntu as well. But again on RHEL8 it refused to work. 

Again, this is not related to the PR per se but I had to fix this for a client's install and since I did it in the "single-node-slurm" branch I just committed these changes there.

roles/prometheus-node-exporter/templates/docker.node-exporter.service.j2
avolkov1: In my testing on RHEL8 for some reason systemd did not like quotes around those parameters. It would error out on that docker command complaining about the `--cpus` option and in another case about the `--pid` option. I finally got it working when I removed quotes. The docker commands themselves worked fine, just within systemd running these docker commands RHEL8 would choke on these particular quote parameters.

The values don't need quotes since there's not spaces or conventional delimiters in the parameters so it works fine without quotes on Ubuntu as well. But again on RHEL8 it refused to work. 

Again, this is not related to the PR per se but I had to fix this for a client's install and since I did it in the "single-node-slurm" branch I just committed these changes there.

roles/slurm/tasks/controller.yml
avolkov1: I forgot why I had to set this. Maybe it was also for a clients setup that refused to work unless this setting was added. I can remove this.

roles/slurm/tasks/login-compute-setup.yml
avolkov1: Unless I patch epilog `40-lastuserjob-processes` for the caveat of `localusers.backup` all the processes on the node for the user get killed. The user's session gets killed as well so they are logged out of all their "login" sessions not just slurm reservation session.

https://github.com/NVIDIA/deepops/blob/552f7780c878ef4403f3cfbd38ee51d7cae1f262/roles/slurm/templates/etc/slurm/epilog.d/40-lastuserjob-processes#L4-L8

avolkov1:
> A note on reboots
When running with a local Ansible connection, how do you handle the compute node reboot after the grub update in roles/slurm/tasks/compute.yml, or the reboot after installing the NVIDIA driver?
>
> If the answer is "reboot the node manually and then re-run Ansible", then that should be specified in the doc. I can definitely see this issue confusing folks new to Ansible.

Yes, the reboot part is not very streamlined. Frankly I just keep re-running ansible until everything seems to complete to the end. Then I reboot at the very end. When re-running there are sometimes messages about rebooting, but it's unclear to me exactly when a reboot is needed. So I just keep re-running until everything seems to be installed and reboot manually once at the end. I would say the answer is "reboot the node manually and then re-run Ansible", but it's not a very clean process. Even without this setup of login on compute, I find the reboot part confusing.

roles/slurm/tasks/login-compute-setup.yml
ajdecon: Got it. I'd still prefer we modify the file copied in by Ansible, rather than patching it in place, but I don't think that should be a blocker for the PR.

roles/prometheus-node-exporter/templates/docker.node-exporter.service.j2
ajdecon: Got it, ok. Feel free to leave this in, and we can test later on RHEL8 to better understand the issue.

roles/nvidia-dcgm-exporter/templates/docker.dcgm-exporter.service.j2
ajdecon: Got it, ok. Feel free to leave this in, and we can test later on RHEL8 to better understand the issue.

roles/nvidia-hpc-sdk/tasks/main.yml
ajdecon: Yeah, that sounds like a problem with their system configuration. In general it's a bad idea to not validate certs. 😄

avolkov1: Then I'll remove this.

roles/slurm/tasks/controller.yml
avolkov1: I'll remove this too.

ajdecon:
> So I just keep re-running until everything seems to be installed and reboot manually once at the end. I would say the answer is "reboot the node manually and then re-run Ansible", but it's not a very clean process. Even without this setup of login on compute, I find the reboot part confusing.

In the default Slurm playbook, we reboot twice:

* Once after installing the NVIDIA driver, because the driver sometimes requires a reboot to load correctly
* Once after setting some grub options used for Slurm compute nodes to configure cgroups correctly, because we're changing the kernel command line

It's generally not a problem when the compute nodes are separate from the cluster head node or Ansible controller, but of course it can cause problems when running with a local connection!

I think it's fine to use either the "run Ansible repeatedly then reboot workflow", or the "reboot when asked to then run Ansible again". They will both get to the same place in the end. We should just document one or the other in this setting where we expect the user to run with a local connection.

roles/slurm/tasks/login-compute-setup.yml
avolkov1: In the general case where you might have multinode setup, you don't want to patch on compute nodes that are not used as login nodes. So I patch after the epilog is copied. The "playbooks/slurm-cluster/slurm.yml" has a specific order.

1. The epilog files are not copied yet.
```
# Set up slurm controller
- hosts: slurm-master
. . .
        is_controller: yes
```

2. Now the epilog files are copied to all the compute nodes.
```
# Set up slurm nodes
- hosts: slurm-node
. . .
        is_compute: yes
```

3. If a compute node was set as a login node it will have the epilog files after step 2 above. Now patch the epilog files for login node(s). 
```
# Set up slurm-login-on-compute if enabled
- hosts: slurm-master
. . .
        is_controller: yes
      when: slurm_login_on_compute | default(false)
```

roles/slurm/tasks/controller.yml
avolkov1: Fixed in commit 42a8792

roles/nvidia-hpc-sdk/tasks/main.yml
avolkov1: Fixed in commit 42a8792.

ajdecon (approved):
LGTM

PR #821, opened 2021-01-08
set dns_mode: none when using hosts

michaelfuckner:
Signed-off-by: Michael Fuckner <michael@fuckner.net>

### DIFF
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -40,6 +40,7 @@ slurm_ha_state_save_location: "/sw/slurm"
 # Optional installs                                                            #
 ################################################################################
 slurm_configure_etc_hosts: yes
+dns_mode: none
 slurm_cluster_install_cuda: yes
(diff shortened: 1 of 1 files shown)


ajdecon (changes requested):
@michaelfuckner : Can you please provide some more detail for this change? It's not clear why you think this should go in the default configuration for the slurm-cluster playbook.

In particular, it would help if you could:

* Add some detail to the PR on why you'd like to do this
* Add some detail to the PR on how you tested this change
* Add a short comment to `config.example/group_vars/slurm-cluster.yml` for this line to explain it, as it's not terribly obvious what it does unless you're familiar with Kubespray.

michaelfuckner:
it is a fix for issue 783. If you don't have a separate DNS Server, docker fails to start. For Slurm it uses hosts anyway, so I guess it should be configured to not use DNS to get a working setup if you simply run slurm-playbook. Or what are the prerequisites to install slurm via playbook?

ajdecon (approved):

PR #824, opened 2021-01-12
Rootless docker.

avolkov1:
Add ansible scripts to install rootless docker and add a module file for it. Rootless docker can be run within a Slurm cluster unprivileged and respects resource constraints i.e. CPU, GPU allocations etcetera.

TODO: Add documentation.

### DIFF
--- a/playbooks/container/docker-rootless.yml
+++ b/playbooks/container/docker-rootless.yml
@@ -1,0 +1,5 @@
+---
+- hosts: all
+  become: yes
+  roles:
+    - name: docker-rootless
--- a/playbooks/container/docker-rootless.yml
+++ b/playbooks/container/docker-rootless.yml
@@ -1,18 +1,15 @@
 ---
+- include: ../nvidia-software/nvidia-driver.yml
+
+- include: nvidia-docker.yml
+  vars:
+    nvidia_docker_skip_docker_reload: true
+
+- include: ../slurm-cluster/lmod.yml
+
 - hosts: all
   become: yes
   tasks:
-    - name: install nvidia-docker
-      include_role:
-        name: nvidia.nvidia_docker
-      vars:
-        nvidia_docker_skip_docker_reload: true
-
-    - name: install lmod
-      include_role:
-        name: lmod
-
     - name: install rootless docker
       include_role:
         name: docker-rootless
-
--- a/playbooks/container/docker-rootless.yml
+++ b/playbooks/container/docker-rootless.yml
@@ -5,6 +5,8 @@
     - name: install nvidia-docker
       include_role:
         name: nvidia.nvidia_docker
+      vars:
+        nvidia_docker_skip_docker_reload: true
 
     - name: install lmod
       include_role:
--- a/playbooks/container/docker-rootless.yml
+++ b/playbooks/container/docker-rootless.yml
@@ -1,5 +1,16 @@
 ---
 - hosts: all
   become: yes
-  roles:
-    - name: docker-rootless
+  tasks:
+    - name: install nvidia-docker
+      include_role:
+        name: nvidia.nvidia_docker
+
+    - name: install lmod
+      include_role:
+        name: lmod
+
+    - name: install rootless docker
+      include_role:
+        name: docker-rootless
+
--- a/roles/docker-rootless/tasks/main.yml
+++ b/roles/docker-rootless/tasks/main.yml
@@ -1,0 +1,73 @@
+---
+
+- name: install dependencies (ubuntu)
+  apt:
+    name:
+    - "uidmap"
+    state: present
+  when: ansible_distribution == "Ubuntu"
+
+- name: install dependencies (Red Hat)
+  yum:
+    name:
+    - "shadow-utils"
+    - "fuse-overlayfs"
+    state: present
+  when: ansible_os_family == "RedHat"
+
+- name: Ensure directory structure exists
+  file:
+    path: '{{ rootlessdocker_install_dir }}/{{ item.path }}'
+    owner: 'root'
+    group: 'root'
+    state: directory
+    mode: '0755'
+  with_filetree: '{{ role_path }}/templates/rootless-docker'
+  when: item.state == 'directory'
+
+- name: Copy over templates.
+  template:
+    src: '{{ item.src }}'
+    dest: '{{ rootlessdocker_install_dir }}/{{ item.path }}'
+    owner: 'root'
+    group: 'root'
+    mode: '{{ item.mode }}'
+    force: no
+  with_filetree: '{{ role_path }}/templates/rootless-docker'
+  when: item.state == 'file'
+
+- name: add rootless docker profile script to source modules
+  template:
+    src: '{{ item }}'
+    dest: '/etc/profile.d/{{ item }}'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
+    force: no
+  with_items:
+  - 'z96_rootlessdocker_modules.sh'
+
+- name: check if rootless docker already installed successfully
+  stat: 
+    path: '{{ rootlessdocker_install_dir }}/config/install_success'
+  register: rootless_docker_installed
+
+- name: Template setup rootless docker script
+  template:
+    src: '{{ role_path }}/templates/rootless_docker_setup.sh.j2'
+    dest: /tmp/rootless_docker_setup.sh
+    mode: 0755
+  when: not rootless_docker_installed.stat.exists
+
+- name: Execute setup rootless docker script
+  become: true
+  shell:
+    /tmp/rootless_docker_setup.sh
+  args:
+    executable: /bin/bash
+    creates: '{{ rootlessdocker_install_dir }}/config/install_success'
+
+- name: Clean up rootless docker set script
+  file:
+    path: /tmp/rootless_docker_setup.sh
+    state: absent
--- a/roles/docker-rootless/tasks/main.yml
+++ b/roles/docker-rootless/tasks/main.yml
@@ -71,3 +71,14 @@
   file:
     path: /tmp/rootless_docker_setup.sh
     state: absent
+
+- name: Add the br_netfilter module
+  modprobe:
+    name: br_netfilter
+    state: present
+
+- name: Load br_netfilter module at boot
+  lineinfile:
+    path: /etc/modules
+    create: yes
+    line: "br_netfilter"
--- a/roles/docker-rootless/templates/rootless_docker_setup.sh.j2
+++ b/roles/docker-rootless/templates/rootless_docker_setup.sh.j2
@@ -49,6 +49,11 @@ fi
 curl -fsSL https://get.docker.com/rootless | \
   SKIP_IPTABLES=1 FORCE_ROOTLESS_INSTALL=1 DOCKER_BIN="{{ rootlessdocker_install_dir }}/bin" sh
 
+installsuccess=false
+if [ $? -eq 0 ] ; then
+  installsuccess=true
+fi
+
 if [ "$cleansubuid" = true ] ; then
   sed -i "/^${userid}:/d" /etc/subuid
 fi
@@ -57,7 +62,7 @@ if [ "$cleansubgid" = true ] ; then
   sed -i "/^${userid}:/d" /etc/subgid
 fi
 
-if [ $? -eq 0 ] ; then
+if [ "$installsuccess" = true ] ; then
   touch "{{ rootlessdocker_install_dir }}/config/install_success"
 else
   echo curl -fsSL https://get.docker.com/rootless \| \
--- a/roles/docker-rootless/templates/rootless_docker_setup.sh.j2
+++ b/roles/docker-rootless/templates/rootless_docker_setup.sh.j2
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+{% if ansible_os_family == "RedHat" %}
+
+mun=$(sysctl -n user.max_user_namespaces 2>/dev/null || echo 0 )
+
+if [ "$mun" -lt "28633" ] ; then
+
+cat <<EOT > /etc/sysctl.d/51-rootless.conf
+user.max_user_namespaces = 28633
+EOT
+sudo sysctl --system
+
+fi
+
+# On RedHat the following might be needed:
+# grubby --update-kernel=ALL --args="namespace.unpriv_enable=1 user_namespace.enable=1"
+
+{% endif %}
+
+configfiledst="{{ rootlessdocker_install_dir }}/config/nvidia-container-runtime/config.toml"
+configfilesrc=/etc/nvidia-container-runtime/config.toml
+
+if [ -f "$configfilesrc" ]; then
+  cp $configfilesrc $configfiledst
+  # delete and set no-cgroups to true
+  sed -i '/^\[nvidia-container-cli\]$/,/^\[/{/no-cgroups/d}' $configfiledst
+  sed -i '/^\[nvidia-container-cli\]$/a no-cgroups = true' $configfiledst
+fi
+
+
+userid=$(id -u)
+largestuid=$(cat /etc/subuid | awk '{split($0,a,":"); print a[2]}' | sort -n | tail -1)
+mysubuid=$(( largestuid + 65536 ))
+cleansubuid=false
+if ! grep "^$(id -un):\|^$(id -u):" /etc/subuid >/dev/null 2>&1; then
+  echo ${userid}:${mysubuid}:65536 >> /etc/subuid
+  cleansubuid=true
+fi
+
+largestgid=$(cat /etc/subgid | awk '{split($0,a,":"); print a[2]}' | sort -n | tail -1)
+mysubgid=$(( largestgid + 65536 ))
+cleansubgid=false
+if ! grep "^$(id -un):\|^$(id -u):" /etc/subgid >/dev/null 2>&1; then
+  echo ${userid}:${mysubgid}:65536 >> /etc/subgid
+  cleansubgid=true
+fi
+
+curl -fsSL https://get.docker.com/rootless | \
+  SKIP_IPTABLES=1 FORCE_ROOTLESS_INSTALL=1 DOCKER_BIN="{{ rootlessdocker_install_dir }}/bin" sh
+
+if [ "$cleansubuid" = true ] ; then
+  sed -i "/^${userid}:/d" /etc/subuid
+fi
+
+if [ "$cleansubgid" = true ] ; then
+  sed -i "/^${userid}:/d" /etc/subgid
+fi
+
+if [ $? -eq 0 ] ; then
+  touch "{{ rootlessdocker_install_dir }}/config/install_success"
+else
+  echo curl -fsSL https://get.docker.com/rootless \| \
+    SKIP_IPTABLES=1 FORCE_ROOTLESS_INSTALL=1 DOCKER_BIN="{{ rootlessdocker_install_dir }}/bin" sh
+  exit 1
+fi
+
+chown root:root {{ rootlessdocker_install_dir }}/bin/*
--- a/docs/container/docker-rootless.md
+++ b/docs/container/docker-rootless.md
@@ -1,0 +1,80 @@
+# Rootless Docker
+
+Rootless docker enables unprivileged end-users to utilize docker containers.
+One motivation for this is that typically it is easier to manage unprivileged
+users within HPC clusters such as SLURM. Alternatives, such as Singularity and
+Enroot, also enable an unprivileged user to utilize containerization. The added
+convenience of rootless docker is that if one is working with docker containers
+it would be more straightforward to build and run docker containers instead of
+having to convert to another container format.
+
+Further information about rootless docker can be found here:
+<https://docs.docker.com/engine/security/rootless/>
+
+
+## Installing Rootless Docker
+
+An ansible role and playbook are provided to install rootless docker
+[docker-rootless.yml](../../playbooks/container/docker-rootless.yml).
+The default install location is set via variable `rootlessdocker_install_dir`
+with default setting of `/sw/software/rootless_docker`. Set the variable to a
+different value (use `--extra-vars "rootlessdocker_install_dir=<somepath>"`) if
+a different install location is desired. Run the playbook via ansible command:
+```bash
+ansible-playbook -K --limit slurm-cluster playbooks/container/docker-rootless.yml
+```
+
+## Using Rootless Docker in Slurm
+
+The playbook installs an environment module for loading rootless docker via
+modules. Users with bash shells will see the following modules in a typical
+DeepOps deployed Slurm cluster:
+```
+$ module avail
+
+-------------------------------------------------------------------------- /sw/hpc-sdk/modulefiles ----------------------------------
+   nvhpc/20.7    nvhpc/20.9 (D)    nvhpc-byo-compiler/20.7    nvhpc-byo-compiler/20.9 (D)    nvhpc-nompi/20.7    nvhpc-nompi/20.9 (D)
+
+------------------------------------------------------------------ /sw/software/rootless-docker/modulefiles -------------------------
+   rootless-docker
+
+  Where:
+   D:  Default Module
+```
+
+Once the rootless docker module is loaded scripts "start_rootless_docker.sh"
+and "stop_rootless_docker.sh" will become available (added via PATH). Use
+these scripts to start an stop rootless docker daemon. The commands below
+illustrate how to start and run a rootless docker container on Slurm.
+```
+$ srun --ntasks=1 --gpus-per-task=1 --cpus-per-task=5 --gres-flags=enforce-binding --pty bash
+
+$ module load rootless-docker
+
+$ start_rootless_docker.sh # specify --quiet option to hide rootles docker messages
+
+$ docker run --gpus all -it --rm nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
+
+root@445bf5cca686:/# echo NGPUS: $(nvidia-smi -L | wc -l)
+NGPUS: 1
+
+root@445bf5cca686:/# exit # exit docker interactive sessions
+exit
+
+$ stop_rootless_docker.sh # Optionally explicitly stop rootless docker. Will be killed on Slurm exit regardless.
+
+$ exit # end slurm session
+exit
+
+```
+
+It is important to be aware that rootless docker runs via user namespace
+remapping. Therefore even though one might appear as root in a docker
+container, files created are owned by the user, and within the container a user
+would not have write/execute permissions to filesystem or executables that the
+user already does not have permission to outside of the container.
+
+
+
+
+
--- a/docs/container/docker-rootless.md
+++ b/docs/container/docker-rootless.md
@@ -24,6 +24,19 @@ a different install location is desired. Run the playbook via ansible command:
 ansible-playbook -K --limit slurm-cluster playbooks/container/docker-rootless.yml
 ```
 
+For networking to properly work with rootless docker the `br_netfilter` kernel
+module should be loaded. Check via:
+```
+$ lsmod | grep br_netfilter
+```
+The `br_netfilter` kernel module is typically installed by default on various
+Linux platforms. If not loaded then load via:
+```
+$ sudo modprobe br_netfilter
+# to load by default on bootup
+$ sudo sh -c 'echo "br_netfilter" > /etc/modules-load.d/br_netfilter.conf'
+```
+
 ## Using Rootless Docker in Slurm
 
 The playbook installs an environment module for loading rootless docker via
--- a/roles/docker-rootless/defaults/main.yml
+++ b/roles/docker-rootless/defaults/main.yml
@@ -2,4 +2,6 @@
 # Install for rootless docker.
 
 # Directory to install in
+sm_prefix: "/sw"
+sm_software_path: "{{ sm_prefix }}/software"
 rootlessdocker_install_dir: "{{ sm_software_path }}/rootless-docker"
--- a/roles/docker-rootless/defaults/main.yml
+++ b/roles/docker-rootless/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+# Install for rootless docker.
+
+# Directory to install in
+rootlessdocker_install_dir: "{{ sm_software_path }}/rootless-docker"
--- a/roles/docker-rootless/templates/rootless-docker/bin/nvidia-container-runtime-hook
+++ b/roles/docker-rootless/templates/rootless-docker/bin/nvidia-container-runtime-hook
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+/usr/bin/nvidia-container-runtime-hook -config="{{ rootlessdocker_install_dir }}/config/nvidia-container-runtime/config.toml" "$@"
--- a/roles/docker-rootless/templates/rootless-docker/bin/start_rootless_docker.sh
+++ b/roles/docker-rootless/templates/rootless-docker/bin/start_rootless_docker.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+quiet=false
+
+usage() {
+cat <<EOF
+Usage: $(basename $0) [-h|--help] [--quiet]
+    Start rootless docker daemon.
+    Example:
+        $(basename $0)
+
+    To ommit rootless docker daemon messages redirect output to dev null or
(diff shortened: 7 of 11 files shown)


supertetelman:
I don't see any documentation in this script. I know we are doing a devblog on this, we might want to add a short README that either points to the devblog or copies key parts or both.

avolkov1:
> I don't see any documentation in this script. I know we are doing a devblog on this, we might want to add a short README that either points to the devblog or copies key parts or both.

I just added a doc: `docs/container/docker-rootless.md`

I want to test this again. It's almost ready for review. I need to figure out apt dependencies and add that to the role's tasks. Rootless docker requires "uidmap" package. Or should I leave the dependencies up to end users to install on the nodes.

avolkov1:
Could someone test and try this PR. For rootless docker you need namespace remapping. I rely on prologs and epilogs scripts in Slurm to add a user to `/ets/subuid` and `/ets/subgid`. Otherwise rootless docker daemon will not start. I did not add these prolog/epilog scripts in this PR. Those can be found in my other #820.
https://github.com/avolkov1/deepops/blob/single-node-slurm/roles/slurm/templates/etc/slurm/prolog.d/95-all-rootless
https://github.com/avolkov1/deepops/blob/single-node-slurm/roles/slurm/templates/etc/slurm/epilog.d/95-lastuserjob-rootless

Refer to the docs "docs/container/docker-rootless.md" section "Using Rootless Docker in Slurm" for testing.

I am curious what issues one might run in. One potential improvement I could make is to specify in ansible that `uidmap` package should be installed. But there could be other package dependencies. The rootless docker install script will print out a message and instructions which additional packages to install if those are missing. I installed those independently of the rootless docker playbook, but I can add that to the playbook.

dholt:
I think in order for us to merge this, all of the prerequisites should be included so it works out of the box. Is there a reason not to include the `uidmap` package and prolog/epilog scripts?

playbooks/container/docker-rootless.yml
```diff
@@ -0,0 +1,7 @@
+---
+- hosts: all
+  become: yes
+  roles:
+    - name: docker-rootless
+      vars:
+        rootlessdocker_install_dir: "{{ sm_software_path }}/rootless-docker"
```
dholt, line 7: This should be removed since it exists as a default in the role. Our convention is also to add any default variables to the appropriate `config.example/group_vars` files.

roles/docker-rootless/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    dest: '/etc/profile.d/{{ item }}'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
+    force: no
+  with_items:
+  - 'z96_rootlessdocker_modules.sh'
+
+- name: install rootless docker
+  shell: |
```
dholt, line 36: This all needs to be written in Ansible as much as possible

avolkov1:
> I think in order for us to merge this, all of the prerequisites should be included so it works out of the box. Is there a reason not to include the `uidmap` package and prolog/epilog scripts?

I did not add the prolog/epilog scripts in this PR, because rootless docker functionality is not intrinsically related to Slurm i.e. rootless docker does not need Slurm nor does Slurm need rootless docker. Out of convenience I can add the scripts here though. Should I have the playbook copy those scripts to the Slurm prolog/epilog directories?

The uidmap I did not include initially, because if I include that package I probably should include a variety of other packages. And then I should deal with Ubuntu and RHEL options. I was thinking to leave that responsibility to the admin who is installing rootless docker.
I'll add the package and make another commit.

dholt:
I think the subuid/subgid prolog/epilog changes should go in the slurm role (we used to set them up already in deepops but i think it was removed). I agree this is independent of slurm, and the subuid/subgid can be independent of this role.

roles/docker-rootless/tasks/main.yml
avolkov1: I cannot completely get away from running some sort of command/shell to run the downloaded rootless docker installation script.

I can refactor the inlined shell script with a separate task to use `get_url` instead of curl within the script. But that's one line and with a separate task it becomes several lines. I'll do that if that's a better convention.

The other parts are not so easy to implement directly with Ansible. The subuid/subgid logic is temporary otherwise the rootless docker install script will fail. It's simple to do with shell script, a couple of lines. Otherwise it's somewhat complicated with Ansible. I found an example here:
https://github.com/debops/debops/blob/d1ed2f31d9816637e773033982c33c511c870edc/ansible/roles/root_account/templates/etc/ansible/facts.d/root_account.fact.j2#L45

There's no builtin Ansible feature for subuid/subgid that I found (I did find a feature request for this functionality).

playbooks/container/docker-rootless.yml
avolkov1: Will do.

avolkov1:
> I think the subuid/subgid prolog/epilog changes should go in the slurm role (we used to set them up already in deepops but i think it was removed). I agree this is independent of slurm, and the subuid/subgid can be independent of this role.

I started PR #820 with those prolog/epilog changes.

roles/docker-rootless/tasks/main.yml
avolkov1: I don't think this is better because it is more verbose and I have to keep track of tasks inter-dependencies. I can refactor to something like this (I haven't tested this):
```
- name: check if rootless docker already installed successfully
  stat: 
    path: "{{ rootlessdocker_install_dir }}/config/install_success"
  register: rootless_docker_installed

- name: Download rootless docker install script
  get_url:
    url="https://get.docker.com/rootless"
    dest=/tmp/
    mode: '0755'
  register: get_rootless_docker
  when: not rootless_docker_installed.stat.exists

- name: install rootless docker
  shell: |
    userid=$(id -u)
    largestuid=$(cat /etc/subuid | awk '{split($0,a,":"); print a[2]}' | sort -n | tail -1)
    mysubuid=$(( largestuid + 65536 ))
    cleansubuid=false
    if ! grep "^$(id -un):\|^$(id -u):" /etc/subuid >/dev/null 2>&1; then
      echo ${userid}:${mysubuid}:65536 >> /etc/subuid
      cleansubuid=true
    fi

    largestgid=$(cat /etc/subgid | awk '{split($0,a,":"); print a[2]}' | sort -n | tail -1)
    mysubgid=$(( largestgid + 65536 ))
    cleansubgid=false
    if ! grep "^$(id -un):\|^$(id -u):" /etc/subgid >/dev/null 2>&1; then
      echo ${userid}:${mysubgid}:65536 >> /etc/subgid
      cleansubgid=true
    fi

    FORCE_ROOTLESS_INSTALL=1 DOCKER_BIN="{{ rootlessdocker_install_dir }}/bin" /tmp/rootless

    if [ $? -eq 0 ] ; then
      touch "{{ rootlessdocker_install_dir }}/config/install_success"
    fi

    chown root:root {{ rootlessdocker_install_dir }}/bin/*

    if [ "$cleansubuid" = true ] ; then
      sed -i "/^${userid}:/d" /etc/subuid
    fi

    if [ "$cleansubgid" = true ] ; then
      sed -i "/^${userid}:/d" /etc/subgid
    fi
  args:
    executable: /bin/bash
    creates: "{{ rootlessdocker_install_dir }}/config/install_success"

- name: Clean up rootless docker script
  file:
    path: /tmp/rootless
    state: absent

```

I end up having to add three additional tasks just to avoid curl command in the shell command i.e. part of a single line:
```
    curl -fsSL https://get.docker.com/rootless | \
      FORCE_ROOTLESS_INSTALL=1 DOCKER_BIN="{{ rootlessdocker_install_dir }}/bin" sh
```

I have not written ansible scripts before, so I'm not very experienced. Perhaps someone could suggest a better implementation, but it just does not seem to be good code to make something so verbose when it can be a single line. If there was an ansible module to manipulate subuid/subgid, then using just ansible modules would be simpler. But I did not find such a module.

avolkov1:
This PR is ready to be reviewed again.

I refactored the in-lined shell script into a templated setup script that ansible calls. I ran the playbook and tested it. It worked correctly on Ubuntu DGX based system.

Do you squash when you merge PRs? Otherwise I can squash commits and rebase to whatever is latest on the master branch for your convenience.

avolkov1:
I'm going to squash and rebase to latest master. I'll do a force push so if you have a local branch checked out of my rootless-docker you will need to delete and re-clone or re-checkout from remote.

avolkov1:
Hold off on merging this. I can't get rootless docker to work on RHEL7. I'm investigating.

avolkov1:
The PR is ready. I updated bits for RHEL. Rootless docker installs fine on RHEL7 and RHEL8, although it only works on properly on RHEL8. The RHEL7 kernel is too old to support rootless docker under Slurm cgroups.

Some functionality works under RHEL7; you can pull, push, and build docker images, but not run them. This could still be useful as one can then convert and run these containers via enroot.

On Ubuntu and RHEL8, within a Slurm reservation, rootless docker works without any issues.

elgalu:
It may also make sense to configure a [docker pull-through cache](https://docs.docker.com/registry/recipes/mirror/#run-a-registry-as-a-pull-through-cache) as with pyxis+enroot there was cache out-of-the-box which gets lost if fully switching to this solution.

avolkov1:
I'm going to rebase and squash for latest changes in master.

avolkov1:
I rebased/squashed `rootless-docker` branch against latest master. Since I `push --force` please delete your local checked out branch of `rootless-docker` (if you have one) and re-fetch/re-checkout from remote.

This PR is ready to be reviewed. Please update status at your convenience.

roles/docker-rootless/templates/rootless_docker_setup.sh.j2
```diff
@@ -0,0 +1,68 @@
...
+
+if [ "$cleansubuid" = true ] ; then
+  sed -i "/^${userid}:/d" /etc/subuid
+fi
+
+if [ "$cleansubgid" = true ] ; then
+  sed -i "/^${userid}:/d" /etc/subgid
+fi
+
+if [ $? -eq 0 ] ; then
```
dholt, line 60: Does this test the exit code of the rootless docker install as intended, or of the clean subuid/gid sed commands?

playbooks/container/docker-rootless.yml
```diff
@@ -0,0 +1,5 @@
+---
```
dholt, line 1: Since this role depends on docker being installed, suggest you add the dependency here, i.e.

```
---
- include: ../container/docker.yml
```

dholt:
In addition to the two comments above, I'm running into this:

```
(env) ubuntu@polite-tapir:~$ /sw/software/rootless-docker/bin/start_rootless_docker.sh
/sw/software/rootless-docker/bin/start_rootless_docker.sh: line 56: dockerd-rootless.sh: command not found
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/json: dial unix /var/run/docker.sock: connect: permission denied
```

Does my user need to be in the `docker` group? That seems like it would defeat the purpose of having rootless docker.

roles/docker-rootless/templates/rootless_docker_setup.sh.j2
avolkov1: I need to fix this. Good catch.

playbooks/container/docker-rootless.yml
avolkov1: No, there is not dependency on docker being installed. Rootless docker installs its own docker.
```
$ which docker
/sw/software/rootless-docker/bin/docker
```

avolkov1:
> In addition to the two comments above, I'm running into this:
> 
> ```
> (env) ubuntu@polite-tapir:~$ /sw/software/rootless-docker/bin/start_rootless_docker.sh
> /sw/software/rootless-docker/bin/start_rootless_docker.sh: line 56: dockerd-rootless.sh: command not found
> Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.40/containers/json: dial unix /var/run/docker.sock: connect: permission denied
> ```
> 
> Does my user need to be in the `docker` group? That seems like it would defeat the purpose of having rootless docker.

No, something it not installed properly then. The `dockerd-rootless.sh` should be in:
```
$ which dockerd-rootless.sh 
/sw/software/rootless-docker/bin/dockerd-rootless.sh
```
Do you have `/sw/software/rootless-docker/bin` in the path?

I add a module that sets the PATH and a few other environment variables.
```
$ ls -1 /sw/software/rootless-docker/modulefiles
rootless-docker.lua
```
So you should be able to set `module use ...` and then do `module load rootless-docker`. Otherwise explicitly set the variables. Look at the module file:
```
$ less /sw/software/rootless-docker/modulefiles/rootless-docker.lua
prepend_path("PATH", "/sw/software/rootless-docker/bin")

local userid = capture("id -u")
local userid = string.gsub(userid, '%s+', '')
local xdg_runtime_dir = "/var/tmp/xdg_runtime_dir_" .. userid
local docker_dataroot = "/var/tmp/docker-container-storage-" .. userid

pushenv("XDG_RUNTIME_DIR", xdg_runtime_dir)
setenv("DOCKER_HOST", "unix://"..xdg_runtime_dir.."/docker.sock")
setenv("DOCKER_DATAROOT", docker_dataroot)
```

roles/docker-rootless/templates/rootless_docker_setup.sh.j2
avolkov1: I fixed this logic in commit b492d90.

dholt:
Ah, I didn't use modules, just ran the scripts. Maybe we should make the lmod playbook a dependency and state that you need to use the module or set paths manually?

I got a failure trying to run a container with GPUs, is the nvidia-container-runtime a dependency too?

avolkov1:
> Ah, I didn't use modules, just ran the scripts. Maybe we should make the lmod playbook a dependency and state that you need to use the module or set paths manually?
> 
> I got a failure trying to run a container with GPUs, is the nvidia-container-runtime a dependency too?

Yes, there is a dependency on nvidia-container-runtime. In particular there is a file `/sw/software/rootless-docker/bin/nvidia-container-runtime-hook` which uses `/usr/bin/nvidia-container-runtime-hook`.

When you run rootless docker for GPUs add the `--gpus` flag. Something like:
```
docker run --gpus all -it --rm nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
```
or vary the `--gpus` to pass only the desired GPUs.

Do you have the syntax for adding dependency for lmod playbook and nvidia-container-runtime. In the README it's documented to use the module.
https://github.com/avolkov1/deepops/blob/single-node-slurm/docs/container/docker-rootless.md#using-rootless-docker-in-slurm

avolkov1:
Should I add the lmod dependency as this?
```
- include: slurm-cluster/lmod.yml
  when: slurm_install_lmod
```

dholt:
That seems fine as long as `slurm_install_lmod` defaults to true, otherwise you may want to set it to true in the rootless-docker playbook.

Including nvidia-container-runtime is similar. You could include `playbooks/container/nvidia-docker.yml` while setting `docker_install` to false, or just include the nvidia docker role.

avolkov1:
> That seems fine as long as `slurm_install_lmod` defaults to true, otherwise you may want to set it to true in the rootless-docker playbook.
> 
> Including nvidia-container-runtime is similar. You could include `playbooks/container/nvidia-docker.yml` while setting `docker_install` to false, or just include the nvidia docker role.

Ok, I think I made these changes in commit f09ba47. The playbook runs and install rootless docker with lmod and nvidia container runtime.

Note, that the nvidia docker role seems to actually mess up the container runtime hook setup on RHEL on DGXs. But this is not an issue due to rootless docker. I have to deal with this when deploying Slurm on DGXs running RHEL regardless of rootless docker.
With DGXs on RHEL after running deepops slurm playbook (and now this playbook) I have to run
```
yum group install -y 'NVIDIA Container Runtime'
or
dnf group install -y 'NVIDIA Container Runtime'
```
To fix issues with nvidia container runtime. Again, this bug/issue is unrelated to rootless docker. Someone should fix it though.

dholt:
Unfortunately by including the nvidia-docker role, but not installing docker, the nvidia-docker role fails when trying to restart the docker daemon:

```RUNNING HANDLER [nvidia.nvidia_docker : reload docker] **********************************************************************************************************
Monday 22 March 2021  10:36:55 -0700 (0:00:00.170)       0:00:46.896 **********
fatal: [polite-tapir]: FAILED! => changed=false
  msg: 'Could not find the requested service docker: host'
```

I'll open a PR for the nvidia-docker role to add a var `skip_docker_reload` that lets you skip restarting the daemon. In the `docker-rootless.yml` playbook you'll want to pass `skip_docker_reload: true`

dholt:
I'm still seeing some things that look like errors and fail to run a container:

```
+ exec rootlesskit --net=vpnkit --mtu=1500 --slirp4netns-sandbox=auto --slirp4netns-seccomp=auto --disable-host-loopback --port-driver=builtin --copy-up=/etc --c
opy-up=/run --propagation=rslave /sw/software/rootless-docker/bin/dockerd-rootless.sh --experimental --data-root=/var/tmp/docker-container-storage-1000 --storage
-driver overlay2
Cannot connect to the Docker daemon at unix:///var/tmp/xdg_runtime_dir_1000/docker.sock. Is the docker daemon running?
Cannot connect to the Docker daemon at unix:///var/tmp/xdg_runtime_dir_1000/docker.sock. Is the docker daemon running?
...
+ exec dockerd --experimental --data-root=/var/tmp/docker-container-storage-1000 --storage-driver overlay2                                                       Cannot connect to the Docker daemon at unix:///var/tmp/xdg_runtime_dir_1000/docker.sock. Is the docker daemon running?
...
INFO[2021-03-22T18:10:08.702972015Z] skip loading plugin "io.containerd.snapshotter.v1.aufs"...  error="aufs is not supported (modprobe aufs failed: exit status
1 \"modprobe: ERROR: could not insert 'aufs': Operation not permitted\\n\"): skip plugin" type=io.containerd.snapshotter.v1
...
WARN[2021-03-22T18:10:08.711870834Z] failed to load plugin io.containerd.internal.v1.opt  error="mkdir /opt/containerd: permission denied"
...
WARN[2021-03-22T18:10:08.775565878Z] Running modprobe bridge br_netfilter failed with message: modprobe: ERROR: could not insert 'bridge': Operation not permitte
d
modprobe: ERROR: could not insert 'br_netfilter': Operation not permitted
insmod /lib/modules/5.4.0-67-generic/kernel/net/llc/llc.ko
insmod /lib/modules/5.4.0-67-generic/kernel/net/llc/llc.ko
, error: exit status 1
```

```
ubuntu@polite-tapir:~$ docker run --gpus all -it --rm nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
Unable to find image 'nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04' locally
11.0-base-ubuntu18.04: Pulling from nvidia/cuda
171857c49d0f: Pull complete
419640447d26: Pull complete
61e52f862619: Pull complete
2a93278deddf: Pull complete
c9f080049843: Pull complete
8189556b2329: Pull complete
Digest: sha256:4182cf1fbd8a16a17dba572c0f4d1cb655eb9942a761fab7eeec2bcec077c880
Status: Downloaded newer image for nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
time="2021-03-22T18:12:44.954922450Z" level=info msg="starting signal loop" namespace=moby path=/var/tmp/xdg_runtime_dir_1000/docker/containerd/daemon/io.containerd.runtime.v2.task/moby/83e2e1918907b763d1f5bc2198e57153ec57fce798dd735f140805ec2e4f6063 pid=9618
                                                                                                  INFO[2021-03-22T18:12:44.995102309Z] shim disconnected                             id=83e2e1918907b763d1f5bc2198e57153ec57fce798dd735f140805ec2e4f6063
                                                                                       ERRO[2021-03-22T18:12:44.996218624Z] stream copy error: reading from a closed fifo
         ERRO[2021-03-22T18:12:45.152776118Z] 83e2e1918907b763d1f5bc2198e57153ec57fce798dd735f140805ec2e4f6063 cleanup: failed to delete container from containerd: no such container
ERRO[2021-03-22T18:12:45.157964867Z] Handler for POST /v1.41/containers/83e2e1918907b763d1f5bc2198e57153ec57fce798dd735f140805ec2e4f6063/start returned error: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: driver error: failed to process request: unknown
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: driver error: failed to process request: unknown.
```

avolkov1:
I'm not sure what is going on. It looks like you can pull a container with rootless docker but not run it?
Do the following work:
```
docker pull nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
docker images
```

avolkov1:
Can you share the contents of `/sw/software/rootless-docker/bin/nvidia-container-runtime-hook`.
Should be:
```
$ cat /sw/software/rootless-docker/bin/nvidia-container-runtime-hook
#!/bin/sh

/usr/bin/nvidia-container-runtime-hook -config="/sw/software/rootless-docker/config/nvidia-container-runtime/config.toml" "$@"
```

Then share the contents of `/sw/software/rootless-docker/config/nvidia-container-runtime/config.toml`
On Ubuntu it should be something like this:
```
disable-require = false
#swarm-resource = "DOCKER_RESOURCE_GPU"
#accept-nvidia-visible-devices-envvar-when-unprivileged = true
#accept-nvidia-visible-devices-as-volume-mounts = false

[nvidia-container-cli]
no-cgroups = true
#root = "/run/nvidia/driver"
#path = "/usr/bin/nvidia-container-cli"
environment = []
#debug = "/var/log/nvidia-container-toolkit.log"
#ldcache = "/etc/ld.so.cache"
load-kmods = true
#user = "root:video"
ldconfig = "@/sbin/ldconfig.real"

[nvidia-container-runtime]
#debug = "/var/log/nvidia-container-runtime.log"
```

On RHEL/CentOS it would look like this:
```
disable-require = false
#swarm-resource = "DOCKER_RESOURCE_GPU"
#accept-nvidia-visible-devices-envvar-when-unprivileged = true
#accept-nvidia-visible-devices-as-volume-mounts = false

[nvidia-container-cli]
no-cgroups = true
#root = "/run/nvidia/driver"
#path = "/usr/bin/nvidia-container-cli"
environment = []
#debug = "/var/log/nvidia-container-toolkit.log"
#ldcache = "/etc/ld.so.cache"
load-kmods = true
#user = "root:video"
ldconfig = "@/sbin/ldconfig"

[nvidia-container-runtime]
#debug = "/var/log/nvidia-container-runtime.log"
```

avolkov1:
The other error I see is that `br_netfilter` is not present. This should not happen:
```
WARN[2021-03-22T18:10:08.775565878Z] Running modprobe bridge br_netfilter failed with message: modprobe: ERROR: could not insert 'bridge': Operation not permitte
d
modprobe: ERROR: could not insert 'br_netfilter': Operation not permitted
```
The `br_netfilter` should already be loaded. I don't know when or how that's installed, but on all the systems I have access to (NSL-A, Linux VDI, DVT08)  it's already loaded.
```
$ lsmod | grep -i br_netfilter
br_netfilter           22256  0 
bridge                151336  1 br_netfilter
```

dholt:
> I'm not sure what is going on. It looks like you can pull a container with rootless docker but not run it?
> Do the following work:
> 
> ```
> docker pull nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
> docker images
> ```

Yes, this works

dholt:
> Can you share the contents of `/sw/software/rootless-docker/bin/nvidia-container-runtime-hook`.
> Should be:
> 
> ```
> $ cat /sw/software/rootless-docker/bin/nvidia-container-runtime-hook
> #!/bin/sh
> 
> /usr/bin/nvidia-container-runtime-hook -config="/sw/software/rootless-docker/config/nvidia-container-runtime/config.toml" "$@"
> ```
> 
> Then share the contents of `/sw/software/rootless-docker/config/nvidia-container-runtime/config.toml`
> On Ubuntu it should be something like this:
> 
> ```
> disable-require = false
> #swarm-resource = "DOCKER_RESOURCE_GPU"
> #accept-nvidia-visible-devices-envvar-when-unprivileged = true
> #accept-nvidia-visible-devices-as-volume-mounts = false
> 
> [nvidia-container-cli]
> no-cgroups = true
> #root = "/run/nvidia/driver"
> #path = "/usr/bin/nvidia-container-cli"
> environment = []
> #debug = "/var/log/nvidia-container-toolkit.log"
> #ldcache = "/etc/ld.so.cache"
> load-kmods = true
> #user = "root:video"
> ldconfig = "@/sbin/ldconfig.real"
> 
> [nvidia-container-runtime]
> #debug = "/var/log/nvidia-container-runtime.log"
> ```
> 
> On RHEL/CentOS it would look like this:
> 
> ```
> disable-require = false
> #swarm-resource = "DOCKER_RESOURCE_GPU"
> #accept-nvidia-visible-devices-envvar-when-unprivileged = true
> #accept-nvidia-visible-devices-as-volume-mounts = false
> 
> [nvidia-container-cli]
> no-cgroups = true
> #root = "/run/nvidia/driver"
> #path = "/usr/bin/nvidia-container-cli"
> environment = []
> #debug = "/var/log/nvidia-container-toolkit.log"
> #ldcache = "/etc/ld.so.cache"
> load-kmods = true
> #user = "root:video"
> ldconfig = "@/sbin/ldconfig"
> 
> [nvidia-container-runtime]
> #debug = "/var/log/nvidia-container-runtime.log"
> ```

```
ubuntu@polite-tapir:~$ cat /sw/software/rootless-docker/bin/nvidia-container-runtime-hook
#!/bin/sh

/usr/bin/nvidia-container-runtime-hook -config="/sw/software/rootless-docker/config/nvidia-container-runtime/config.toml" "$@"
```

```
ubuntu@polite-tapir:~$ cat /sw/software/rootless-docker/config/nvidia-container-runtime/config.toml
disable-require = false
#swarm-resource = "DOCKER_RESOURCE_GPU"
#accept-nvidia-visible-devices-envvar-when-unprivileged = true
#accept-nvidia-visible-devices-as-volume-mounts = false

[nvidia-container-cli]
no-cgroups = true
#root = "/run/nvidia/driver"
#path = "/usr/bin/nvidia-container-cli"
environment = []
#debug = "/var/log/nvidia-container-toolkit.log"
#ldcache = "/etc/ld.so.cache"
load-kmods = true
#user = "root:video"
ldconfig = "@/sbin/ldconfig.real"

[nvidia-container-runtime]
#debug = "/var/log/nvidia-container-runtime.log"
```

dholt:
> The other error I see is that `br_netfilter` is not present. This should not happen:
> 
> ```
> WARN[2021-03-22T18:10:08.775565878Z] Running modprobe bridge br_netfilter failed with message: modprobe: ERROR: could not insert 'bridge': Operation not permitte
> d
> modprobe: ERROR: could not insert 'br_netfilter': Operation not permitted
> ```
> 
> The `br_netfilter` should already be loaded. I don't know when or how that's installed, but on all the systems I have access to (NSL-A, Linux VDI, DVT08) it's already loaded.
> 
> ```
> $ lsmod | grep -i br_netfilter
> br_netfilter           22256  0 
> bridge                151336  1 br_netfilter
> ```

It's not loaded:

```
ubuntu@polite-tapir:~$ lsmod | grep -i br_netfilter
ubuntu@polite-tapir:~$
```

This is a vanilla Ubuntu 20.04 install via MAAS (roughly a cloud image)

avolkov1:
The `br_netfilter` is probably nice to have for various docker port mapping options i.e. `-p port:port` but the underlying issue looks like the nvidia-container-runtime is not happy with something.

Regarding this error:
```
ERRO[2021-03-22T18:12:45.157964867Z] Handler for POST /v1.41/containers/83e2e1918907b763d1f5bc2198e57153ec57fce798dd735f140805ec2e4f6063/start returned error: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: driver error: failed to process request: unknown
docker: Error response from daemon: OCI runtime create failed: container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: Running hook #0:: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: initialization error: driver error: failed to process request: unknown.
```

Maybe something is not setup right with nvidia driver?

Can you try:
https://github.com/NVIDIA/nvidia-docker/issues/726#issuecomment-386717947
```
nvidia-container-cli -k -d /dev/tty info
```
The end of the output should be something like:
```
I0322 20:09:24.513946 3512589 nvc.c:427] shutting down library context
I0322 20:09:24.515229 3512591 driver.c:156] terminating driver service
I0322 20:09:24.515712 3512589 driver.c:196] driver service terminated successfully
```

avolkov1:
Regarding `br_netfilter` looks like this is the package that might be required on Ubuntu:
https://packages.ubuntu.com/focal/linux-modules-5.4.0-26-generic

I have it installed:
```
linux-modules-5.4.0-67-generic                  install
```
I'll have to search for CentOS/RHEL version of this package.

dholt:
I was just testing sans-GPU 🤦  does this require a GPU or should it work CPU-only?

avolkov1:
It should work CPU only as well. GPUs are not required for rootless docker, but if GPUs are used then nvidia-container-runtime is required.

dholt:
Using the playbook I suggested above, it does seem to work with a GPU, even though I get some messages like:

```
Cannot connect to the Docker daemon at unix:///var/tmp/xdg_runtime_dir_1000/docker.sock. Is the docker daemon running?
```

Without a GPU it also seems to work, but I get all of the errors listed previously above

supertetelman:
Before we merge this through, are there any rootless-docker specific tests we should add to Jenkins? I don't see anything checked in here, but would love to have some confidence that this continues to work with the few different deployment options we now have.

dholt:
> Before we merge this through, are there any rootless-docker specific tests we should add to Jenkins? I don't see anything checked in here, but would love to have some confidence that this continues to work with the few different deployment options we now have.

While I agree, this stuff isn't run by default anywhere so it shouldn't break anything else

avolkov1:
That message `Cannot connect to the Docker daemon at unix:///var/tmp/xdg_runtime_dir_1000/docker.sock. Is the docker daemon running?` is temporary until the docker daemon comes up. In the `start_rootless_docker.sh` script there's a snippet:
```
    # Insure that docker daemon started.
    docker ps >/dev/null
    while [ $? -ne 0 ]; do
        docker ps >/dev/null
    done
```

Eventually it should come up, or if it doesn't then something broke. Anyways, yes there are a lot of warning and message with rootless docker. I added a `--quiet` flag to hide those (first make sure it's actually working before using `--quiet` flag).
```
$ start_rootless_docker.sh --quiet
```
So after the rootless docker daemon starts if running without GPUs do this:
```
docker run -it --rm nvcr.io/nvidia/cuda:11.0-base-ubuntu18.04
```
If above works, then the installation works.

dholt:
Sounds good, I think it's working for me then. My only remaining issue is that you should either document the dependencies (nvidia driver, nvidia-container-runtime for GPU, Lmod for CPU/GPU) or use a playbook that includes those playbooks vs roles, i.e.

```
---
- include: ../nvidia-software/nvidia-driver.yml
- include: nvidia-docker.yml
- include: ../slurm-cluster/lmod.yml

- hosts: all
  become: yes
  tasks:
    - name: install rootless docker
      include_role:
        name: docker-rootless
```

avolkov1:
> ```
> ---
> - include: ../nvidia-software/nvidia-driver.yml
> - include: nvidia-docker.yml
> - include: ../slurm-cluster/lmod.yml
> 
> - hosts: all
>   become: yes
>   tasks:
>     - name: install rootless docker
>       include_role:
>         name: docker-rootless
> ```

Ok, if installing all that other stuff, which again you don't need `nvidia-driver` or `nvidia-docker` if not using GPUs, then the `hosts: all` needs to be filtered for compute nodes. I don't think you want to run these GPU specific roles on login/management nodes.

dholt (approved):

PR #829, opened 2021-01-15
Update helm to 3.4.1

supertetelman:


supertetelman:
This now detects if a system is running an older version of Helm and automatically upgrades it (while backing up the old binary).

michael-balint (approved):

PR #830, opened 2021-01-15
Install nfs on k8s-cluster nodes, not all during k8s install

supertetelman:
Restrict the installation of nfs utils to nodes in the k8s-cluster group during installation of the nfs-client-provisioner.

Some recent changes to the nfs role caused this to be problematic in our testing.

### DIFF
--- a/playbooks/k8s-cluster/nfs-client-provisioner.yml
+++ b/playbooks/k8s-cluster/nfs-client-provisioner.yml
@@ -27,7 +27,7 @@
     - nfs_is_server: yes
     when: k8s_deploy_nfs_server
 
-- hosts: "all"
+- hosts: "k8s-cluster"
   become: yes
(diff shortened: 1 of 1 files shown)


playbooks/k8s-cluster/nfs-client-provisioner.yml
```diff
@@ -27,7 +27,7 @@
     - nfs_is_server: yes
     when: k8s_deploy_nfs_server
 
-- hosts: "all"
+- hosts: "k8s-cluster"
   become: yes
   roles:
     - nfs
```
dholt, line 33: Is this just the client parts? If so, you need to set that var, i.e.:

```
roles:
    - { role: nfs, nfs_is_client: yes }
```

dholt (approved):

PR #846, opened 2021-02-02
Bump GPU Operator to 1.5.1 and add nightly testing

supertetelman:
* Add a test to the nightly tests of CentOS/Ubuntu to deploy K8s with GPU Operator and run a GPU job (keep existing tests as is)
* Bump GPU Operator version from 1.1.7 to 1.5.1


supertetelman:
This is now passing in Jenkins.

Although, for whatever reason the GPU Feature Discovery is not labeling nodes properly, this seems like a bug in the deployment and not the testing, so we should follow up with a different PR/issue.

dholt (approved):

Issue #833, opened 2021-01-20
NFS role is failing due to missing firewalld package/services on RHEL and CentOS


supertetelman:
It looks like we have some missing dependencies and/or edge cases when installing the NFS role into a k8s-cluster. The nightly tests have started failing on the CentOS install and I am currently working through a mising package issue with a customer running RHEL 7.

I am seeing he below error in the nightly tests when it is installing K8s onto CentOS

```sh
TASK [nfs : configure firewall to allow NFS server] ****************************
failed: [virtual-mgmt01-0] (item=mountd) => changed=false 
  ansible_loop_var: item
  item: mountd
  msg: firewall is not currently running, unable to perform immediate actions without a running firewall daemon
failed: [virtual-mgmt01-0] (item=rpc-bind) => changed=false 
  ansible_loop_var: item
  item: rpc-bind
  msg: firewall is not currently running, unable to perform immediate actions without a running firewall daemon
failed: [virtual-mgmt01-0] (item=nfs) => changed=false 
  ansible_loop_var: item
  item: nfs
  msg: firewall is not currently running, unable to perform immediate actions without a running firewall daemon
```

I also saw a similar issue working in a RHEL environment on a customer deployment:

```sh
TASK [nfs : configure firewall to allow NFS server] ******************************************************************
failed: [cpu01] (item=mountd) => changed=false
  ansible_loop_var: item
  item: mountd
  msg: 'Python Module not found: firewalld and its python module are required for this module,                         version 0.2.11 or newer required (0.3.9 or newer for offline operations)'
failed: [cpu01] (item=rpc-bind) => changed=false
  ansible_loop_var: item
  item: rpc-bind
  msg: 'Python Module not found: firewalld and its python module are required for this module,                         version 0.2.11 or newer required (0.3.9 or newer for offline operations)'
failed: [cpu01] (item=nfs) => changed=false
  ansible_loop_var: item
  item: nfs
  msg: 'Python Module not found: firewalld and its python module are required for this module,                         version 0.2.11 or newer required (0.3.9 or newer for offline operations)'
```

supertetelman:
Fixed

PR #849, opened 2021-02-02
Modify firewall rules only when firewalld is installed

dholt:
Fix for #833

**Test Plan:**
On CentOS/RHEL systems without the `firewalld` package, the task to configure firewall rules should be skipped.


supertetelman:
I'm still seeing the same error in the Jenkins test for this branch.

supertetelman (approved):
LGTM

PR #851, opened 2021-02-03
Remove unused Apt k8s service

dholt:
Remove this service so we can also remove the stale container (https://hub.docker.com/r/deepops/apt-online)


ajdecon (approved):
A search of the repo doesn't turn up any references to `apt.yml` so I'm pretty sure nothing uses this. :) 

LGTM

PR #860, opened 2021-02-05
Simple role for installing docker-ce.

mantolini-nv:
This is one of multiple PRs upcoming for NVIDIA effort to add support for Scenario Runner to DeepOps.

We have built a POC to validate this internally.

The full list of planned PRs:
1. docker-ce role
2. rclone role
3. scenario-runner role and playbooks

Signed-off-by: Martin Antolini <mantolini@nvidia.com>


ajdecon:
@mantolini-nv : Apologies that I didn't remember this when we chatted, but for Docker installs in other contexts we are embedding the role from Kubespray. 

E.g., see the docker playbook: https://github.com/NVIDIA/deepops/blob/master/playbooks/container/docker.yml

Can you try using this playbook in your framework and see if it works for you, instead of adding a new role?

mantolini-nv:
@ajdecon I hd to write a new docker-ce role because I kept having problems with the main one. I can't remember the issue but I think it was either forcing a version or configuration that I didn't want. But my playbook changed a lot since that original decision, so I'll try it again.

mantolini-nv:
@ajdecon the playbook https://github.com/NVIDIA/deepops/blob/master/playbooks/container/docker.yml has a lot of kubespray stuff in it and it's failing for me since I don't want k8s. Is there a generic docker role that I'm missing? thanks!

ajdecon:
@mantolini-nv : Can you share the failures you're seeing? While we are piggybacking on the Kubespray docker role, we definitely use it outside a Kubernetes context, as in our [Grafana playbook](https://github.com/NVIDIA/deepops/blob/master/playbooks/slurm-cluster/grafana.yml) which is used on generic Slurm clusters.

mantolini-nv:
@ajdecon ah interesting, here is the error: 

mantolini@mantolini-drivesim:~/git/constellationbatchpoc$ ansible-playbook -i config/inventory.sae -l slurm-cluster playbooks/test.yml
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
ERROR! vars file ../../submodules/kubespray/roles/kubespray-defaults/defaults/main.yaml was not found
Could not find file on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
mantolini@mantolini-drivesim:~/git/constellationbatchpoc$ 

and this is how I'm calling it from the playbook:

- include: container/docker.yml
  tags: docker
  vars:
    hostlist: "{{ slurm-master[0] }}"

thanks!

ajdecon (changes requested):
Requesting changes on this to block merge as I think we figured out offline how to make the existing solution work for this use case.

Please ping me if there are still challenges and we should revisit this!

ajdecon:
Closing as obsolete, but let's discuss offline if this is still needed!

PR #861, opened 2021-02-05
Add additional configurations options for GPU Operator 

supertetelman:
This PR should not impact existing GPU Operator installs.

* Bumps version of GPU Operator,GFD, Device Plugin, and GPU Operator
* Fixes some GPU Operator install steps for CentOS
* It adds the proper support to pass through the existing migStrategy variable to the GFD and Device Plugin.
* It adds support for the creation of a secret key and configuration to use that secret key in the operator. This is needed for vGPU and for beta drivers.

*Note* This PR does not add full support for an automated vGPU install. This will require manual steps to build and push a driver container as per https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/getting-started.html#considerations-to-install-gpu-operator-with-nvidia-vgpu-driver. As of now, I do not think this is something that should be automated in DeepOps.

**Testing**

There haven't been many significant changes in this PR. In general the below test cases should validate all functionality changes. For MIG tests, verifying that the correct labels are applied to the node and the correct resource types are available will suffice. Otherwhise, running a basic GPU job in each case is validation enough. In order to test vGPU, you will need to create a private registry hosting the special vGPU driver container.

1. Test standard GPU Operator install on Ubuntu. Covered by regression tests.
2. Test standard GPU Operator install on CentOS 8.0. Currently not covered by regressions tests and untested.
3. Test on a MIG-enabled GPU. We should deploy this wih the `k8s_gpu_mig_strategy` flag set to a few different values  (`single` and `mixed`) and make sure it is passed all the way through to GFD and device plugin.
4. Test on vGPU with custom built drivers and MIG.
5. Test on vGPU with custom built drivers and no MIG.


supertetelman:
If you are doing an upgrade or re-install it is currently necessary to run `rmmod nvidia_modeset nvidia_uvm nvidia` on all GPU nodes. We should add this as a init task.

michael-balint:
@supertetelman can you work with @gianluca10 to test this PR out?

supertetelman:
Just merged master into this so that we can actually see tests passing.

supertetelman:
As of nightly run 779, this is now passing all tests with the exception of Operator+CentOS (https://github.com/NVIDIA/deepops/issues/868).

ajdecon (approved):
Overall LGTM. One comment with a nit, but that shouldn't be a blocker.

PR #864, opened 2021-02-08
Rclone role for deepops

mantolini-nv:
This is one of multiple PRs upcoming for NVIDIA effort to add support for Scenario Runner to DeepOps.

We have built a POC to validate this internally. I have tested this role individually.

The full list of planned PRs:

    docker-ce role - https://github.com/NVIDIA/deepops/pull/860 
    rclone role - this one
    scenario-runner role and playbooks - coming soon

Signed-off-by: Martin Antolini mantolini@nvidia.com

### DIFF
--- a/roles/rclone/README.md
+++ b/roles/rclone/README.md
@@ -1,0 +1,3 @@
+Rclone is a command line program to manage files on cloud storage. It supports over 40 cloud storage products including S3.
+
+For more information or rclone see: https://rclone.org
--- a/roles/rclone/tasks/main.yml
+++ b/roles/rclone/tasks/main.yml
@@ -1,0 +1,30 @@
+---
+- name: Uninstall current version of rclone
+  become: yes
+  shell: rm /usr/bin/rclone /usr/local/share/man/man1/rclone.1 '{{ rclone_config_location }}/rclone.conf'
+  ignore_errors: yes
+
+- name: Install rclone
+  shell: curl https://rclone.org/install.sh | sudo bash
+
+- name: Create rclone config directory
+  file:
+    path: '{{ rclone_config_location }}'
+    state: directory
+    mode: 0777
+  become: yes
+
+- name: Set up rclone config
+  template:
+    src: rclone.conf.j2
+    dest: '{{ rclone_config_location }}/rclone.conf'
+  become: yes
+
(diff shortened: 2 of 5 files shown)


roles/rclone/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Uninstall current version of rclone
+  become: yes
+  shell: rm /usr/bin/rclone /usr/local/share/man/man1/rclone.1 '{{ rclone_config_location }}/rclone.conf'
+  ignore_errors: yes
+
+- name: Install rclone
+  shell: curl https://rclone.org/install.sh | sudo bash
```
ajdecon, line 8: It looks like rclone is provided in the Ubuntu repositories and in the EPEL repository for RHEL systems.

Instead of using the install script with `sudo bash` (which may be unstable for future installs), can you install from the repository versions instead?

roles/rclone/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Uninstall current version of rclone
+  become: yes
+  shell: rm /usr/bin/rclone /usr/local/share/man/man1/rclone.1 '{{ rclone_config_location }}/rclone.conf'
```
ajdecon, line 4: If you swap to the repository versions (previous comment) this will be unnecessary

roles/rclone/README.md
```diff
@@ -0,0 +1,3 @@
+Rclone is a command line program to manage files on cloud storage. It supports over 40 cloud storage products including S3.
```
ajdecon, line 1: Can you make the following additions to this README?

- Add a note that explicitly says access keys are required to use this role
- Add a usage example

roles/rclone/templates/rclone.conf.j2
```diff
@@ -0,0 +1,7 @@
+{% for config in rclone_configs %}
+[{{ config.name }}]
+{% for key, value in config.properties.items() %}
+{{ key }} = {{ value }}
+{% endfor %}
+
+{% endfor %}
```
ajdecon, line 7: Please add newline at the end of this file

ajdecon (changes requested):

ajdecon:
Closing as obsolete, but let's discuss offline if this is still needed!

PR #866, opened 2021-02-08
Fix EPEL install on RHEL systems

ajdecon:
## Summary

RHEL systems may not include `epel-release` in the default repos, so installing this as a bare package name may fail.

The instructions on the [EPEL site](https://fedoraproject.org/wiki/EPEL) suggest installing direct from their site, so we do that here. We also pull the install location into a variable so we can override it when needed.

## Test plan

Turned up Kubernetes cluster on CentOS using these changes, validated that EPEL was installed correctly. This validates that the install mechanism from the EPEL URL works.

Still need to test directly on RHEL, waiting on a test host.

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -42,7 +42,8 @@ as_user(){
 case "$ID" in
     rhel*|centos*)
         # Enable EPEL (required for Pip)
-        EPEL_URL="https://dl.fedoraproject.org/pub/epel/epel-release-latest-${VERSION_ID}.noarch.rpm"
+        EPEL_VERSION="$(echo ${VERSION_ID} | sed  's/^[^0-9]*//;s/[^0-9].*$//')"
+        EPEL_URL="https://dl.fedoraproject.org/pub/epel/epel-release-latest-${EPEL_VERSION}.noarch.rpm"
         as_sudo "yum -y install ${EPEL_URL}"
 
         # Install pip
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -42,7 +42,8 @@ as_user(){
 case "$ID" in
     rhel*|centos*)
         # Enable EPEL (required for Pip)
-        as_sudo 'yum -y install epel-release'
+        EPEL_URL="https://dl.fedoraproject.org/pub/epel/epel-release-latest-${VERSION_ID}.noarch.rpm"
+        as_sudo "yum -y install ${EPEL_URL}"
 
         # Install pip
         if ! which ${PIP} >/dev/null 2>&1; then
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -87,6 +87,8 @@ deepops_set_hostname: true
 #    path: /export/shared
 #    options: async,vers=3
 
+# Extra Packages for Enterprise Linux (RHEL/CentOS)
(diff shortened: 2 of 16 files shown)


avolkov1:
I'm testing on RHEL8 right now.

I tried to run this, but I can't get ansible to proceed, it errors out here:
```
ASK [pick coredns cluster IP or default resolver] ******************************************************
fatal: [dvt08]: FAILED! => 
  msg: 'An unhandled exception occurred while templating ''{{ kube_service_addresses|ipaddr(''net'')|ipaddr(3)|ipaddr(''address'') }}''. Error was a <class ''ansible.errors.AnsibleFilterError''>, original message: The ipaddr filter requires python''s netaddr be installed on the ansible controller'
```

I had to manually run:
```
sudo pip3 install netaddr
```

The problem is that when I run `./scripts/setup.sh` it doesn't install everything needed such as "sshpass". Is it possible to add epel repo in this setup script? Once I manually installed the epel repo and ran `setup.sh` then all prerequisites installed and everything proceeded correctly.

ajdecon:
That's weird, as we definitely install `netaddr` in the setup script. (In several places, depending on the distro and Python config! But for example, [here](https://github.com/NVIDIA/deepops/blob/master/scripts/setup.sh#L70))

We also attempt to install EPEL if the controller is a CentOS or RHEL machine, [here](https://github.com/NVIDIA/deepops/blob/master/scripts/setup.sh#L45)

However, that script needs to be fixed in the same way this PR is doing. 😄 I'll update it to use the Fedora URL in the same way we do here.

ajdecon:
@avolkov1 : Latest commit should fix the setup script

scripts/setup.sh
```diff
@@ -42,7 +42,8 @@ as_user(){
 case "$ID" in
     rhel*|centos*)
         # Enable EPEL (required for Pip)
-        as_sudo 'yum -y install epel-release'
+        EPEL_URL="https://dl.fedoraproject.org/pub/epel/epel-release-latest-${VERSION_ID}.noarch.rpm"
+        as_sudo "yum -y install ${EPEL_URL}"
```
avolkov1, line 46: This still didn't work. In the script that `VERSION_ID` appears as a decimal such as 7.9 or 8.3

Then it tries that url:
<https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.3.noarch.rpm>

But it doesn't exist. I modified the code to extract the first digit:
```
        version_=$(echo ${VERSION_ID} | sed  's/^[^0-9]*//;s/[^0-9].*$//')
        EPEL_URL="https://dl.fedoraproject.org/pub/epel/epel-release-latest-${version_}.noarch.rpm"
```

Then it goes to a valid url:
<https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm>

And the setup proceeds correctly.

avolkov1 (changes requested):
Script still goes to an invalid epel url. I made a comment with a fix suggestion that seems to work.

scripts/setup.sh
ajdecon: Thanks! Applied this fix.

avolkov1 (approved):
The epel install works now. Thank you!

michael-balint (approved):

Issue #870, opened 2021-02-16
`sinfo --version` hangs if SRV records exist but slurm cluster does not exist yet


ehudkaldor:
In [/roles/slurm/tasks/build.yml](https://github.com/NVIDIA/deepops/blob/master/roles/slurm/tasks/build.yml) the code tries to get the slurm cluster version by running `sinfo --version`:

```- name: check installed slurm version
  shell: "{{ slurm_install_prefix }}/bin/sinfo --version | awk '{print $2}'"
  register: sinfo_version
  ignore_errors: yes
```

in Slurm 20, if SRV records already exist in the DNS but the cluster is not up yet, `sinfo` will hang forever. to get only the current node's (binary) version, you should use `slurmd --info`. Creating a PR for that now.

PR #871, opened 2021-02-16
Update build.yml

ehudkaldor:
swapped `sinfo --version` with `slurmd --version`. fixes issue #870

### DIFF
--- a/roles/slurm/tasks/build.yml
+++ b/roles/slurm/tasks/build.yml
@@ -36,24 +36,24 @@
     slurm_downgrade: no
 
 - name: check installed slurm version
-  shell: "{{ slurm_install_prefix }}/bin/sinfo --version | awk '{print $2}'"
-  register: sinfo_version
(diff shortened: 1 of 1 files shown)


ajdecon (approved):
Tested this PR, LGTM.

Test plan as follows:

- **Initial install**: Verified I can turn up a virtual cluster with default of Slurm 20.02.4 with this PR, no errors.
- **Clean run on existing cluster**: Re-ran the Slurm playbook and confirmed that we correctly detect not to build again.
- **Upgrade**: Set `slurm_version: "20.11.3"` and confirmed we correctly detect to build with the new version.

ajdecon:
@ehudkaldor : I've tested that this works with our typical workflows, and agree `slurmd --version` is a better version check, so I'm merging the PR. Thanks for your help!

I don't happen to have a convenient test environment where I can set SRV records, so you'll have to confirm on your end if this fixes #870 for you.

ehudkaldor:
closes #870

PR #872, opened 2021-02-16
Update our default Slurm version to 20.11.3

ajdecon:
The 20.11 branch of Slurm seems to be stable enough at this point to get broad adoption, so let's bump our default version to the latest 20.11.3.

## Test plan

- [x] Virtual cluster turnup with new Slurm version works
- [x] Basic job submission tests
- [x] Verify Jenkins tests all pass


dholt:
Also need to update the version here: https://github.com/NVIDIA/deepops/blob/9b97f1175c778ac1817a2396b833549bc0411e3d/config.example/group_vars/slurm-cluster.yml#L6

dholt (approved):

PR #873, opened 2021-02-16
Add documentation on how to configure DeepOps

ajdecon:
Configuring DeepOps isn't very straightforward for new users, especially those without previous Ansible experience. 

Ideally we'd find ways to make this easier to use (e.g., an install wizard sort of experience?)... 

But for now, I've written up some documentation on how the configuration directory works, how to use custom Ansible vars, and how to add your own custom playbooks. This includes a bunch of advice that I've given folks ad-hoc in Slack and elsewhere, hopefully in a more polished format. 😄 

Changes include:

- Adding a document on configuring DeepOps
- Adding a `config.example/playbooks` directory with an example playbook, to demonstrate the pattern of writing custom playbooks
- Adding a `config.example/host_vars` directory with an example host file, to demonstrate host-specific variables
- Adding a link to the new doc from the getting started guide


config.example/host_vars/example01
```diff
@@ -0,0 +1,6 @@
+---
```
supertetelman, line 1: Would it be better if we made this gpu01? That would make it more consistent with the examples we have in the inventory file and as long as everything is commented out it shouldn't conflict with anything. I'd be worried that users will see "example01" and not necessarily correlate that to the host.

config.example/playbooks/example.yml
```diff
@@ -0,0 +1,14 @@
...
+# Your custom playbooks should go in your DeepOps configuration directory,
+# under `config/playbooks`.
+# These playbooks can be used to make any customizations to your cluster
+# that aren't already provided by DeepOps!
+# For example, this playbook installs `cowsay` on all your Kubernetes nodes:
+
+- hosts: kube-node
+  become: yes
+  tasks:
+  - name: install cowsay
```
supertetelman, line 11: +1 for cowsay. Let's hope nobody runs this locally and can't figure out how to turn it off. :)

config.example/playbooks/example.yml
```diff
@@ -0,0 +1,14 @@
+---
+# Your custom playbooks should go in your DeepOps configuration directory,
+# under `config/playbooks`.
+# These playbooks can be used to make any customizations to your cluster
+# that aren't already provided by DeepOps!
+# For example, this playbook installs `cowsay` on all your Kubernetes nodes:
```
supertetelman, line 6: Since this is for new users, should we link them to the Ansible docs?  I could go either way on this one. https://docs.ansible.com/ansible/latest/user_guide/playbooks.html

docs/README.md
```diff
@@ -23,6 +23,11 @@ git checkout tags/20.12
 
 3. Pick one of the [Deployment Options](/README.md#deployment-options) mentioned in the main [README](/README.md), following the installation instructions. For example, if deploying a hybrid cluster, all features of DeepOps may be deployed by following the [DGX POD Hybrid Cluster](deepops/dgx-pod.md) guide.
 
+## Configuration
+
+The various components of DeepOps include a variety of configuration options which can be adjusted using the [DeepOps configuration directory](../config.example).
+To learn more about how to use this directory and customize your DeepOps cluster, read more about [how to configure DeepOps](./deepops/configuration.md).
```
supertetelman, line 29: Wording is strange, recommend:

To learn more about how to use this directory and customize your DeepOps cluster, see [how to configure DeepOps](./deepops/configuration.md).

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+- `config/group_vars/slurm-cluster.yml`: Variables specific to deploying Slurm clusters
+
+It's expected that most DeepOps deployments will make changes to these files!
+The inventory file will be different for every cluster;
+but we also expect that most people will make changes to the Ansible variables that determine how DeepOps behaves.
+While the example configuration contains good defaults (in our opinion!), every cluster is different, and a lot of the features in DeepOps can be configured or are optional.
+
+For example, you may choose to:
+
+- Enable or disable optional features, such as installing a container registry;
```
supertetelman, line 23: No need for the semicolons here

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+- Change the version of software being installed, such as the Slurm version;
```
supertetelman, line 24: Or here.

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+- Or change the download URL used for a software install, if you want to point to a different mirror
+
+However, unless you're developing changes to DeepOps itself, you should never have to make changes *outside your configuration directory*.
+(If you do, that's a sign that we haven't made our playbooks configurable enough, and you should open an issue or PR to fix that!)
+Ideally, we expect that all customizations for a specific cluster should be made in the configuration directory for that cluster.
```
supertetelman, line 29: There are a few exceptions to this in the services directory, such as `workloads/examples/k8s/services/nfs-client.yml`. We may want to consider noting that here.

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Modifying the Ansible inventory
+
+Every cluster deployed with DeepOps will have its own [Ansible inventory file](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html).
+This file defines the list of hosts that are in your DeepOps cluster, and has several sections depending on how your cluster is configured.
+
+In general, we expect that there will always be a section labeled `[all]` that contains a list of every host in the cluster.
+If you need to supply the IP addresses of your hosts, you will usually specify those IPs here as well.
+For example:
+
```
supertetelman, line 40: This would be a good place to mention that these are hostnames and that they will be overwritten if different from existing hostnames.

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+[slurm-master]
+my-cluster-controller
+
+[slurm-node]
+my-cluster-compute-01
+my-cluster-compute-02
+```
+
+If you want to further sub-divide your cluster, or target particular hosts with playbooks, you can add other inventory groups to this file!
+For example, if you have a host that you will use as a separate NFS server, you may want to create an `nfs-server` group and only run the [NFS server playbook](../../playbooks/generic/nfs-server.yml) on that host.
```
supertetelman, line 64: Good example good, should we supply an example kubectl command using the -l flag here? We have them run with -l in a few places in the deployment guide, but I've encountered many users who do not know why they are doing that or realize that it is actually not necessary.

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+
+The variables in `host_vars/example01` would only be applied for a host named `example01`.
+
+
+## Adding custom playbooks
+
+In the process of setting up your cluster, you may want to write your own Ansible playbooks for features that DeepOps doesn't include!
+
+For example, DeepOps doesn't (yet) include a playbook for installing the classic Linux game [Pingus](https://pingus.seul.org/).
+If you wrote an Ansible playbook for installing Pingus on your cluster (to give your users a way to [entertain themselves while their code compiles](https://xkcd.com/303/)),
```
supertetelman, line 143: Classic.

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+you would put it in your `config/playbooks` directory:
+
+```
+playbooks/
```
supertetelman, line 147: Should be config/playbooks

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+## Managing your configuration directory in Git
+
+We recommend creating a separate Git repository for managing your cluster configuration, so that you can track changes to your cluster independently of DeepOps changes and upgrades.
+(The DeepOps [`.gitignore` file](../../.gitignore) is set up to faciliate this by ignoring directories that start with the string `config`, except for `config.example`.)
+
+A good practice is to start a new Git repository by copying the example configuration:
+
+```
+$ cp -R config.example/ config.my-cluster/
+$ cd config.my-cluster/
```
supertetelman, line 169: Why are we suggesting they use config.my-cluster instead of just the config cluster that we create?

docs/deepops/configuration.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+These configuration directories can each have their own inventory files, Ansible variables, and playbooks.
+
+You can then run Ansible for each of the clusters independently by specifying the inventory file on the command line:
+
+```
+ansible-playbook -i config.cluster-1/inventory playbooks/slurm-cluster.yml
+```
+
```
supertetelman, line 219: We might want to add a section on environment variables. Most of our deployment scripts have reasonable defaults, but will allow us to enable/disable features or install specific versions based of GLOBAL_ENVIRONMENT_VARIABLES.

Perhaps we suggest that they maintain a config/environment.sh file and source it before running scripts? In fact, maybe this is something we should consider adding to all of our scripts.

config.example/host_vars/example01
ajdecon: Sure, this makes sense. I'll make this change throughout.

config.example/playbooks/example.yml
ajdecon: Throwing the Ansible docs in here is a good idea, yes.

docs/deepops/configuration.md
ajdecon: Some old draft language sneaking in here, will fix

docs/deepops/configuration.md
ajdecon: I'm not sure this fits in with the rest of the doc, as a lot of it will be script-specific. However, I think documenting that better is definitely a good idea... need to think about the right way to do that.

config.example/playbooks/example.yml
ajdecon: They can learn about `state: absent` ;-)

michael-balint (approved):

PR #684, opened 2020-10-01
Update mig playbook to enable MIG per device rather than all

supertetelman:
Allow MIG to be enabled/disabled on individual devices, not just all.

Introduced a new variable `deepops_mig_devices`. If it is kept at "all" there will be no change. If it is defined or blank there are considered no devices to enable/disable. Otherwise set this to a comma seperated list of device IDs to enable/disable mig on those interfaces.

Having this capability will better allow us to support "mixed" mode in Kubernetes.

PR #876, opened 2021-02-17
Update mig playbook to enable MIG per device rather than all

supertetelman:
Bug fix for reverted PR here: https://github.com/NVIDIA/deepops/pull/684

I was able to test this a bit on a DGX Station A100 (which had some different services running that required modification to the service vars list.).


supertetelman:
It's worth noting that we might be removing this playbook entirely or re-working it to leverage the new mig-parted tool: https://github.com/NVIDIA/mig-parted

dholt (approved):

PR #888, opened 2021-02-24
Add a GPU job test for Slurm

ajdecon:
Right now we don't have a Jenkins test that executes a GPU job under the
Slurm scheduler. I think we validate GPU functionality pretty well under
Kubernetes, but it would be good to check it under Slurm as well.

This test uses the CUDA toolkit in nvhpc to build the `deviceQuery`
sample and run it on the GPU node.

### DIFF
--- a/workloads/jenkins/Jenkinsfile
+++ b/workloads/jenkins/Jenkinsfile
@@ -100,6 +100,11 @@ pipeline {
           sh '''
              bash -x ./workloads/jenkins/scripts/test-rsyslog-slurm.sh
           '''
+
+          echo "Test GPU job"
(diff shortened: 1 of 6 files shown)


supertetelman:
While we are adding this, do we need to add another test that validates Slurm can run a docker/enroot/singularity container?

Also, I recently added the new Jenkinsfile-matrix. It's not super important, but you could also add this test under the Slurm-services section there or the slurm-deployment if this is a core test for Slurm.

Otherwhise, code LGTM.

supertetelman (approved):

supertetelman (approved):

PR #893, opened 2021-02-26
Fix wrong path of DeepOps custom configurations on kubeflow install

henrysecond1:
Below error occurs when installing kubeflow using `deploy_kubeflow.sh`

```
cp: cannot stat 'workloads/services/k8s/kubeflow-install/': No such file or directory
```

Since the script assumes that the command runs in root path, it is not able to find the given directory.
This PR fixed this issue so that the given custom configuration is correctly applied on kubeflow.

### DIFF
--- a/scripts/k8s/deploy_kubeflow.sh
+++ b/scripts/k8s/deploy_kubeflow.sh
@@ -190,7 +190,7 @@ function stand_up() {
   fi
 
   # Copy over DeepOps custom configurations & bugfixes
-  cp -r workloads/services/k8s/kubeflow-install/ config/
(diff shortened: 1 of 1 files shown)


supertetelman:
Thanks for the fix Henry! Would you mind signing your commit and agreeing to this [doc](https://raw.githubusercontent.com/NVIDIA/deepops/master/CONTRIBUTING.md) before I merge it through?

henrysecond1:
> Thanks for the fix Henry! Would you mind signing your commit and agreeing to this [doc](https://raw.githubusercontent.com/NVIDIA/deepops/master/CONTRIBUTING.md) before I merge it through?

I signed it! Thanks for letting me know.

supertetelman (approved):

PR #907, opened 2021-03-13
removed {{ }} from when clause

ryanolson:
Warning from install

```
TASK [set /etc/hostname] ******************************************************************************************************************************************************
[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ deepops_set_hostname | default(true) }}
[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ deepops_set_hostname | default(true) }}
[WARNING]: conditional statements should not include jinja2 templating delimiters such as {{ }} or {% %}. Found: {{ deepops_set_hostname | default(true) }}
```

### DIFF
--- a/playbooks/generic/hosts.yml
+++ b/playbooks/generic/hosts.yml
@@ -5,7 +5,7 @@
     - name: set /etc/hostname
       hostname:
         name: "{{ inventory_hostname }}"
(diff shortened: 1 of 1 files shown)


ajdecon (approved):
LGTM

PR #909, opened 2021-03-15
Fixes for DGX firmware update role

ajdecon:
## Summary

- Correct variable for firmware cleanup
- Correctly stop services and unload modules before updating
- Fix docker run command

## Test plan

Used to update DGX A100 firmware successfully!

### DIFF
--- a/roles/nvidia-dgx-firmware/tasks/main.yml
+++ b/roles/nvidia-dgx-firmware/tasks/main.yml
@@ -30,6 +30,22 @@
   - include: load-image.yml
     when: load_firmware
 
+  # Shut down services and unload drivers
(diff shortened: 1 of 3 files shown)


roles/nvidia-dgx-firmware/tasks/main.yml
```diff
@@ -30,6 +30,22 @@
   - include: load-image.yml
     when: load_firmware
 
+  # Shut down services and unload drivers
```
supertetelman, line 33: Should we add a note that this is to be extra safe or work around current firmware container limitations here?

supertetelman (approved):
LGTM

PR #912, opened 2021-03-16
New setup script

dholt:
Adds a new setup script that aims to make fewer modifications to the system and conflict less with existing software. Also fixes a few bugs with the old setup script.


scripts/setup.sh
```diff
@@ -54,7 +60,7 @@ case "$ID" in
         ${PIP} --version
 
         # Use virtualenv vs system pip when we're running under Jenkins
-        if ! [ -z "${VIRT_DIR}" ] && ! [ -z "${JENKINS}" ]; then
+        if ! [ -z "${VIRT_DIR}" ] && ! [ -z "${JENKINS}" ] ; then
```
supertetelman, line 63: Extra space?

scripts/setup-2.sh
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env bash
+
```
supertetelman, line 2: Can you add some more documentation to the top of this script regarding what it is for, what it does, etc?

scripts/setup-2.sh
```diff
@@ -0,0 +1,142 @@
...
+DEPS_EL8=(git python3-virtualenv sshpass wget)
+EPEL_VERSION="$(echo ${VERSION_ID} | sed  's/^[^0-9]*//;s/[^0-9].*$//')"
+EPEL_URL="https://dl.fedoraproject.org/pub/epel/epel-release-latest-${EPEL_VERSION}.noarch.rpm"
+PROXY_USE=`grep -v ^# ${SCRIPT_DIR}/deepops/proxy.sh 2>/dev/null | grep -v ^$ | wc -l`
+
+# No interactive prompts from Apt during this process
+export DEBIAN_FRONTEND=noninteractive
+
+# Exit if run as root
+if [ $(id -u) -eq 0 ] ; then
```
supertetelman, line 33: Extra space?

supertetelman:
Aside from some of the comments I left, should we add some automated testing for this? I'm thinking we could add a flag in the setup scripts to swap between the two setup scripts and then we test ubuntu 18/centos7 with setup.sh and ubuntu 20/centos8 with setup-2.sh

scripts/setup.sh
dholt: I like it like that ;) but I can remove it since I didn't make the rest of the file consistent.

dholt:
I can also make this the regular setup script in this PR and it will get tested as-is.

I think in general it would be nice to have something cleaner like this. My only reluctance is forcing everyone to deal with a virtual env. Not sure yet if we want to go that route.

scripts/setup.sh
```diff
@@ -1,290 +1,153 @@
...
+
+# Configuration
+ANSIBLE_VERSION="${ANSIBLE_VERSION:-2.9.5}"     # Ansible version to install
+ANSIBLE_OK="${ANSIBLE_OK:-2.7.8}"               # Oldest allowed Ansible version
+CONFIG_DIR="${CONFIG_DIR:-./config}"            # Default configuration directory location
+DEEPOPS_TAG="${1:-master}"                      # DeepOps branch to set up
+JINJA2_VERSION="${JINJA2_VERSION:-2.11.1}"      # Jinja2 required version
+PIP="${PIP:-pip3}"                              # Pip binary to use
+PYTHON_BIN="${PYTHON_BIN:-/usr/bin/python3}"    # Python3 path
+VENV_DIR="${VENV_DIR:-/opt/deepops/env}"        # Path to python virtual environment to create
```
ajdecon, line 18: I don't love that this is in `/opt` on the Ansible controller (which is possibly just the user workstation!)

Possibly put this in their home directory, e.g. `${HOME}/.deepops`? 

I'm not sure of the right answer tbh 🤷🏻‍♂️ it just feels odd to be adding directories outside the homedir.

scripts/setup.sh
```diff
@@ -1,290 +1,153 @@
...
+    if [ ! -d "${CONFIG_DIR}" ] ; then
+        cp -rfp ./config.example "${CONFIG_DIR}"
+        echo "Copied default configuration to ${CONFIG_DIR}"
+    else
+        echo "Configuration directory '${CONFIG_DIR}' exists, not overwriting"
+    fi
 fi
+
+# Add Ansible virtual env to PATH when using Bash
+if [ -f "${VENV_DIR}/bin/activate" ] ; then
```
ajdecon, line 145: I'm unsure whether I like doing this by default, or whether I'd prefer to print instructions for the user to do this.

If possible, a flag to enable/disable this would be handy.

ajdecon (approved):
I like this a lot! I actually do think using a virtualenv is a good idea, because it somewhat isolates the changes we're making to the system.

I've commented inline with a few nits, but none of those are blockers.

Approving so you can merge when you like, but not merging so others can leave feedback.

PR #913, opened 2021-03-16
Move nvidia-peer-memory logic into a role

ajdecon:
This is mostly to support future refactoring and make the layout consistent with what we do in other playbooks/roles.

Test plan is passing CI tests.


supertetelman (approved):
LGTM

PR #916, opened 2021-03-18
Add network operator to enable multinode RoCE

yangatgithub:


roles/nvidia-network-operator/files/values.yaml
```diff
@@ -0,0 +1,61 @@
+nfd:
+  enabled: true
+sriovNetworkOperator:
+  enabled: true
+node-feature-discovery:
+  worker:
+    tolerations:
```
supertetelman, line 7: I don't think this is 100% correct.

Looking for nvidia.com/gpu will cause this to not launch on MIG nodes. The more generic label that is applied to nodes by GFD is nvidia.gpu.count (see the device plugin role: https://github.com/NVIDIA/deepops/blob/master/roles/nvidia-k8s-gpu-device-plugin/tasks/main.yml)

Is the behavior we want for non-MIG nodes only?

roles/nvidia-network-operator/tasks/k8smaster.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
supertetelman, line 1: Can we give this file a more descriptive (install-network-operator.yml) or more standard (main.yml) name for this file?

roles/nvidia-network-operator/tasks/k8smaster.yaml
```diff
@@ -0,0 +1,25 @@
...
+# git clone; helm install; create node policy; create networkattachdefinition
+#
+
+- name: Git clone
```
supertetelman, line 5: "Git clone network operator"

roles/nvidia-network-operator/tasks/k8smaster.yaml
```diff
@@ -0,0 +1,25 @@
...
+  ansible.builtin.git:
+    repo: 'https://github.com/Mellanox/network-operator.git'
+    dest: /tmp/network-operator
+
+- name: Deploy network-operator chart from local path
+  community.kubernetes.helm:
+    name: network-operator
+    chart_ref: /tmp/network-operator/deployment/network-operator/
```
supertetelman, line 13: Why are we storing things in /tmp? We should try to download things to the local deepops dir (~/deepops) or into the {{ deepops_dir }} location (/opt/deepops).

In fact, I think a better way to do this overall would be to follow the pattern I set with the nfs-client-provisioner. In that case I added the external repo as a git submodule (https://github.com/NVIDIA/deepops/blob/master/.gitmodules), this gets downloaded into the submodules directory (at a specific commit) during setup. After that I run all the helm install commands using localhost, rather than downloading the files to the master nodes and then expecting them to have the right version of helm install and also having RW in /tmp.

playbooks/k8s-cluster/roce-network-operator.yaml
```diff
@@ -0,0 +1,9 @@
+---
+# Playbook for installing Network Operator by Helm
+- hosts: kube-master
```
supertetelman, line 3: kube-master[0] to avoid attempting download/helm install multiple times on HA clusters.

playbooks/k8s-cluster/roce-network-operator.yaml
```diff
@@ -0,0 +1,9 @@
...
+  become: true
+  become_method: sudo
+  tasks:
+  - include_role:
+      name: nvidia-network-operator
+      tasks_from: k8smaster
```
supertetelman, line 9: Recommend using a different name as mentioned below.

roles/nvidia-network-operator/tasks/k8smaster.yaml
supertetelman: See https://github.com/NVIDIA/deepops/blob/master/roles/nfs-client-provisioner/tasks/main.yml for an example on how I delegated localhost to run helm install for the cloned repos.

If the git clone is not necessary, it would also be best-practice to do a helm repo add/helm repo update in place of a git clone (see the GPU Device Plugin role for an example of this).

supertetelman:
Can we add a Jenkins test for this?I'm thinking something like the [kubeflow script](https://github.com/NVIDIA/deepops/blob/master/workloads/jenkins/scripts/test-kubeflow.sh) that would run an ansible command to install the operator and then either run a sample workload or verify that the expected PODs are launching.

I do not know if this operator requires special hardware to actually run, so testing could be difficult.

Issue #784, opened 2020-12-07
slurm validation playbook fails - Pyxis/Enroot failing to run jobs on CentOS


miketice22:
After installing/deploying slurm with 'ansible-playbook -l slurm-cluster playbooks/slurm-cluster.yml' The validation playbook fails.

```
$ ansible-playbook -l slurm-cluster playbooks/slurm-cluster/slurm-validation.yml -e '{num_gpus: 1}'
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details

PLAY [slurm-master[0]] **********************************************************************************************************************************************************************************************

TASK [Get node count from sinfo] ************************************************************************************************************************************************************************************
changed: [aplcdhen01.datalake.jhuapl.edu]

TASK [Set num_nodes variable] ***************************************************************************************************************************************************************************************
ok: [aplcdhen01.datalake.jhuapl.edu]

TASK [Set cmd variable] *********************************************************************************************************************************************************************************************
ok: [aplcdhen01.datalake.jhuapl.edu]

TASK [Print node/gpu counts] ****************************************************************************************************************************************************************************************
ok: [aplcdhen01.datalake.jhuapl.edu] => 
  msg:
  - Detected 1 nodes with 1 gpus each.
  - 'Proceeding to run validation test, this may take several minutes: srun -N 1 -G 1 --ntasks-per-node=1 --mpi=pmix --exclusive  --container-image=deepops/nccl-tests-tf20.06-ubuntu18.04:latest all_reduce_perf -b 1M -e 4G -f 2 -g 1.'

TASK [Execute NCCL test across all nodes and GPUs] ******************************************************************************************************************************************************************
fatal: [aplcdhen01.datalake.jhuapl.edu]: FAILED! => changed=true 
  cmd: |-
    srun -N 1 -G 1 --ntasks-per-node=1 --mpi=pmix --exclusive  --container-image=deepops/nccl-tests-tf20.06-ubuntu18.04:latest all_reduce_perf -b 1M -e 4G -f 2 -g 1
  delta: '0:01:47.939670'
  end: '2020-12-07 19:07:46.163125'
  msg: non-zero return code
  rc: 1
  start: '2020-12-07 19:05:58.223455'
  stderr: |-
    pyxis: importing docker image ...
    pyxis: creating container filesystem ...
    pyxis: starting container ...
    slurmstepd: error: pyxis: container start failed with error code: 1
    slurmstepd: error: pyxis: printing contents of log file ...
[...]

PR #918, opened 2021-03-19
Update to nvidia.enroot v0.4.0 role

ajdecon:
## Summary

This adds support for configuring the kernel command line options for Enroot on EL distros.

Also makes the following DeepOps changes:
- By default, only run nvidia.enroot role in group `slurm-node`, instead of on `slurm-cluster`. We make this change because `enroot` is only needed on the compute nodes, and following this change, adding the kernel flags can result in nodes rebooting.
- Pyxis role should only set up enroot hooks on `slurm-node`, for the same reason.
- Re-enables Jenkins CI tests for Enroot on CentOS.

Addresses #784 .

## Test plan

Tested manually to confirm that Enroot now works on CentOS.

Need to validate Jenkins CI tests pass for all stacks.

### DIFF
--- a/playbooks/container/pyxis.yml
+++ b/playbooks/container/pyxis.yml
@@ -21,8 +21,12 @@
         enroot_environ_config_files: "{{ enroot_environ_config_files_dgx }}"
       when: ansible_product_name is search("DGX")
 
+- hosts: slurm-node
+  become: yes
+  roles:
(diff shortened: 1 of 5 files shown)


supertetelman:
Looks good to me and passing tests.

supertetelman (approved):

PR #920, opened 2021-03-24
Fix multinode testing

supertetelman:
I broke the testing a few weeks back with some of the new CentOS 8 tests. We fixed it in the single node tests, but the commit for the multinode case was accidentally pushed after the PR went through.


ajdecon (approved):
LGTM

PR #922, opened 2021-03-25
Install DCGM using the CUDA repositories

ajdecon:
## Summary

Since DCGM is now open source, the user does not need to manually download the package using their NVIDIA Developer account. Instead, DCGM can be installed directly from the CUDA repositories.

This PR replaces the "manual download and install" flow with the recommended installation flow from the [DCGM download page](https://developer.nvidia.com/dcgm#Downloads).

We also enable DCGM by default in the Slurm and NGC Ready top-level playbooks.

## Test plan

Run the DCGM playbook without doing anything special for package downloads:

```
$ ansible-playbook --flush-cache -i virtual/config/inventory -l virtual-gpu01 playbooks/nvidia-software/nvidia-dcgm.yml
```

Confirm the playbook executes successfully, and that dcgm is now installed:

```
vagrant@ubuntu1804:~$ dpkg -l | grep datacenter
ii  datacenter-gpu-manager                1:2.1.7                           amd64        NVIDIA® Datacenter GPU Management Tools
```

We should also expect to see this playbook succeed in CI tests now that it's enabled by default in configuration.

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -243,17 +243,7 @@ maas_kvm: false
 ################################################################################
 # NVIDIA Datacenter GPU Manager                                                #
 ################################################################################
-# DCGM is available through package repositories for DGX, but for other host
-# types you must download from the NVIDIA Developer portal at
-# https://developer.nvidia.com/dcgm
-# To install DCGM, please download the correct package type(s) and set the
-# following variables to a path where the file exists on the Ansible control
-# machine.
-#
-# dcgm_rpm_package: "/tmp/datacenter-gpu-manager.rpm"
-# dcgm_deb_package: "/tmp/datacenter-gpu-manager.deb"
-
-install_dcgm: false
+install_dcgm: true 
 
 ################################################################################
 # Misc.                                                                        #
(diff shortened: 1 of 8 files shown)


config.example/group_vars/all.yml
```diff
@@ -243,17 +243,7 @@ maas_kvm: false
...
-# https://developer.nvidia.com/dcgm
-# To install DCGM, please download the correct package type(s) and set the
-# following variables to a path where the file exists on the Ansible control
-# machine.
-#
-# dcgm_rpm_package: "/tmp/datacenter-gpu-manager.rpm"
-# dcgm_deb_package: "/tmp/datacenter-gpu-manager.deb"
-
-install_dcgm: false
+install_dcgm: true
```
dholt, line 246: extra space at end of line :)

roles/nvidia-dcgm/tasks/install-redhat.yml
```diff
@@ -1,18 +1,21 @@
 ---
-- name: ensure package var is defined
-  fail:
-    msg: "dcgm_rpm_package must be defined with file path to package"
-  when: dcgm_rpm_package is not defined
+- name: RedHat | add epel repo
+  become: yes
+  yum:
+    name:
```
dholt, line 5: extra space

dholt (dismissed):

dholt (approved):

PR #923, opened 2021-03-25
Add better/consistent vagrant cleanup to jenkins vagrant-startup script

supertetelman:
This adds a call to vagrant_shutdown in the Jenkins vagrant-startup scrip.

There were a few edge-cases in the nightly testing where tests were not properly cleaning up after a failure and then causing the next tests to fail. This will address that last cleanup edge case.


ajdecon (approved):
LGTM

PR #934, opened 2021-04-19
Update Trident deployment role to use Helm chart

mboglesby:
Updating the 'netapp-trident' role to use the new Trident Helm chart to deploy Trident.

### DIFF
--- a/config.example/group_vars/netapp-trident.yml
+++ b/config.example/group_vars/netapp-trident.yml
@@ -89,6 +89,30 @@ storageClasses_to_create:
       backendType: "ontap-nas-flexgroup"
   # Add additional StorageClasses as needed
 
+# Denotes whether or not to enable volume snapshots in your Kubernetes cluster
+# For more info, refer to: https://netapp-trident.readthedocs.io/en/stable-v21.01/kubernetes/operations/tasks/volumes/snapshots.html
+enable_volume_snapshots: true
+
+# List of volume snapshot CRDs to create
+volume_snapshot_CRD_manifest_URLs:
+  - https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/release-3.0/client/config/crd/snapshot.storage.k8s.io_volumesnapshotclasses.yaml
+  - https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/release-3.0/client/config/crd/snapshot.storage.k8s.io_volumesnapshotcontents.yaml
+  - https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/release-3.0/client/config/crd/snapshot.storage.k8s.io_volumesnapshots.yaml
+
+# List of snapshot controller manifest URLs
+snapshot_controller_manifest_URLs:
+  - https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/release-3.0/deploy/kubernetes/snapshot-controller/rbac-snapshot-controller.yaml
+  - https://raw.githubusercontent.com/kubernetes-csi/external-snapshotter/release-3.0/deploy/kubernetes/snapshot-controller/setup-snapshot-controller.yaml
+
+# List of volume snapshot classes to create
+volume_snapshot_classes_to_create:
+  - apiVersion: snapshot.storage.k8s.io/v1beta1
+    kind: VolumeSnapshotClass
+    metadata:
+      name: csi-snapclass
+    driver: csi.trident.netapp.io
+    deletionPolicy: Delete
+
 # Denotes whether or not to copy tridenctl binary to localhost
 copy_tridentctl_to_localhost: true
 # Directory that tridentctl will be copied to on localhost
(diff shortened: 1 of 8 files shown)


roles/netapp-trident/tasks/main.yml
```diff
@@ -7,153 +7,23 @@
...
-- name: Create Trident namespace (ubuntu)
-  k8s:
-    state: present
-    kind: Namespace
-    name: trident
-    wait: yes
+- name: Copy namespace config file to control node
+  template:
+    src: namespace.j2
+    dest: /root/trident-installer/namespace.yaml
```
supertetelman, line 13: Can we replace /root-rident-installer with a variable rather than hardcoding this path place?

Also, we've been trying to keep most of the deepops files in a cenral place. We define deepops_dir in all.yaml and by default that is /opt/deepops. If there is a bunch of existing Trident documentation that references /root/trident-installer, I think it makes sense to leave it there. But I'd still replace this directory with a variable and reference that in all the places that follow.

config.example/group_vars/netapp-trident.yml
```diff
@@ -2,14 +2,11 @@
...
+trident_version: "21.01.2"
 trident_installer_url: "https://github.com/NetApp/trident/releases/download/v{{ trident_version }}/trident-installer-{{ trident_version }}.tar.gz"
 
-# Kubernetes version
-# Note: Do not include patch version, e.g. provide value of 1.16, not 1.16.7.
-# Note: Versions 1.14 and above are supported when deploying Trident with DeepOps.
-#   If you are using an earlier version, you must deploy Trident manually.
-k8s_version: 1.17
+# Namespace to install Trident in
+trident_namespace: trident
```
supertetelman, line 9: I can see leaving this as-is if there is documentation/examples around this.

But in general, we have been trying to refactor new/existing deepops namespaces to be "deepops-<application>"

roles/netapp-trident/tasks/main.yml
```diff
@@ -166,26 +36,30 @@
...
+  when: create_StorageClasses == true
   run_once: true
-  environment:
-    PYTHONHOME: '{{ deepops_venv }}'
-  when: create_StorageClasses == true and ansible_distribution == 'Ubuntu'
 
-- name: Create NetApp StorageClass(es) (Red Hat / CentOS)
-  k8s:
+- name: Create StorageClasses
+  kube:
```
supertetelman, line 55: Does this work with a standard/default DeepOps install? I seem to recall seeing a bunch of TODOs around this and most of are kubectl commands directly use the binary throuh "shell".

I'll take your word for it if you've tested this.

supertetelman (changes requested):
This looks like a great set of changes if you can just address the two comments around naming/directories.

I don't currently have a NetApp device to validate this on, but I'm assuming you've run through this a few times on fresh installs of DeepOps?

Is there any additional README/documentation we need to add for this as well? It looks like you made the relevant changes, but now would be a good time to add any additional examples you think might help.

roles/netapp-trident/tasks/main.yml
mboglesby: @supertetelman - I'll replace this hardcoded path with a variable, and I'll incorporate the existing deepops_dir variable. Stay tuned for an updated PR.

config.example/group_vars/netapp-trident.yml
mboglesby: We can use "deepops-trident" instead. I'll make this change.

roles/netapp-trident/tasks/main.yml
mboglesby: @supertetelman - I actually switched to using the "kube" module instead of the "k8s" module because I was running into a bunch of weird Python interpreter when using the "k8s" module with Ubuntu 20.04. I noticed that Kubespray was using the "kube" module instead, so I modeled my steps after theirs. I tested all of this on a clean install using standard/default DeepOps, and it worked.

mboglesby:
@supertetelman -

> This looks like a great set of changes if you can just address the two comments around naming/directories.
> 
> I don't currently have a NetApp device to validate this on, but I'm assuming you've run through this a few times on fresh installs of DeepOps?

Yes, I validated all of this a few times on a fresh install.

> 
> Is there any additional README/documentation we need to add for this as well? It looks like you made the relevant changes, but now would be a good time to add any additional examples you think might help.

I updated all of the README/docs that needed to be updated. There may be some additional updates required as part of the changes discussed in the other comments above, and I will make those as needed.

roles/netapp-trident/tasks/main.yml
mboglesby: @supertetelman - The PR is updated, but I need to validate the installation process with the changes. I'll ping back after I validate it (will likely be tomorrow).

config.example/group_vars/netapp-trident.yml
mboglesby: @supertetelman - This change has been made.

roles/netapp-trident/tasks/main.yml
mboglesby: @supertetelman - I validated a clean installation with the new changes.

supertetelman (approved):
LGTM

PR #940, opened 2021-04-27
Skips nvidia hpc sdk installation if localrc is already present

### LINKED ISSUE #933: TASK - download nvidia hpc sdk - always runs
I'm doing some prototyping with the SLURM deployment on Ubuntu 20.04 and the 

```
TASK [nvidia-hpc-sdk : download nvidia hpc sdk]
```

always runs and always takes a significant amount of time to download. Can we create a check that skips this if the nvidia hpc sdk has already been deployed?

zyndagj:
Fixes #933 and greatly speeds up iteration when prototyping changes on a SLURM deployment.


ajdecon (approved):
LGTM

PR #941, opened 2021-04-27
Update dcgm-exporter version

ajdecon:
This PR updates `dcgm-exporter` version to a version which supports MIG metrics. We don't actually select these metrics or add any specific dashboards for them at this time, but this should lay the ground work for users who wish to use this.

Stand-alone install:

- [x] Update dcgm-exporter role to pull from nvcr.io by default
- [x] Pin to a particular container tag rather than latest
- [x] Specify tag for dcgm-exporter that supports MIG metrics
- [x] Add capability to specify a custom set of metrics via a CSV file
- [x] Tweak default list of metrics to enable our current dashboards with the new version
- [x] Verify all expected metrics show up on dashboard

k8s install (no GPU operator):

- [x] Update tag to version that supports MIG metrics
- [x] Add capability to specify a custom set of metrics via a CSV file
- [x] Tweak default list of metrics to enable our current dashboards with the new version
- [x] Verify all expected metrics show up on dashboard

~~k8s install (GPU operator):~~

- [ ] ~~Update tag to version that supports MIG metrics~~
- [ ] ~~Add capability to specify a custom set of metrics via a CSV file~~
- [ ] ~~Tweak default list of metrics to enable our current dashboards with the new version~~
- [ ] ~~Verify all expected metrics show up on dashboard~~

### DIFF
--- a/scripts/k8s/deploy_monitoring.sh
+++ b/scripts/k8s/deploy_monitoring.sh
@@ -26,6 +26,7 @@ ingress_name="ingress-nginx"
 
 PROMETHEUS_YAML_CONFIG="${PROMETHEUS_YAML_CONFIG:-${DEEPOPS_CONFIG_DIR}/helm/monitoring.yml}"
 PROMETHEUS_YAML_NO_PERSIST_CONFIG="${PROMETHEUS_YAML_NO_PERSIST_CONFIG:-${DEEPOPS_CONFIG_DIR}/helm/monitoring-no-persist.yml}"
+DCGM_CONFIG_CSV="${DCGM_CONFIG_CSV:-${DEEPOPS_CONFIG_DIR}/files/k8s-cluster/dcgm-custom-metrics.csv}"
 
 function help_me() {
     echo "This script installs the DCGM exporter, Prometheus, Grafana, and configures a GPU Grafana dashboard."
@@ -149,6 +150,11 @@ function setup_gpu_monitoring() {
     if ! kubectl -n monitoring get configmap kube-prometheus-grafana-gpu >/dev/null 2>&1 ; then
         kubectl create configmap kube-prometheus-grafana-gpu --from-file=${ROOT_DIR}/src/dashboards/gpu-dashboard.json -n monitoring
         kubectl -n monitoring label configmap kube-prometheus-grafana-gpu grafana_dashboard=1
+    fi
+
+    # Create DCGM metrics config map
+    if ! kubectl -n monitoring get configmap dcgm-custom-metrics >/dev/null 2>&1 ; then
+        kubectl create configmap dcgm-custom-metrics --from-file=${DCGM_CONFIG_CSV} -n monitoring
     fi
 
     # Label GPU nodes
--- a/workloads/services/k8s/dcgm-exporter.yml
+++ b/workloads/services/k8s/dcgm-exporter.yml
@@ -23,7 +23,7 @@ spec:
       nodeSelector:
         hardware-type: NVIDIAGPU
       containers:
-      - image: "nvcr.io/nvidia/k8s/dcgm-exporter:2.0.13-2.1.2-ubuntu20.04"
+      - image: "nvcr.io/nvidia/k8s/dcgm-exporter:2.1.8-2.4.0-rc.2-ubuntu20.04"
         name: nvidia-dcgm-exporter
         env:
         - name: "DCGM_EXPORTER_LISTEN"
--- a/workloads/services/k8s/dcgm-exporter.yml
+++ b/workloads/services/k8s/dcgm-exporter.yml
@@ -25,6 +25,7 @@ spec:
       containers:
       - image: "nvcr.io/nvidia/k8s/dcgm-exporter:2.1.8-2.4.0-rc.2-ubuntu20.04"
         name: nvidia-dcgm-exporter
+        command: ["/usr/bin/dcgm-exporter", "-f", "/etc/dcgm-config/dcgm-custom-metrics.csv"]
         env:
         - name: "DCGM_EXPORTER_LISTEN"
           value: ":9400"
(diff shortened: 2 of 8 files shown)


workloads/services/k8s/dcgm-exporter.yml
```diff
@@ -23,7 +23,7 @@ spec:
       nodeSelector:
         hardware-type: NVIDIAGPU
```
supertetelman, line 24: While we are updating this, should we update this nodeSelector to use labels applied by the GFD?

GFD is standard in all K8s installs at this point, so I'd like to start removing traces of this if nobody has any objections.

supertetelman:
Strange, I'm not seeing any device utilization information on my standard DGX Station test box and Prometheus returns nothing for `DCGM_FI_DEV_DEC_UTIL`. I'll try to validate this on another system, but it is possible this introduces some sort of conflict with the default driver version, grafana dashboard, and dcgm-exporter version.

supertetelman:
Same issue on a DGX Station A100 with the latest DGX OS 5.x

dholt:
@supertetelman I was seeing the same issue (lacking GPU utilization) with the last version of this container too, so I'm not sure this is introducing a new problem

roles/nvidia-gpu-operator/defaults/main.yml
```diff
@@ -19,7 +19,17 @@ gpu_operator_namespace: "gpu-operator-resources"
 gpu_operator_default_runtime: "docker"
 gpu_operator_driver_registry: "nvcr.io/nvidia"
 gpu_operator_driver_version: "460.32.03"
-gpu_operator_plugin_args: "--mig-strategy={{ k8s_gpu_mig_strategy }},--pass-device-specs=true,--fail-on-init-error=true,--device-list-strategy=envvar,--nvidia-driver-root=/run/nvidia/driver"
+deepops_dir: "/opt/deepops"
+gpu_operator_config_dir: "{{ deepops_dir }}/gpu-operator"
+gpu_operator_plugin_args:
```
supertetelman, line 24: Have you tested that this works? It's used in the below way, and I was having a very hard time getting Ansible to parse all this correctly.

I was trying to avoid checking a new config yml into the repo or creating a temporary one, but it might be about time to do that.

"--set devicePlugin.args={"{{ gpu_operator_plugin_args }}"}"

roles/nvidia-gpu-operator/defaults/main.yml
```diff
@@ -19,7 +19,17 @@ gpu_operator_namespace: "gpu-operator-resources"
...
+- "--mig-strategy={{ k8s_gpu_mig_strategy }}"
+- "--pass-device-specs=true"
+- "--fail-on-init-error=true"
+- "--device-list-strategy=envvar"
+- "--nvidia-driver-root=/run/nvidia/driver"
+gpu_operator_dcgm_version: "2.1.8-2.4.0-rc.2-ubuntu20.04"
```
supertetelman, line 30: We need to remember to upload this var (and all the above defaults) every time we bump the version of the GPU Operator.

I have been trying to keep our defaults in-sync with the vars in the GPU Operator values.yml.

roles/nvidia-gpu-operator/tasks/main.yml
```diff
@@ -27,7 +26,24 @@
...
+- name: Copy values.yml to host
+  template:
+    src: "{{ gpu_operator_values_file }}"
+    dest: "{{ gpu_operator_config_dir }}/{{ gpu_operator_values_file }}"
+
 # XXX: This currently installs into the default namespace, as per the GPU Operator docs
 # https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/getting-started.html
 - name: install nvidia gpu operator
-  command: /usr/local/bin/helm upgrade --install "{{ gpu_operator_release_name }}" "{{ gpu_operator_chart_name }}" --version "{{ gpu_operator_chart_version }}" --set migStrategy="{{ k8s_gpu_mig_strategy }}" --set operator.defaultRuntime="{{ gpu_operator_default_runtime }}" --set driver.repository="{{ gpu_operator_driver_registry }}" --set driver.version="{{ gpu_operator_driver_version }}" {{ gpu_operator_registry_secret }} --set gfd.migStrateg="{{ k8s_gpu_mig_strategy }}" --set devicePlugin.args={"{{ gpu_operator_plugin_args }}"} --wait
+  command: /usr/local/bin/helm upgrade --install "{{ gpu_operator_release_name }}" "{{ gpu_operator_chart_name }}" --version "{{ gpu_operator_chart_version }}" -f "{{ gpu_operator_config_dir }}/{{ gpu_operator_values_file }}"
```
supertetelman, line 49: Ah, I see. You just went ahead and used the temp config. This is cleaner now that we have so much junk we are tacking on here.

roles/nvidia-gpu-operator/tasks/main.yml
```diff
@@ -13,7 +13,6 @@
 # TODO: If this already exists we fail, so skip that for now with failed_when: false
 - name: Create namespace for GPU Operator resources
   command: kubectl create namespace {{ gpu_operator_namespace }}
-  when: gpu_operator_create_secret
```
supertetelman, line 16: Not sure we need to remove this. It is only necessary if we need to pre-create a secret.

roles/nvidia-gpu-operator/tasks/main.yml
supertetelman: Actually, I take that back. It looks like you dropped `gpu_operator_registry_secret` and this will cause the vGPU installation to fail.

supertetelman:
This has minor overlap with my old branch where I was trying to transition us to the helm dcgm-exporter install. https://github.com/NVIDIA/deepops/compare/master...supertetelman:dcgm-exporter-helm-install

scripts/k8s/deploy_monitoring.sh
```diff
@@ -151,6 +152,11 @@ function setup_gpu_monitoring() {
         kubectl -n monitoring label configmap kube-prometheus-grafana-gpu grafana_dashboard=1
     fi
 
+    # Create DCGM metrics config map
+    if ! kubectl -n monitoring get configmap dcgm-custom-metrics >/dev/null 2>&1 ; then
+        kubectl create configmap dcgm-custom-metrics --from-file=${DCGM_CONFIG_CSV} -n monitoring
```
ajdecon, line 157: Create or apply?

supertetelman (approved):

PR #943, opened 2021-04-30
Update k8s examples to run on A100, with smaller MIG slices, and usin…

supertetelman:
* The default batch size was causing problems in some smaller GPUs, so we made it smaller
* The newer version of docker requires --gpus all, so we added that flag
* The newer TensorFlow containers have software that better supports A100


ScottESanDiego (approved):
Good change.

PR #946, opened 2021-05-05
Update OOD wrapper role to support Ubuntu 20.04

dholt:


ajdecon (approved):
Tested on virtual cluster with Ubuntu 20.04, LGTM.

image

PR #947, opened 2021-05-06
Fix spelling for setting Ubuntu version in vagrant

ajdecon:
Currently we can't correctly set the version of Ubuntu to 20.04 using our virtual cluster setup, because the env var is mis-spelled in `vagrant_startup.sh`. This PR fixes the spelling mistake.


supertetelman (approved):
Oops

PR #948, opened 2021-05-07
Fix disabling NVIDIA driver install on Slurm cluster install

ajdecon:
The Slurm cluster playbook provides the var `slurm_cluster_install_nvidia_driver` to disable installs of the NVIDIA driver. However, the Slurm cluster playbook also sources the CUDA playbook, which installs the driver itself. And the CUDA playbook does not respect the flag in the Slurm cluster playbook.

Aren't dependencies fun?

This commit adds a flag to the CUDA playbook to allow disabling the driver install, and plumbs it through so the user can set the flag in the Slurm cluster playbook.

To test, set `slurm_cluster_install_nvidia_driver` to false and set up a Slurm cluster, and the driver should not be installed.

### DIFF
--- a/playbooks/nvidia-software/nvidia-cuda.yml
+++ b/playbooks/nvidia-software/nvidia-cuda.yml
@@ -19,6 +19,7 @@
       when:
         - ansible_local['gpus']['count']
         - is_dgx.stat.exists == False
+        - cuda_playbook_install_driver|default(true)
       tags:
       - nvidia_driver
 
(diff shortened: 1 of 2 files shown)


dholt (approved):

dholt:
Just a thought, maybe we should remove the driver install from the CUDA playbook? Historically we've tried to make these playbooks work well standalone, but this could be an instance where it's not needed.

ajdecon:
> Just a thought, maybe we should remove the driver install from the CUDA playbook? Historically we've tried to make these playbooks work well standalone, but this could be an instance where it's not needed.

I thought about doing that (and would technically prefer that!)... But I'm aware of users who just run the CUDA playbook directly to set up their GPU hosts, and that would break their workflow in an annoying way. Hence this, which is kinda a hack 😛

PR #953, opened 2021-05-18
Introduce a common script library, config for env vars, and inject these into all scripts

supertetelman:
This PR gives us the ability to improve the debugabbility and verbosity of all scripts, as well as making the default configuration overriding something that can be version controlled.

We do this by the following:

* Create `scripts/common.sh` to be sourced in all *sh files. This will source the new `env.sh` and will also print out a standard debug header when each script runs.
* Create `config/env.sh` as a place to version control/save the overrides to our global env vars, prevoiusly we had been running `export DEEPOPS_XXX=123` commands, which was not consistent.
* Source common.sh in most of the *sh file


config.example/env.sh
```diff
@@ -0,0 +1,6 @@
+# This file acts as a location to override the default configurations of deepops/scripts/*
+# Many of the scripts in this directory define global variables and set reasonable defaults
+# Global variables (in all caps) that are defined here will be automatically sourced and used in all scripts
+# See deepops/scripts/common.sh for implementation details
+
+DEEPOPS_VERSION="development-branch"
```
dholt, line 6: This is a little confusing, does it set something to a development branch?

dholt (commented):
Looks good to me, just one question about the default env file

config.example/env.sh
supertetelman: This does nothing.  I was just adding an example variable and thought maybe we could put an actual version there to be imported in a future mechanism for easier debug.

dholt: Do you mind commenting it out then? Maybe it's just me, but it looks like it's doing something

supertetelman: Done.

dholt (approved):

dholt (approved):

PR #954, opened 2021-05-19
Create a README that outlines how to request GPUs in K8s with examples

supertetelman:
This PR adds a new README and set of example yaml files. The README explains how to use NodeSelectors, NodeAffinitys, and resource limits to request GPUs. There is a light pros/cons disucssion about the different method and I have outlined snippets of what a user can put in their Pod yaml.

In addition to the README I have included a (untested) yaml for for each of the most common configuration methods.

This still needs to be tested and testing this will require a large set of different resources/cluster-types.

### DIFF
--- a/workloads/examples/k8s/gpu-usage/README.md
+++ b/workloads/examples/k8s/gpu-usage/README.md
@@ -1,0 +1,79 @@
+# Example GPU YAML configurations
+
+This directory contains several yaml files that act as examples of how to consume GPUs within Kubernetes.
+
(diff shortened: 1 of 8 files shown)


supertetelman:
I went through and validated the yaml. It needs to be modified to match the nodes these are being deployed to, but they should work as a valid reference format.

Good to merge after someone give this a review/spellcheck.

ajdecon (approved):
LGTM

PR #961, opened 2021-05-20
yum install docker-py for centos-slurm

supertetelman:
This is the same issue we were seeing in the standalone registry. The Ansible modules are relying on Python 2 in CentOS/RedHat and this requires us to install the python-docker pages with yum and not pip.


ajdecon (approved):
Approving but not merging yet. @supertetelman , I'll let you merge this when your CI test completes.

PR #962, opened 2021-05-26
Add a ansible-galaxy requirements.yml config file

supertetelman:
The existing galaxy requirements.yml file is in `roles`. This is tied to the Galaxy roles required by DeepOps.

This PR introduces a new requirements.yml in the `config` directory that DeepOps users can put their customer requirements in.

Because the provided blank custom requirements file is blank, I have added the `-i` flag to the install of this in the setup script to avoid "no role found" errors.

No testing required beyond running setup.sh in the regression testing.


ajdecon:
I really like this, thank you!

Can you add a change to our [configuration doc](https://github.com/NVIDIA/deepops/blob/master/docs/deepops/configuration.md) that points out the user can add dependencies here?

PR #964, opened 2021-05-27
Install Docker SDK correctly for all Pythons

ajdecon:
## Summary

On distros where we are running under Python 2, we should use the OS package for python-docker in order to ensure compatibility with the distro python.

On distros where we are running under Python 3, we will install via pip to get the correct and most up-to-date package.

## Test plan

Ran Prometheus playbook on each of CentOS 7 and 8, and Ubuntu 18.04 and 20.04. Python 2 was used in CentOS 7, all other distros used Python 3.

### DIFF
--- a/roles/nginx-docker-registry-cache/tasks/server.yml
+++ b/roles/nginx-docker-registry-cache/tasks/server.yml
@@ -1,15 +1,20 @@
 ---
-- name: Install package dependencies
-  package:
-    name:
-    - "python3-pip"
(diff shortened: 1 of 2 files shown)


supertetelman (commented):
Very cool magic ansible variable. LGTM.

supertetelman (approved):

ajdecon:
@dholt FYI this merged against master, you may need to pull it into the release branch

PR #965, opened 2021-05-27
Improved cleanup in Slurm epilog

dholt:
Instead of removing all files in `/tmp` and `/dev/shm` after the last user job on a node, just remove files owned by a particular user once their last job on that node ends.


ajdecon (approved):
LGTM

PR #966, opened 2021-05-27
Introduce a test to verify metrics from the GPU dashboard are being returned by DCGM in Slurm/K8s

supertetelman:
@ajdecon added a new dashboard with custom DCGM metrics and I am introducing this new test to our Slurm/K8s installs to verify that all DGCM* metrics mentioned in our default dashboard are retuned by the deployed DCGM-exporter.

I am not sure what the behavior will be if DCGM is unable to return a metric (such as IB on a non-IB node), but I believe it will return a commented out (`#DCGM_*`) value if it is configured and working, but unable to get the metric.


scripts/k8s/deploy_ingress.sh
```diff
@@ -55,5 +55,5 @@ fi
 # Set up the ingress controller
 if ! helm status "${app_name}" >/dev/null 2>&1; then
 	helm repo update
-	helm install --wait "${app_name}" "${helm_arguments[@]}" ingress-nginx/ingress-nginx --create-namespace --namespace deepops-ingress
+	helm upgrade --install --wait "${app_name}" "${helm_arguments[@]}" ingress-nginx/ingress-nginx --create-namespace --namespace deepops-ingress
```
supertetelman, line 58: This is required for a section in the test that was previously silently failing due to an existing ingress install that was not properly deleted. This is also the mechanism we have been moving all our other helm installs to use.

workloads/jenkins/scripts/test-dcgm-metrics.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+# Verify that the DCGM-exporter is returning all metrics configured in the Grafana dashboard, for each node
+set -ex
+source workloads/jenkins/scripts/jenkins-common.sh
+
+# Ensure working directory is root
+cd "${ROOT_DIR}"
+
+
```
supertetelman, line 9: Remove extra line

workloads/jenkins/scripts/test-dcgm-metrics.sh
```diff
@@ -0,0 +1,51 @@
...
+DCGM_EXPORTER_PORT=9400
+
+# Run DCGM metric checks against all nodes in the group passed in (kube-node or slurm-node)
+group="${1}"
+
```
supertetelman, line 14: Remove extra line

PR #972, opened 2021-05-28
Fix installation of EPEL for Ansible 2.9.13+

ajdecon:
Ansible 2.9.13 and greater fixed CVE-2020-14365, where the yum/dnf API
was not correctly verifying GPG keys.

https://github.com/ansible/ansible/pull/71537

This behavior allowed us to install EPEL directly from the package URL
without first verifying its key, but once we upgrade Ansible this
breaks!

We will see this in DeepOps by default now that we've bumped our
default Ansible version above 2.9.13.

https://github.com/NVIDIA/deepops/commit/db93f62b10fcfe4223c8999313e1997172f388bf

This PR adds a step to explicitly import the Fedora GPG keys that sign
EPEL packages before installing the epel-release package.


supertetelman:
Seems like a lot of copy/ paste, but looks good.

supertetelman (approved):

ajdecon:
Yeah, the copy/paste is somewhat difficult to avoid if we want the roles to be able to "stand alone", which is what we've done so far.

We should probably eventually refactor all this into a single epel role and add role dependencies.

PR #984, opened 2021-06-22
Use full paths for files in setup.sh

supertetelman:
In the last release we introduced a minor issue with the refactoring of the setup script.

When `setup.sh` was executed from somewhere other than the deepops directory it expected config.example and roles to be in the local directory. This resulted in failed deployments.

This fixes the issue by using full paths.


ajdecon (approved):
LGTM. Approving but will let @supertetelman merge in case he wants to make other changes.

PR #989, opened 2021-06-22
Fix PAM Slurm adopt/login

dholt:


avolkov1 (approved):
I pulled this down and ran the slurm playbook. The pam sshd settings file `/etc/pam.d/sshd` looks good. Non-localusers without active slurm sessions cannot ssh to the compute node as expected.

Thank you!

ajdecon (approved):
LGTM

Issue #987, opened 2021-06-22
EasyBuild setup fails on Red Hat nodes owing to lack of Bash profile


verdurin:
The default config is to use the `csh` profile on Red Hat hosts (not sure why?) and I found that setting up EasyBuild failed with the following error:
```
TASK [easy-build-packages : install the selected easybuild modules] ********************************************************************************************************************************************************************************************************
failed: <controller>] (item=/apps/easybuild_files/000_binutils-2.32.eb) => changed=true
  ansible_loop_var: item
  cmd: |-
    . /etc/profile.d/z00_lmod.sh &&
    . /etc/profile.d/z01_eb.sh   &&
    eb --parallel=16 "/apps/easybuild_files/000_binutils-2.32.eb" --robot
  delta: '0:00:00.016574'
  end: '2021-06-22 12:58:01.968135'
  item: /apps/easybuild_files/000_binutils-2.32.eb
  msg: non-zero return code
  rc: 1
  start: '2021-06-22 12:58:01.951561'
  stderr: '/bin/bash: line 1: /etc/profile.d/z01_eb.sh: No such file or directory'
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
```

The fix should be just to allow the file `/etc/profile.d/z01_eb.sh` on Red Hat nodes.

PR #993, opened 2021-06-30
EasyBuild: Enable both shells on all distros

ajdecon:
Our current EasyBuild role enables the bash profile on Debian-based systems, and the csh profile on RedHat-based systems, but not vice-versa. There's no clear reason for this limitation, so let's remove it.

Addresses #987 .

### DIFF
--- a/roles/easy-build/tasks/main.yml
+++ b/roles/easy-build/tasks/main.yml
@@ -106,7 +106,7 @@
     owner: root
     group: root
     mode: 0777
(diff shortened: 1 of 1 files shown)


ajdecon:
@issamsaid : This behavior was included in your PR https://github.com/NVIDIA/deepops/commit/54b2636fde779ac07512241984682b55a6a95e68, can you clarify if this was intended and why?

michael-balint (approved):

Issue #969, opened 2021-05-28
files in  ENROOT_CACHE_PATH not group readable


mathrock74:
the files created by enroot in ENROOT_CACHE_PATH are not readable by the group and therefore can't be used by other users
I solved this for me with adding 

cache_path="$(sudo -u "$SLURM_JOB_USER" sh -c 'echo "/var/lib/enroot-cache/group-$(id -g)"')"
find "$cache_path" -type f -exec chmod 0660 {} \\;

to  /etc/slurm/epilog.d/50-lastuserjob-all-enroot-dirs

ajdecon:
This is a nice performance optimization to allow users to share cache, but I don't think this is something we want to enable by default. Different users on the cluster may have access to different containers, and we want to maintain stricter permissions by default.

mathrock74:
Actually this is not an optimization for me, it makes the system work. When I leave it at the default and user foo runs a container, then user bar runs the same container he get's "permission denied" from pyxis:

foo@deepops-login:~$ srun -w deepops-node3 --container-image=centos grep PRETTY /etc/os-release
pyxis: importing docker image ...
PRETTY_NAME="CentOS Linux 8"

bar@deepops-login:~$ srun -w deepops-node3 --container-image=centos grep PRETTY /etc/os-release 
pyxis: importing docker image ...
slurmstepd: error: pyxis: child 20158 failed with error code: 1
slurmstepd: error: pyxis: failed to import docker image
slurmstepd: error: pyxis: printing contents of log file ...
slurmstepd: error: pyxis:     [INFO] Querying registry for permission grant
slurmstepd: error: pyxis:     [INFO] Authenticating with user: <anonymous>
slurmstepd: error: pyxis:     [INFO] Authentication succeeded
slurmstepd: error: pyxis:     [INFO] Fetching image manifest list
slurmstepd: error: pyxis:     [INFO] Fetching image manifest
slurmstepd: error: pyxis:     [INFO] Found all layers in cache
slurmstepd: error: pyxis:     [INFO] Extracting image layers...
slurmstepd: error: pyxis:     tar (child): /var/lib/enroot-cache/group-100/7a0437f04f83f084b7ed68ad9c4a4947e12fc4e1b006b38129bac89114ec3621: Cannot open: Permission denied
slurmstepd: error: pyxis:     tar (child): Error is not recoverable: exiting now
slurmstepd: error: pyxis:     tar: Child returned status 2
[...]

PR #997, opened 2021-06-30
Enroot: adjust cache directory to be per-user

ajdecon:
## Summary

Previously, the enroot configuration directory on a node was allocated on a per-group basis according to the user's primary group. However, the permissions are not set to be group readable as per #969.

From a permissions standpoint, we probably don't want the enroot cache directory to be group-readable by default. Users may have access to different sets of private container images, so we should default to making these permissions restrictive.

This PR therefore switches Enroot to use a per-user cache directory by default.

## Test plan

Set up a test Slurm cluster with this configuration, and created two test users (`user1` and `user2`) with the same primary group.

Started an enroot job as `user1`:

```
user1@virtual-login01:~$ id
uid=20001(user1) gid=100(users) groups=100(users),27(sudo)
user1@virtual-login01:~$ srun -n1 --container-image="centos" --pty bash
pyxis: importing docker image ...
root@virtual-gpu01:/#
```

Then started a simultaneous job on the same host as `user2` and confirmed this works (addressing the error condition in #969):

```
user2@virtual-login01:~$ id
uid=20002(user2) gid=100(users) groups=100(users),27(sudo)
user2@virtual-login01:~$ srun -n1 -w virtual-gpu01 --container-image="centos" --pty bash
pyxis: importing docker image ...
root@virtual-gpu01:/#
```

Checked the cache directories on `virtual-gpu01` and verified two cache directories exist:

```
vagrant@virtual-gpu01:~$ ls /var/lib/enroot-cache/
user-20001  user-20002
```

### DIFF
--- a/config.example/group_vars/all.yml
+++ b/config.example/group_vars/all.yml
@@ -173,7 +173,7 @@ docker_iptables_enabled: true
 #
 # See: https://github.com/NVIDIA/pyxis/wiki/Setup#enroot-configuration-example
 enroot_runtime_path: '/run/enroot/user-$(id -u)'
-enroot_cache_path: '/var/lib/enroot-cache/group-$(id -g)'
+enroot_cache_path: '/var/lib/enroot-cache/user-$(id -u)'
 enroot_data_path: '/tmp/enroot-data/user-$(id -u)'
 enroot_config: |
   ENROOT_CONFIG_PATH ${HOME}/.config/enroot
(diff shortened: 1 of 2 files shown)


dholt:
I'm a little conflicted on this. The point was to re-use cached containers among many users. If users want their containers to remain private, can't they change the cache directory themselves to point to another location?

ajdecon:
@dholt : I went back and forth on this to be honest, but the enroot configuration isn't well-understood by admins in most cases, let alone by regular users. I don't think expecting users to set their own private cache dirs is reasonable (unless we somehow expose this more prominently).

I generally favor "failing closed" rather than "failing open" which is why I made this change, but there's definitely a trade-off.

The cache should still be useful for users who use the same containers in subsequent jobs.

dholt:
Fair enough.

Probably also need to change permissions here:

https://github.com/NVIDIA/deepops/blob/f24ea7146d61dadb81942d2d7a7298177fbe6b35/roles/slurm/templates/etc/slurm/prolog.d/50-all-enroot-dirs#L18

dholt (approved):

Issue #744, opened 2020-10-30
Grafana not showing System Power Draw on DGX-A100


itzsimpl:
In Grafana dashboard on the GPU Nodes view there are two panels; System Power Draw and System Total Power Draw (bottom right corner). On non-DGX systems the two panels work as they should, but on a DGX-A100 both display no data (N/A).

supertetelman:
Was this on the `master` branch or on the `20.10` release? Master has some updates to the monitoring stack, so I'm wondering if the newest DCGM-exporter version may have caused/fixed this.

itzsimpl:
This is on the 20.10 release, as far as I was able to discover it has something to do with `lmsensors`. On a non-DGX system it returns for example:
```
power_meter-acpi-0
Adapter: ACPI interface
power1:      357.00 W  (interval =   1.00 s)

coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +25.0°C  (high = +81.0°C, crit = +91.0°C)
Core 0:        +24.0°C  (high = +81.0°C, crit = +91.0°C)
Core 1:        +22.0°C  (high = +81.0°C, crit = +91.0°C)
Core 2:        +23.0°C  (high = +81.0°C, crit = +91.0°C)
Core 3:        +22.0°C  (high = +81.0°C, crit = +91.0°C)
Core 4:        +23.0°C  (high = +81.0°C, crit = +91.0°C)
Core 5:        +21.0°C  (high = +81.0°C, crit = +91.0°C)
Core 6:        +24.0°C  (high = +81.0°C, crit = +91.0°C)
Core 7:        +22.0°C  (high = +81.0°C, crit = +91.0°C)
Core 8:        +23.0°C  (high = +81.0°C, crit = +91.0°C)
Core 9:        +24.0°C  (high = +81.0°C, crit = +91.0°C)
Core 10:       +21.0°C  (high = +81.0°C, crit = +91.0°C)
Core 11:       +21.0°C  (high = +81.0°C, crit = +91.0°C)
Core 12:       +23.0°C  (high = +81.0°C, crit = +91.0°C)
Core 13:       +22.0°C  (high = +81.0°C, crit = +91.0°C)
Core 14:       +22.0°C  (high = +81.0°C, crit = +91.0°C)
Core 15:       +23.0°C  (high = +81.0°C, crit = +91.0°C)

coretemp-isa-0001
Adapter: ISA adapter
[...]

PR #999, opened 2021-07-01
Grafana: redesigned dashboards, added System Power Draw reporting for DGX-A100

itzsimpl:
I updated the slurm-exporter to the latest version (https://github.com/vpenso/prometheus-slurm-exporter) which can provide also information about the GPUs (the docker image is build on the fly). Note, that unfortunately this may not work with newer versions of slurm (see https://github.com/vpenso/prometheus-slurm-exporter/issues/38). I also slightly redesigned the Grafana SLURM and GPU node dashboards. And I have found a solution for System Power Draw reporting to solve #744. It is based on ipmi-exporter (https://github.com/soundcloud/ipmi_exporter) (the docker image is build on the fly).


dholt:
Thanks for the PR!

I've seen substantial hangs using an IPMI exporter before, have you tried this over a few weeks on more than a single node?

itzsimpl:
I've been using this IPMI exporter for about 6 months. It is just on a tiny 3 node SLURM setup with one login node, one DGX-A100, and one Intel based system with 4 GPUs. However, until now I have not experienced any hangs.

dholt:
Did you intend to include the Enroot-specific prolog/epilog bits? If not, do you mind moving those to their own PR?

itzsimpl:
I apologise; these were not intended to be included. Frankly speaking I'm unsure how they got included in the first place; they are all from my previous (separate) PRs, which have already been merged into the main branch (and possibly modified afterwards). Any pointers on how to remove them from the PR? Will deleting them using GitHub UI remove the files from the entire repo, or just mark as unchanged with respect to the main repo?

itzsimpl:
I've reverted the unintended commits.

itzsimpl:
Is there anything else that needs to be corrected?

dholt:
I still have a couple concerns, 1) that this won't work with the default newer version of Slurm in DeepOps, and 2) not all nodes have IPMI, does this still work on non-ipmi machines? Unfortunately I haven't had time to test this one yet

supertetelman:
In the future, we should look at integrating these dashboards into the official DCGM repo here: https://github.com/NVIDIA/dcgm-exporter/tree/main/grafana

But, for now it looks like it will be easier for us to just keep it separate.

PR #1002, opened 2021-07-02
Adding proxy support for downloading of hwloc, pmix, nhc and slurm.

mkunin-nvidia:
Those task didn't use proxy configuration for downloads.


ajdecon (approved):
LGTM

PR #1009, opened 2021-07-04
[yaml] Fix double 'when' on nvidia-hpc-sdk//main.yml

elgalu:
> roles/nvidia-hpc-sdk/tasks/main.yml, line 54, column 3, found a duplicate dict key (when). Using last defined value only.

Signed-off-by: Leo Gallucci <elgalu3@gmail.com>


ajdecon (approved):
LGTM

PR #1023, opened 2021-08-16
Eric/work

eanderson4:


ajdecon:
This PR looks like it may have been opened accidentally, and in any case includes a bunch of environment-specific stuff in `config/` that shouldn't be part of the upstream repo.

Can you comment on what you're trying to achieve here?

eanderson4:
Yes this was in error, can we please delete this

PR #1035, opened 2021-09-15
Create stale.yml

dholt:
This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,27 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'Stale issue message'
```
ajdecon, line 24: Do we need to fill in more specific messages in this configuration?

dholt: Good catch :) I just clicked the button Github offered and it created the PR

ajdecon (approved):
Looks reasonable!

PR #1036, opened 2021-09-15
Fix submodule update

dholt:
Ensure update of kubespray submodule only happens once


playbooks/k8s-cluster.yml
```diff
@@ -163,7 +164,7 @@
         mode: pull
         src: "/usr/local/bin/kubectl"
         dest: "{{ artifacts_dir }}/kubectl"
-      run_once: yes
+      run_once: true
```
dholt, line 167: Just modifying this to conform to the rest of the file

ajdecon (approved):
LGTM

Issue #1039, opened 2021-10-01
Error `grep: Invalid range end` in `scripts/deepops/doctl.sh`


murata-yu:
Hi, all

I built a k8s cluster and a slurm cluster and then ran  `scripts/deepops/doctl.sh` to change the cluster configuration.

However, I got the following error and could not execute it.

```bash
$ bash scripts/deepops/doctl.sh <node-name> state

# grep: Invalid range end
```
I think the regex used in grep is wrong.

```diff
diff --git a/scripts/deepops/doctl.sh b/scripts/deepops/doctl.sh
index 8b9b37af..d0c1d0d7 100755
--- a/scripts/deepops/doctl.sh
+++ b/scripts/deepops/doctl.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash


-nodes=$(scontrol show node "${1}" | grep -oe "^NodeName=[a-Z,0-9,-]*" | cut -d= -f2)
+nodes=$(scontrol show node "${1}" | grep -oe "^NodeName=[a-z,A-Z,0-9,-]*" | cut -d= -f2)

 shift
```

Thanks.

#### OS
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.5 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic

#### deepops
deepops=21.06

#### grep
```bash
$ grep --version
grep (GNU grep) 3.1
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

ajdecon:
Closed by #1040

PR #1040, opened 2021-10-01
fixed regex (#1039)

murata-yu:
I fixed grep error in #1039


ajdecon (approved):
LGTM

ajdecon:
Merged. Thank you for the fix!

PR #1042, opened 2021-10-01
Enhanced nccl tests slurm validation playbook.

avolkov1:
Ability to specify any base container and dynamically compile a new enroot container with NCCL tests. Then runs the NCCL tests using the compiled enroot container.

### DIFF
--- a/playbooks/slurm-cluster/slurm-validation.yml
+++ b/playbooks/slurm-cluster/slurm-validation.yml
@@ -1,6 +1,7 @@
 ---
 # Playbook designed to run a NCCL test across nodes in a cluster of DGXs
-# Example to run on NSL-B with two nodes:
+# Example to run on two nodes with no UCX, no HCOLL, and enp set for
+# out-of-band NCCL init:
 #   ansible-playbook -l slurm-cluster playbooks/slurm-cluster/slurm-validation.yml \
 #     -e '{num_nodes: 2}' \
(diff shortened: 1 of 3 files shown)


playbooks/slurm-cluster/slurm-validation.yml
```diff
@@ -1,33 +1,212 @@
 ---
-# Playbook designed to run a NCCL test across all nodes in a cluster of DGX-1s
+# Playbook designed to run a NCCL test across nodes in a cluster of DGXs
+# Example to run on NSL-B with two nodes:
```
supertetelman, line 3: We shouldn't have references to internal clusters if it can be avoided.

avolkov1: I can reword to:
```
# Example to run on two nodes with enp set for out-of-band NCCL init, no UCX and no HCOLL:
```

Or please suggest how to reword. This example will work in clusters without RDMA (no RoCE or infiniband). I want to document in the comments how to run on such clusters. Sometimes without the RDMA setup it is useful to test such cases.

avolkov1: Fixed in commit af08cba.

ajdecon (approved):
LGTM. Two minor nit-picks mentioned inline but I don't think those are worth blocking this.

Issue #890, opened 2021-02-25
Integrate k8s/deploy_monitoring.sh tooling with GPU Operator


supertetelman:
Since the initial creation of the `deploy_monitoring.sh` script, dcgm-exporter has been made available through a [helm deployment(https://github.com/NVIDIA/gpu-monitoring-tools#quickstart-on-kubernetes). We should look into replacing the existing script with another script/playbook that leverages this.

In addition to that, the GPU Operator now installs the dcgm-exporter by default. I believe that deployment classhes with our monitoring setup and the prometheus integration does not work if the operator is used for the install. As part of this issue, we should update these to not clash or better yet, tie prometheus into a operator instaled dcgm-export.

supertetelman:
We are now using the helm deployment.

We still need to make updates to integrate prometheus/Grafana with he Operator.

supertetelman:
Completed.

PR #1043, opened 2021-10-06
Kubespray v2.18.0 and containerd runtime

dholt:
Upgrades Kubespray to v2.18.0 and sets containerd as the default k8s runtime instead of docker.

Using containerd also means we're now using the GPU operator by default


playbooks/k8s-cluster.yml
```diff
@@ -133,7 +153,9 @@
 - include: container/nvidia-docker.yml
   tags:
     - nvidia
-  when: deepops_gpu_operator_enabled | default('false') | bool == false
+  when:
+    - deepops_gpu_operator_enabled | default('false') | bool == false
```
supertetelman, line 157: Can you change this to use the same "is defined and == false" syntax?

supertetelman:
Before merging this and making GPU Operator the default we should really fix monitoring: https://github.com/NVIDIA/deepops/issues/890

I have an open PR to bump the version and fix DGX support: https://github.com/NVIDIA/deepops/issues/1069

playbooks/k8s-cluster.yml
```diff
@@ -1,6 +1,26 @@
...
 # Kubernetes Cluster Playbook
 
+# Set facts depending on container runtime
+# Use GPU operator when container runtime is not docker
+# etcd_deployment_type must be `host` when container_manager is not `docker`
+- hosts: all
+  tasks:
+    - name: Set facts when not using docker container runtime (default)
+      set_fact:
+        deepops_gpu_operator_enabled: true
```
supertetelman, line 11: Why are we enforcing GPU Operator with containerd?

roles/nvidia-gpu-operator/defaults/main.yml
```diff
@@ -18,7 +18,7 @@ gpu_operator_namespace: "gpu-operator-resources"
 gpu_operator_grid_config_dir: "{{ deepops_dir }}/gpu_operator"
 
 # Defaults from https://github.com/NVIDIA/gpu-operator/blob/master/deployments/gpu-operator/values.yaml
-gpu_operator_default_runtime: "docker"
+gpu_operator_default_runtime: "containerd"
```
supertetelman, line 21: I'm removing this in a new PR

playbooks/k8s-cluster.yml
dholt: I don't recall exactly, but I thought it was required

supertetelman:
Before merging this through and essentially making GPU Operator the default path, I propose we get GPU Operator/ DGX support working (https://github.com/NVIDIA/deepops/pull/1074) get DCGM/Prometheus support integrated (https://github.com/NVIDIA/deepops/issues/890)

Until then, could we just check this PR, but keep the default as Docker?

supertetelman:
DGX Support with GPU Operator is complete. GPU Operator integration into our monitoring stack is complete. Just need to merge https://github.com/NVIDIA/deepops/pull/1074 with the content of this PR.

ajdecon:
DeepOps is currently pinned to Kubespray v2.17.1, which is newer than this PR is using.

Additionally, we're seeing issues in DGX OS 5.1.x (#1110) up to Kubespray v2.17.1.

So when we're ready for this PR to move forward, we should bump the Kubespray version to v2.18.0 (or whatever is latest).

supertetelman (approved):
LGTM.

I'd like to add some additional documentation here eventually and some additional testing to outline the different K8s configurations options and validate them.

I think we have

* Device Plugin + Docker (with nvidia-driver)
* Device Plugin + containerd (with nvidia-driver)
* GPU Operator + Docker (with driver container or with nvidia-driver)
* GPU Operator + containerd(with driver container or with nvidia-driver)

Is Device Plugin with Docker a supported path now? Do we have 6 total core configurations here?

PR #1048, opened 2021-10-20
Enable rootless docker-daemon per Slurm job.

avolkov1:
Fix the scenario of multiple Slurm jobs on a given node under the same user account. The incorrect behavior was that any subsequent rootless docker sessions for the same user, will use the same dockerd daemon and socket. Thus the resources i.e. GPU reservations, nproc, etc can get mixed up. This PR handles per-slurm-job rootless dockerd instances.

### DIFF
--- a/roles/docker-rootless/tasks/main.yml
+++ b/roles/docker-rootless/tasks/main.yml
@@ -67,10 +67,10 @@
     executable: /bin/bash
     creates: '{{ rootlessdocker_install_dir }}/config/install_success'
 
-- name: Clean up rootless docker set script
-  file:
-    path: /tmp/rootless_docker_setup.sh
-    state: absent
+#- name: Clean up rootless docker set script
+#  file:
+#    path: /tmp/rootless_docker_setup.sh
+#    state: absent
 
 - name: Add the br_netfilter module
   modprobe:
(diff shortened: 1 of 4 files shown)


avolkov1:
With latest Ubuntu 20.04 rootless docker does not terminate nicely under Slurm when a Slurm job or session ends. I think it's because of latest systemd and cgroups changes, as this worked fine under Ubuntu 18. I tried using systemd legacy mode instead of hybrid and that did not help.

I think it might work properly again, once cgroups v2 is supported by Slurm. The next Ubuntu LTS (22.04 LTS) will use cgroup v2 by default I believe.

avolkov1:
> This PR is stale because it has been open for 180 days with no activity. Please update the PR or it will be closed in 7 days.

Keep open.

avolkov1:
No, this has not been fixed. I haven't been able to figure out the root cause of this. It worked fine on Ubuntu 18 (works fine in Ubuntu 18 still, but that's very old now) and early versions of Ubuntu 20, until some updates were made to systemd (I'm suspecting). But I don't know exactly what updates or changes happened such that Slurm is unable to kill rootless docker processes properly when a job ends. I believe it has something to do with systemd, but I don't know for sure.

Manually calling "stop_rootless_docker.sh" prior to the job ending will typically kill rootless docker and a job will end normally. But I couldn't figure out a way to incorporate that into an epilogue. In the past, I didn't need to do this. When a Slurm job ended it was able to kill all user processes cleanly.

Maybe with Ubuntu 22 this will be fixed. On other OS's such as RHEL I believe this works fine.

PR #1051, opened 2021-11-02
Default to testing in Ubuntu 20.04

ajdecon:
Summary
-------

Propose that we switch our default for DeepOps testing from Ubuntu 18.04 to Ubuntu 20.04. This is largely to align with DGX OS 5, and to test on the platform that will be most likely to deploy on future systems.

- Jenkinsfile for PR testing to use Ubuntu 20.04 instead of 18.04
- Default to using Ubuntu 20.04 in the virtual cluster

Test plan
---------

Passing tests in Jenkins

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -8,7 +8,7 @@
 #                         or: curl -sL git.io/deepops | bash -s -- 19.07
 
 # Configuration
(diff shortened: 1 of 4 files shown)


supertetelman (approved):
LGTM

PR #1058, opened 2021-11-12
Improve documentation & playbook for DGX firmware upgrade, include paramiko in setup.sh

supertetelman:
This is an overall improvement to the documentation flow for updating DGX firmware. In updating these docs I went ahead and made some of the playbooks a bit more robust and improved the usability a bit.

* Install paramiko via pip in setup script
* Improve overall documentation flow with better examples and more details for diagnostics/firmware updates
* Modify playbook to not throw errors if log files/diagnostics were not  generated
* Introduced flag for backing up or skipping inactive firmware components

### DIFF
--- a/docs/deepops/dgx-diagnostic-firmware.md
+++ b/docs/deepops/dgx-diagnostic-firmware.md
@@ -114,7 +114,7 @@ This will do the following:
 * Transfer all logs to the provisioning node
 * List nodes that require a manual power cycle.
 
-For large DGX clusters, it is recommended to first perform a single manual firmware update and verify that node before using any automation cluster-wide.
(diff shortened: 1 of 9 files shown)


ajdecon (changes requested):
Walked through this process and tested on a DGX A100 system.

The actual Ansible execution went fine, and I was able to successfully execute both diagnostic and update runs.

Most of the inline comments are suggested additions or clarifications to the docs.

The only blocking request here is in the setup.sh, where `paramiko` is misspelled. 😉

ajdecon (approved):
Approving to merge and unblock dependencies, will add some extra changes in a later PR

PR #1071, opened 2021-12-09
Change default Interpreter in Ansible system default instead of Python3

supertetelman:
With this commit the Testing jobs on CentOS 7 began failing: https://github.com/NVIDIA/deepops/commit/2df93228dc0d9048b37759b290480fbf9e384e52

I am changing the reference to python3 to simply be default python. Hopefully this executes the correct Python version on each platform.


supertetelman:
Pulling some of Doug's bugfixes into this PR because some tests kept failing without them.

dholt (approved):

PR #1084, opened 2022-01-04
Adjust Github stale workflow

ajdecon:
- Increase timeout for PRs to 180 days
- Explicitly specify timeout for issues as 60 days
- Add support for a `no-stale` label to mark PRs and issues as exempt

Note: docs for stale action found here: https://github.com/actions/stale/blob/main/README.md

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -22,7 +22,7 @@ jobs:
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'This issue is stale because it has been open for 60 days with no activity. Please update the issue or it will be closed in 7 days.'
-        stale-pr-message: 'This PR is stale because it has been open for 60 days with no activity. Please update the PR or it will be closed in 7 days.'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -25,3 +25,7 @@ jobs:
         stale-pr-message: 'This PR is stale because it has been open for 60 days with no activity. Please update the PR or it will be closed in 7 days.'
```
dholt, line 25: This should be updated to say "180" for PRs based on the config below

dholt (commented):
Doesn't a `no-stale` tag defeat the purpose? :)

.github/workflows/stale.yml
ajdecon: Good call, fixed

ajdecon:
>Doesn't a no-stale tag defeat the purpose? :)

I mean, sort of. But AFAICT, only the repository admins can set labels, so we can manage which PRs are exempt as the project team.

I basically want us to have a way to tag issues and PRs that we know are long-term WIP, e.g. some of the issues that @supertetelman has been re-opening. It's annoying that they exist, but we shouldn't have to fight our own automation to track them. 😉

dholt (approved):

PR #1088, opened 2022-01-07
Add Molecule testing for Singularity, plus infra for more roles

ajdecon:
## Summary

- Fixes the ability to install Singularity on EL8 by enabling the
  necessary repos
- Move the pre-execution logic for the Singularity Galaxy role into a
  new DeepOps role, singularity_wrapper
- Modify the playbook to use singularity_wrapper
- Add Molecule tests to singularity_wrapper role
- Move Galaxy requirements for singularity wrapper into the
  singularity_wrapper role, and include this from
  `roles/requirements.yml`
- Add a Github Action configuration to execute this and other Molecule
  tests of DeepOps roles
- Add documentation to `docs/deepops/testing.md` to describe these tests
  and provide instructions on adding more

## Test plan

* Clean execution of the Github action for this role
* Clean execution of existing Jenkins tests

## Future work

This change adds a lot of infrastructure that should make it easier to
add testing to additional roles. Future changes should enable testing of
more of the roles in DeepOps, but I didn't want to get *too* carried
away in this one. :wink:

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -155,7 +155,7 @@ if command -v ansible-galaxy &> /dev/null ; then
     as_user ansible-galaxy role install -p "${roles_path}" --force -r "roles/requirements.yml" >/dev/null
 
     # Install any user-defined config requirements
-    if [ -d "${CONFIG_DIR}" ]; then
+    if [ -d "${CONFIG_DIR}" ] && [ -f "${CONFIG_DIR}/requirement.yml" ] ; then
         cd "${CONFIG_DIR}"
         as_user ansible-galaxy collection install -p "${collections_path}" --force -i -r "requirements.yml" >/dev/null
         as_user ansible-galaxy role install -p "${roles_path}" --force -i -r "requirements.yml" >/dev/null
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -7,10 +7,14 @@
 # Can be run standalone with: curl -sL git.io/deepops | bash
 #                         or: curl -sL git.io/deepops | bash -s -- 19.07
 
+# Determine current directory and root directory
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+ROOT_DIR="${SCRIPT_DIR}/.."
+
 # Configuration
 ANSIBLE_VERSION="${ANSIBLE_VERSION:-2.9.21}"     # Ansible version to install
 ANSIBLE_TOO_NEW="${ANSIBLE_TOO_NEW:-2.10.0}"    # Ansible version too new
-CONFIG_DIR="${CONFIG_DIR:-./config}"            # Default configuration directory location
+CONFIG_DIR="${CONFIG_DIR:-${ROOT_DIR}/config}"            # Default configuration directory location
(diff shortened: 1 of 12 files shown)


scripts/setup.sh
```diff
@@ -146,10 +146,25 @@ fi
 # Install Ansible Galaxy roles
 if command -v ansible-galaxy &> /dev/null ; then
     echo "Updating Ansible Galaxy roles..."
-    as_user ansible-galaxy collection install --force -r "${ROOT_DIR}/roles/requirements.yml" >/dev/null
-    as_user ansible-galaxy role install --force -r "${ROOT_DIR}/roles/requirements.yml" >/dev/null
-    as_user ansible-galaxy collection install --force -i -r "${ROOT_DIR}/config/requirements.yml" >/dev/null
-    as_user ansible-galaxy role install --force -i -r "${ROOT_DIR}/config/requirements.yml" >/dev/null
+    initial_dir="$(pwd)"
+    roles_path="${ROOT_DIR}/roles/galaxy"
+    collections_path="${ROOT_DIR}/collections"
```
dholt, line 151: Do we need to add this path to our default ansible config? I get a warning while running the setup script:

```
[WARNING]: The specified collections path '/home/dholt/work/deepops/collections' is not part of the configured Ansible collections paths '/home/dholt/.ansible/collections:/usr/share/ansible/collections'. The installed collection
won't be picked up in an Ansible run.
```

ajdecon: It's actually already in the config: https://github.com/NVIDIA/deepops/blob/master/ansible.cfg#L2

This is a bit of an ugly issue, and I'm interested in your feedback here.

1. Molecule expects any requirements file to exist within the role directory, so that it can install dependencies as-needed to test.

    So I added `requirements.yml` in the role dir, and specified the requirements file in `molecule/default/molecule.yml`:

    ```
    dependency:
        name: galaxy
        options:
        requirements-file: requirements.yml
    ```
    
    So far, so good.

2. I don't want to specify the same requirements more than once, and it turns out the `requirements.yml` file allows includes! So I rework this a bit so that I can include the dependencies in both our global file and the per-role file:
    
    **roles/singularity_wrapper/roles.yml:**
    ```
    - src: abims_sbr.singularity
      version: 3.7.1-1
    - src: gantsign.golang
      version: 2.4.0
    ```
    
    **roles/singularity_wrapper/requirements.yml:**
    ```
    roles:
    - include: roles.yml
    ```
    
    **roles/requirements.yml:**
    ```
    roles:
    - include: singularity_wrapper/roles.yml
    ```

3. However, when `ansible-galaxy` does includes, [it does them using relative paths at the level of its cwd, not the requirements file](https://github.com/ansible/ansible/issues/46385). Hence the changes to `setup.sh` to cd to the directory the requirements file is in.

4. But lastly... when we do that, the `ansible.cfg` isn't respected any more, because we changed the working directory. So we have to specify our role and collections paths manually, and we get that warning.

Honestly, it's a huge pain, and *almost* led to me abandoning the whole thing. But I like being able to get [individual test results for each role](https://github.com/NVIDIA/deepops/actions/runs/1668722135) 😉  and I think this will add a lot of value in testing when we touch multiple roles at once.

So... I dunno. Can you think of a better way to handle this?

ajdecon:
@dholt : This should be ready to test now! I simplified the requirements file handling for Github actions.

dholt (approved):

PR #1090, opened 2022-01-10
Add proxy config to standalone container registry

nuka137:
Current standalone container registry can not work on the proxy environment.
This PR enables us to specify the proxy configurations for standalone container registry.

### DIFF
--- a/roles/standalone-container-registry/defaults/main.yml
+++ b/roles/standalone-container-registry/defaults/main.yml
@@ -19,3 +19,6 @@ standalone_container_registry_cache_enable: false
 standalone_container_registry_cache_upstream: "https://registry-1.docker.io"
 # standalone_container_registry_cache_username: 
 # standalone_container_registry_cache_password:
+
+# standalone_container_registry_http_proxy:
+# standalone_container_registry_https_proxy:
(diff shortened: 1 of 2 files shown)


ajdecon:
@nuka137 : This looks reasonable to me!

Can you please sign off on your commit as detailed in our [CONTRIBUTING.md](https://github.com/NVIDIA/deepops/blob/master/CONTRIBUTING.md) doc? And then I can merge the PR.

nuka137:
@ajdecon 

Recommitted about previous patch.
Could you check it?

ajdecon (approved):

ajdecon (changes requested):
This actually appears to fail in the default case (without a proxy). If the variable isn't defined, Ansible won't run.

We use a different pattern in other tasks ([example](https://github.com/NVIDIA/deepops/blob/708e61d71a2bd2076eaf48661208cb479a50bc6f/roles/facts/tasks/main.yml#L5)) that looks like this:

```
- name: apt install pciutils
  apt: name=pciutils update_cache=yes
  when: ansible_os_family == 'Debian'
  environment: "{{ proxy_env if proxy_env is defined else {} }}"
```

And then define `proxy_env` if needed:

```
# proxy_env:
      # http_proxy: '{{ http_proxy }}'
      # https_proxy: '{{ https_proxy }}'
      # no_proxy: '{{ no_proxy }}'
```

Can you convert this to either use the same `proxy_env` variable, or a similar pattern that works when no proxy is present?

nuka137:
@ajdecon 

Thanks. I fixed it.

ajdecon (approved):
LGTM

PR #1096, opened 2022-01-24
Fixing rsyslog-server role.

mkunin-nvidia:
The role didn't create the directory for storing the client's logs.

In addition, I added an additional variable for the logs location: `rsyslog_log_file_path`.

### DIFF
--- a/docs/deepops/testing.md
+++ b/docs/deepops/testing.md
@@ -81,17 +81,18 @@ To add Molecule tests to a new role, the following procedure can be used.
 
 1. Ensure you have Docker installed in your development environment
 
-2. Install Ansible Molecule in your development environment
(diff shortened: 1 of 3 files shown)


ajdecon (approved):
@mkunin-nvidia : Overall this PR looks fine to me, and I'll approve. Can't hurt to create this explicitly.

However, I'm curious where you've seen this fail? I know that we see this role succeed consistently in Jenkins tests, and the required directory is created. (I think rsyslog does this automatically.)

Sample output from a successful Jenkins test of the existing role here: https://gist.github.com/ajdecon/3fcbc1315c7c695c5aab429793f4be2e

mkunin-nvidia:
@ajdecon I saw this in Zenseact. The folders inside are created but not the `/var/log/deepops-hosts`.

PR #1097, opened 2022-01-24
`docs/deepops/testing.md` fixes.

mkunin-nvidia:
1. The `molecule init` command has not a correct parameter. Instead of `--driver`, it should be `--driver-name`.

2. It is necessary to install  `community.docker` Ansible Galaxy collection. It is not mentioned in any place. However, without this collection, it is not possible to work with molecule.


ajdecon (approved):
Good catch on this! LGTM

PR #1098, opened 2022-01-26
Update roles/rsyslog-server/tasks/main.yml

mkunin-nvidia:
This role doesn't work in CentOS. CentOS doesn't have dedicated user for rsyslog service. This commit fixes this.


ajdecon (approved):
Tested on all supported platforms, LGTM

PR #1100, opened 2022-01-27
Add a basic playbook for installing Mellanox OFED

ajdecon:
Summary
-------------

This PR adds a basic playbook for installing the [NVIDIA Mellanox OFED distribution ](https://www.mellanox.com/products/infiniband-drivers/linux/mlnx_ofed), automating the [instructions for driver install from the website](https://community.mellanox.com/s/article/howto-install-mlnx-ofed-driver).

Test plan
------------

- [x] Added Ansible Molecule test and confirmed that these tests pass
- [ ] Test directly on a bare-metal host with an IB device present


playbooks/utilities/mofed.yml
```diff
@@ -0,0 +1,5 @@
+---
```
supertetelman, line 1: Is it necessary to add a few lines of documentation to the top of this playbook or do you think it is self-describing?

ajdecon: Documentation is always nice to have, I can add some.

dholt:
Because I took so long to review, the MOFED download links are stale (I just get a 404). It's unfortunate that older versions don't exist on the repo, but maybe there's a better way we can handle this?

roles/mofed/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+mofed_version: "5.4-3.1.0.0"
```
dholt, line 2: Needs to be updated, this version worked for me: "5.6-2.0.9.0"

dholt (approved):
Works on my machine!

PR #1111, opened 2022-02-15
Make it optional to use pam_slurm_adopt

ajdecon:
## Summary

We currently configure DeepOps Slurm clusters to use `pam_slurm_adopt.so` in the SSH authentication process. This module does two things:

- When the user has a job on the node, it adopts the SSH process into the job so it can be killed when the job ends
- When the user *doesn't* have a job on the node, it denies access

This behavior is exactly what you want on a production system, in order to prevent one user from interfering with another's workload. However, it often causes confusion for users new to Slurm, and doesn't always correspond to what you want on a development system which may be undergoing frequent change.

This PR adds two new role variables to the Slurm role:

- `slurm_restrict_node_access`: When enabled, denies SSH to users without a job on the node
- `slurm_enable_pam_slurm_adopt`: When enabled, sets up pam_slurm_adopt to adopt SSH processes into the job

By default, both of these are `true`. If `slurm_restrict_node_access` is set to false, SSH processes will still be adopted into the job for accounting and cleanup purposes, but otherwise allow access without a job on the node. If `slurm_enable_pam_slurm_adopt` is set to false, then `pam_slurm_adopt.so` will not be used at all.

## Test plan

- [x] Set `slurm_restrict_node_access: false` and create a new user who is not included in `slurm_allow_ssh_user`. Confirm that the user can SSH from the login node to the compute node.
- [x] Set `slurm_enable_pam_slurm_adopt: false`. Run a job, SSH in from another terminal, and confirm the SSH process isn't adopted.
- [x] Confirm regular DeepOps CI tests still work as expected.

### DIFF
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -39,8 +39,17 @@ slurm_ha_state_save_location: "/sw/slurm"
 ################################################################################
 # Login on compute
 ################################################################################
+# Restrict user SSH access to only allow users with a running job
+slurm_restrict_node_access: true
+
+# List users who should always be able to SSH, even without a running job
 slurm_allow_ssh_user: []
+
+# Enable a Slurm compute node to also function as a login node.
+# This is needed for the "single node Slurm cluster" configuration.
+# See docs/slurm-cluster/slurm-single-node.md
 slurm_login_on_compute: false
+
 
 ################################################################################
 # Optional installs                                                            #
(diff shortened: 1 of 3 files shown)


avolkov1 (approved):
The changes make sense and are easy to follow. I did not test behavior varying every combination 
 of "slurm_enable_pam_slurm_adopt" and "slurm_restrict_node_access", but the Ansible logic looks straightforward.

If you would like me to test this more extensively let me know. Logically this does what is expected, i.e. optionally restrict/permit ssh access in general and make "pam_slurm_adopt" optional.

avolkov1 (approved):

dholt (approved):

PR #1112, opened 2022-02-15
Update Jenkinsfile-nightly to reflect new Jenkins setup

supertetelman:
I modified Jenkins to have multiple jobs. This code change purely adds more debug to the common Jenkins scripting and reduces Jenkinsfile complexity to support this change.

The main benefit of this is that we will run more jobs in parallel in our test harness. 

Previously we had:
* 1 multi-node job that run Ubuntu 18, Ubuntu 20, CentOS 7, CentOS 8
* 1 single-node job that run Ubuntu 18, Ubuntu 20, CentOS 7, CentOS 8

Now we have:
* single-node Ubuntu 18
* single-node Ubuntu 20
* single-node CentOS 7
* single-node CentOS 8
* multi-node Ubuntu 18
* multi-node Ubuntu 20
* multi-node CentOS 7
* multi-node CentOS 8


ajdecon (approved):
Let's try it!

PR #1114, opened 2022-02-22
Fixing scripts/setup.sh.

mkunin-nvidia:
The  scripts/setup.sh crashes because of the  MarkupSafe new release. To fix the issue I added explicit MarkupSafe version 2.0.1.


ajdecon (approved):
LGTM, but opened #1115 to make sure we revisit this and can continue to follow new versions.

PR #1116, opened 2022-02-23
Temporarily disable CentOS 8 tests in Molecule

ajdecon:
Starting on Jan 31, the CentOS 8 mirror lists are no longer working. This
is causing our CentOS 8 tests in Molecule to fail because we can't
install packages.

While I've tested using the CentOS Vault instead of the canonical mirror
list, this isn't a great solution:

- It's a bit on the slow side, because we don't have a global network of
mirrors
- CentOS 8 is no longer supported, so we'd just be targeting an
increasingly old distribution

We're discussing how to address this in our automated tests. But for
now, we'll disable these tests so that our other Molecule tests will
run.

### DIFF
--- a/roles/cachefilesd/molecule/default/molecule.yml
+++ b/roles/cachefilesd/molecule/default/molecule.yml
@@ -34,16 +34,16 @@ platforms:
     - /tmp
     pre_build_image: true
     privileged: true
-  - name: cachefilesd-centos-8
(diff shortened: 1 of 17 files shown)


dholt (approved):

supertetelman:
I am also disabling the tests in Jenkins.

PR #1118, opened 2022-02-24
Migrate CentOS-8 molecule tests to Rocky

ajdecon:
Starting on Jan 31, the CentOS 8 mirror lists are no longer working.
This is causing our CentOS 8 tests in Molecule to fail because we can't
install packages.

While I've tested using the CentOS Vault instead of the canonical mirror
list, this isn't a great solution:

- It's a bit on the slow side, because we don't have a global network of mirrors
- CentOS 8 is no longer supported, so we'd just be targeting an increasingly old distribution

As a proposed solution, I've converted the molecule tests to use [Rocky
Linux](https://rockylinux.org/). This is mostly a drop-in replacement,
with some minor changes needed in the facts and nhc roles due to the
Rocky container being a little more minimal than the CentOS one.

### DIFF
--- a/roles/cachefilesd/molecule/default/molecule.yml
+++ b/roles/cachefilesd/molecule/default/molecule.yml
@@ -34,16 +34,16 @@ platforms:
     - /tmp
     pre_build_image: true
     privileged: true
-#  - name: cachefilesd-centos-8
-#    image: geerlingguy/docker-centos8-ansible
-#    volumes:
-#      - /sys/fs/cgroup:/sys/fs/cgroup:ro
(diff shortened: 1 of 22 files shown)


ajdecon:
Per the [CUDA 11.7 release notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#deprecated-features), Rocky Linux 8 is getting added to the official support list. Marking this PR ready to review.

ajdecon:
TODO - Switch Jenkinsfile to use Rocky as well

PR #1128, opened 2022-03-10
Fix some ordering in k8s-cluster.yml to install Helm properly and run all commands from kube-master[0]- fixing CentOS install

supertetelman:
Our Helm installs were doing a mix of running from localhost and/or kube-master[0]. This was causing issues in the nfs-client-provisioner because the CentOS kubespray installer was not properly installing kubectl on the kube-master nodes.

For now I am aligning everything to what we did in GPU Operator. In the future, it would make sense to use the now functional helm Ansible module and run things from localhost (the provisioning node) instead of the kube-master[0]. This would simply allow us to install less binaries on the management nodes, but beyond that  it is not a necessary change.

Also added the standard proxy commands to a few places where they were missing in helm installs.

Additionally I moved the block of code that runs helm/kubectl commands to be after the block where we actually install the proper kubectl/helm binaries. This was causing issues in the edge-cases on CentOS because of how different software was installed across Ubuntu/CentOS.

The automated testing already tests all the paths that this touches.


playbooks/k8s-cluster/nvidia-gpu-operator.yml
```diff
@@ -3,7 +3,7 @@
 - include: ../bootstrap/bootstrap-openshift.yml
 
 # GPU operator
-- hosts: kube-master[0]
```
supertetelman, line 6: The expectation is that Helm commands are run from the provisioning node. No need to install Helm and run it on the management systems.

roles/nvidia-gpu-operator/tasks/k8s.yml
```diff
@@ -11,10 +11,6 @@
   command: /usr/local/bin/helm repo update
   changed_when: false
 
-- name: Remove existing GPU Operator labels to accommodate potenial configuration changes
-  command: kubectl label node --all nvidia.com/gpu.deploy.container-toolkit-  nvidia.com/gpu.deploy.dcgm- nvidia.com/gpu.deploy.dcgm-exporter- nvidia.com/gpu.deploy.device-plugin- nvidia.com/gpu.deploy.driver- nvidia.com/gpu.deploy.gpu-feature-discovery- nvidia.com/gpu.deploy.node-status-exporter- nvidia.com/gpu.deploy.operator-validator-
```
supertetelman, line 15: I had originally added this line to work around what seemed like a bug in the GPU Operator, but in retrospect this line should not be necessary as all these values are dynamically configured/set by the GPU Operator at install time.

Removing this because it is not needed and the reliance on kubectl in a role was causing failures.

playbooks/k8s-cluster.yml
```diff
@@ -139,24 +139,6 @@
     - nvidia
   when: deepops_gpu_operator_enabled | default('false') | bool == false or gpu_operator_preinstalled_nvidia_software
 
-# Install k8s GPU feature discovery
```
supertetelman, line 142: This block was not changed, just moved.

playbooks/k8s-cluster/nfs-client-provisioner.yml
```diff
@@ -5,8 +5,6 @@
 # After settting up an NFS server, nfs software is installed on all nodes
 # Lastly, the nfs-client-provisioner is helm installed using the role
 #
-# TODO: This uses a local submodule because 'stable' is EOL and a new repo is not yet up
```
supertetelman, line 8: Doug fixed this in his last PR.

ajdecon (approved):

PR #1117, opened 2022-02-23
Update NVIDIA DGX role to match DGX Software Stack for Ubuntu

ajdecon:
## Summary

The [DGX Software Stack for Ubuntu](https://docs.nvidia.com/dgx/dgx-ubuntu-install-guide/index.html) is a supported mechanism to configure an Ubuntu 20.04 system to match NVIDIA DGX OS 5. This PR will update our `nvidia-dgx` role, which is aimed at the same task, to match the procedure laid out in the DGX Software Stack install guide.

Note: Because the DGX Software Stack only supports Ubuntu 20.04, this PR intentionally drops support for Ubuntu 18.04 from this role.

## Tasks

- [x] Update role to match instructions in install guide
- [x] Update documentation

## Test plan

- [x] Run virtual test to validate basic Ansible logic
- [x] Run on DGX-1
- [x] Run on DGX A100

Tests look good on DGX-1 and DGX A100. I don't have a DGX-2 handy to test, though.

### Virtual test

In order to make it easier to test our DGX role, this PR adds two booleans to the configuration:

- `dgx_override_install`: Perform the install even when we don't detect DGX hardware
- `dgx_override_product_type`: Specify a DGX product type for use in testing

```
$ ansible-playbook -i virtual/config/inventory -l virtual-gpu01 -e '{"dgx_override_install": true, "dgx_override_product_type": "DGXA100"}'  playbooks/nvidia-dgx/nvidia-dgx-stack.yml

PLAY [all] *********************************************************************************************************************************************************************************************************************************

TASK [Include NVIDIA DGX role] *************************************************************************************************************************************************************************************************************

TASK [nvidia_dgx_stack : Check for DGX] ****************************************************************************************************************************************************************************************************skipping: [virtual-gpu01]

TASK [nvidia_dgx_stack : Ubuntu tasks for DGX OS 5] ****************************************************************************************************************************************************************************************included: /home/adeconinck/src/github/deepops/roles/nvidia_dgx_stack/tasks/ubuntu.yml for virtual-gpu01

TASK [nvidia_dgx_stack : Check for Ubuntu version] *****************************************************************************************************************************************************************************************skipping: [virtual-gpu01]

TASK [nvidia_dgx_stack : include 20.04 specific variables for DGX OS 5] ********************************************************************************************************************************************************************ok: [virtual-gpu01]

TASK [nvidia_dgx_stack : remove ubuntu nvidia driver ppa if installed] *********************************************************************************************************************************************************************ok: [virtual-gpu01]

[...]

PR #1129, opened 2022-03-13
Automatically run ansible-lint on each role

ajdecon:
## Summary

This PR makes the following changes:

- Adds `scripts/deepops/ansible-lint-roles.sh` which runs `ansible-lint` on each role in the `roles/` directory
- Adds a Github action in `.github/workflows/ansible-lint-roles.yml` to run this automatically on pushes and PRs
- Adds documentation on these scripts to `docs/deepops/testing.md`
- Cleans up linting issues in most existing roles so that the script can pass

(This turned out to be less difficult than I thought! Most roles had a small number of issues, or one minor issue repeated many times. A lot of fixes were solvable with a clever regex or three.)

Note: `ansible-lint-roles.sh` includes the option of specifying an environment variable `ANSIBLE_LINT_EXCLUDE`, that will be used as a regex to skip linting some roles. This is currently used in the Github action to skip linting `nvidia-dgx`, `nvidia-dgx-firmware`, and `nvidia-gpu-tests`. The first role is pending major changes in #1117. The other two are complex enough, and difficult enough to test, that they should get extra testing when making the changes needed for linting to pass.

## Test plan

- [x] All roles should pass `ansible-lint`, unless specifically excluded
- [x] All existing Ansible Molecule tests should pass
- [x] All existing end-to-end Jenkins tests should pass

### DIFF
--- a/docs/deepops/testing.md
+++ b/docs/deepops/testing.md
@@ -1,11 +1,24 @@
 # DeepOps Testing, CI/CD, and Validation
 
+## Linting
+
+`ansible-lint` is automatically run for each role in the `roles/` directory using a [Github action](../../.github/workflows/ansible-lint-roles.yml).
+This action runs `ansible-lint` for each role, and provides both the full output and a list of roles that failed linting.
+If the Github action reports success, all roles should have passed linting.
+
+The linting process can also be executed manually in a checkout of the DeepOps repo,
+by running `./scripts/deepops/ansible-lint-roles.sh`.
+
+Note that the linting script can be configured to skip a subset of roles,
+by providing a regex of roles to skip in the envionment variable `ANSIBLE_LINT_EXCLUDE`.
+(For example, `ANSIBLE_LINT_EXCLUDE='role-1|role-2|role-3'`.)
+This can be useful for excluding specific roles that have known issues or are still in development.
 
 ## DeepOps end-to-end testing
 
-The DeepOps project leverages a private Jenkins server to run continuous integration tests. Testing is done using the [virtual](../../virtual) deployment mechanism. Several Vagrant VMs are created, the cluster is deployed, tests are executed, and then the VMs are destroyed. 
+The DeepOps project leverages a private Jenkins server to run continuous integration tests. Testing is done using the [virtual](../../virtual) deployment mechanism. Several Vagrant VMs are created, the cluster is deployed, tests are executed, and then the VMs are destroyed.
 
-The goal of the DeepOps CI is to prevent bugs from being introduced into the code base and to identify when changes in 3rd party platforms have occurred or impacted the DeepOps deployment mechanisms. In general, K8s and Slurm deployment issues are detected and resolved with urgency. Many components of DeepOps are 3rd party open source tools that may silently fail or suddenly change without notice. The team will make a best-effort to resolve these issues and include regression tests, however there may be times where a fix is unavailable. Historically, this has been an issue with Rook-Ceph and Kubeflow, and those GitHub communities are best equipped to help with resolutions. 
+The goal of the DeepOps CI is to prevent bugs from being introduced into the code base and to identify when changes in 3rd party platforms have occurred or impacted the DeepOps deployment mechanisms. In general, K8s and Slurm deployment issues are detected and resolved with urgency. Many components of DeepOps are 3rd party open source tools that may silently fail or suddenly change without notice. The team will make a best-effort to resolve these issues and include regression tests, however there may be times where a fix is unavailable. Historically, this has been an issue with Rook-Ceph and Kubeflow, and those GitHub communities are best equipped to help with resolutions.
 
 ### Testing Method
 
--- a/roles/autofs/tasks/main.yml
+++ b/roles/autofs/tasks/main.yml
@@ -22,7 +22,12 @@
     - configuration
 
 - name: ensure mountpoint exists
-  file: path={{ autofs_mount }} state=directory
+  file: 
+    path: "{{ autofs_mount }}"
+    state: directory
+    owner: "root"
+    group: "root"
+    mode: "0755"
   when: autofs_mount is defined
 
 - name: make sure autofs is running
--- a/roles/cachefilesd/tasks/main.yml
+++ b/roles/cachefilesd/tasks/main.yml
@@ -8,11 +8,17 @@
   template:
     src: cachefilesd.j2
     dest: /etc/default/cachefilesd
+    owner: "root"
+    group: "root"
+    mode: "0644"
 
 - name: configure
   template:
     src: cachefilesd_config.j2
     dest: /etc/cachefilesd.conf
+    owner: "root"
+    group: "root"
+    mode: "0644"
 
 # Service start not tested in molecule as we require a kernel module
 - name: start
--- a/roles/docker-rootless/tasks/main.yml
+++ b/roles/docker-rootless/tasks/main.yml
@@ -61,8 +61,7 @@
 
 - name: Execute setup rootless docker script
   become: true
-  shell:
-    /tmp/rootless_docker_setup.sh
+  command: /tmp/rootless_docker_setup.sh
   args:
     executable: /bin/bash
     creates: '{{ rootlessdocker_install_dir }}/config/install_success'
@@ -82,3 +81,6 @@
     path: /etc/modules
     create: yes
     line: "br_netfilter"
+    owner: "root"
+    group: "root"
+    mode: "0644"
--- a/roles/easy-build-packages/.ansible-lint
+++ b/roles/easy-build-packages/.ansible-lint
@@ -1,0 +1,4 @@
+warn_list:  # or 'skip_list' to silence them completely
+  - meta-no-info  # meta/main.yml should contain relevant info
+  - no-changed-when  # Commands should not change things if nothing needs doing
+  - role-name
--- a/roles/easy-build-packages/.ansible-lint
+++ b/roles/easy-build-packages/.ansible-lint
@@ -1,4 +1,4 @@
-warn_list:  # or 'skip_list' to silence them completely
+skip_list:
   - meta-no-info  # meta/main.yml should contain relevant info
   - no-changed-when  # Commands should not change things if nothing needs doing
   - role-name
--- a/roles/easy-build-packages/defaults/main.yml
+++ b/roles/easy-build-packages/defaults/main.yml
@@ -1,1 +1,2 @@
 sm_files_path: "{{ sm_prefix }}/easybuild_files"
+sm_files_repo_version: "4ef7ae6cc2284f69412a8db5e10dddd92024eeab"
--- a/roles/easy-build/tasks/main.yml
+++ b/roles/easy-build/tasks/main.yml
@@ -4,12 +4,12 @@
 ---
 - name: "install needed packages"
   become: yes
-  action: apt name=python3-setuptools state=latest
+  action: apt name=python3-setuptools state=present
   when: ansible_os_family == "Debian"
 
 - name: "install needed packages"
   become: yes
-  action: yum name=python3-setuptools state=latest
+  action: yum name=python3-setuptools state=present
   when: ansible_os_family == "RedHat"
 
 - name: "rm previous tmp folders"    
@@ -56,7 +56,7 @@
 
 - name: "get EasyBuild version"
   become: yes
-  shell: ls {{ eb_tmp_dir }}/software/EasyBuild/
+  command: ls {{ eb_tmp_dir }}/software/EasyBuild/
   args:
     executable: /bin/bash
   register: ebsw_version
--- a/roles/k8s-internal-container-registry/tasks/main.yml
+++ b/roles/k8s-internal-container-registry/tasks/main.yml
@@ -6,9 +6,11 @@
 
 - name: install stable helm repository
   command: "{{ helm_cmd }} repo add {{ container_registry_repo_name }} {{ container_registry_chart_location }}"
+  changed_when: false
 
 - name: update helm repos
   command: "{{ helm_cmd }} repo update"
+  changed_when: false
 
 - name: copy values.yaml into place
   template:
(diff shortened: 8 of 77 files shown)


supertetelman:
Wow, 76 files edited?

Is there any specific area you think this could be potentially risky that I should focus on with a review?

docs/deepops/testing.md
```diff
@@ -1,145 +1,158 @@
...
-Additional services such as Kubeflow, Open OnDemand, or Monitoring may have additional validation steps that are documented in the corresponding DeepOps READMEs and the official documentation.
-
-For workloads that can be used as post-deployment validation, see the example workloads for [k8s]( ../../workloads/examples/k8s/) and [slurm]( ../../workloads/examples/slurm).
+# DeepOps Testing, CI/CD, and Validation
+
+## Linting
+
+`ansible-lint` is automatically run for each role in the `roles/` directory using a [Github action](../../.github/workflows/ansible-lint-roles.yml).
+This action runs `ansible-lint` for each role, and provides both the full output and a list of roles that failed linting.
+If the Github action reports success, all roles should have passed linting.
```
supertetelman, line 7: Github -> GitHub

docs/deepops/testing.md
```diff
@@ -1,145 +1,158 @@
...
-The Slurm and Kubernetes deployment guides both document cluster verification steps. These should be run during the installation process to validate a GPU workload can be executed on the cluster.
-
-Additional services such as Kubeflow, Open OnDemand, or Monitoring may have additional validation steps that are documented in the corresponding DeepOps READMEs and the official documentation.
-
-For workloads that can be used as post-deployment validation, see the example workloads for [k8s]( ../../workloads/examples/k8s/) and [slurm]( ../../workloads/examples/slurm).
+# DeepOps Testing, CI/CD, and Validation
+
+## Linting
+
+`ansible-lint` is automatically run for each role in the `roles/` directory using a [Github action](../../.github/workflows/ansible-lint-roles.yml).
```
supertetelman, line 5: Github -> GitHub

roles/autofs/tasks/main.yml
```diff
@@ -22,7 +22,12 @@
     - configuration
 
 - name: ensure mountpoint exists
-  file: path={{ autofs_mount }} state=directory
+  file: 
+    path: "{{ autofs_mount }}"
+    state: directory
+    owner: "root"
+    group: "root"
+    mode: "0755"
```
supertetelman, line 30: Is this mode value just the default? It does not appear in the previous main.yml

roles/cachefilesd/tasks/main.yml
```diff
@@ -8,11 +8,17 @@
   template:
     src: cachefilesd.j2
     dest: /etc/default/cachefilesd
+    owner: "root"
+    group: "root"
+    mode: "0644"
```
supertetelman, line 13: Where did these values come from?

roles/docker-rootless/tasks/main.yml
```diff
@@ -61,8 +61,7 @@
 
 - name: Execute setup rootless docker script
   become: true
-  shell:
-    /tmp/rootless_docker_setup.sh
+  command: /tmp/rootless_docker_setup.sh
```
supertetelman, line 64: Has this been tested? Sometimes shell is needed for parsing output, although that does not appear to be the case here.

roles/easy-build-packages/defaults/main.yml
```diff
@@ -1 +1,2 @@
-sm_files_path: "{{ sm_prefix }}/easybuild_files"
\ No newline at end of file
+sm_files_path: "{{ sm_prefix }}/easybuild_files"
+sm_files_repo_version: "4ef7ae6cc2284f69412a8db5e10dddd92024eeab"
```
supertetelman, line 2: What's up with this repo_version?

roles/easy-build-packages/.ansible-lint
```diff
@@ -0,0 +1,4 @@
+skip_list:
```
supertetelman, line 1: Was this automatically added?

roles/easy-build/tasks/main.yml
```diff
@@ -56,7 +56,7 @@
 
 - name: "get EasyBuild version"
   become: yes
-  shell: ls {{ eb_tmp_dir }}/software/EasyBuild/
+  command: ls {{ eb_tmp_dir }}/software/EasyBuild/
```
supertetelman, line 59: Has this been tested?

roles/k8s-internal-container-registry/tasks/main.yml
```diff
@@ -6,9 +6,11 @@
 
 - name: install stable helm repository
   command: "{{ helm_cmd }} repo add {{ container_registry_repo_name }} {{ container_registry_chart_location }}"
+  changed_when: false
```
supertetelman, line 9: Why did you add this? This should be consistent with all the helm install roles.

roles/lmod/tasks/main.yml
```diff
@@ -73,46 +79,52 @@
...
+    src: "/usr/lib/x86_64-linux-gnu/lua/{{ item }}/posix_c.so"
+    dest: "/usr/lib/x86_64-linux-gnu/lua/{{ item }}/posix.so"
+    owner: "root"
+    group: "root"
+    state: "link"
+    force: true
   when: 
     - ansible_distribution == 'Ubuntu'
     - ansible_distribution_version == '18.04'
-  changed_when: lmod_bugfix_posix.rc != 0
```
supertetelman, line 91: Is it okay that we dropped this?

roles/nvidia-dgx-firmware/tasks/get-data.yml
```diff
@@ -3,6 +3,6 @@
 - name: Run NVSM human-readable health show
   shell: "osversion=`cat /etc/dgx-release | grep 'DGX_SWBUILD_VERSION'`; host=`hostname`; copper_mac=`ip addr | grep -A 1 enp6s0 | grep link | awk '{print $2}'`; bmc_ip=`sudo ipmitool lan print | grep -i 'IP Address              :' | awk '{print $4}'`; bmc_mac=`sudo ipmitool lan print | grep -i 'MAC Address' | awk '{print $4}'`;host_mac=`ifconfig {{ nv_mgmt_interface }} | grep 'ether' | awk '{print $2}'`;host_ip=`ifconfig {{ nv_mgmt_interface }} | grep 'inet ' | awk '{print $2}'`;data=\"${host_ip},${host_mac},${bmc_ip},${bmc_mac},${copper_mac},${host},${osversion}\";echo ${data}"
   register: command_output
-  - debug:
```
supertetelman, line 6: How was this not failing previously from indentation errors?

roles/nvidia-gpu-operator-node-prep/tasks/main.yml
```diff
@@ -7,7 +7,7 @@
   modprobe:
     name: nouveau
     state: absent
-  ignore_errors: true
+  failed_when: false
```
supertetelman, line 10: Any idea what the functional difference here is?

roles/nvidia-gpu-operator/tasks/k8s.yml
```diff
@@ -5,12 +5,16 @@
...
 - name: install gpu-operator helm repo
   command: /usr/local/bin/helm repo add nvidia "{{ gpu_operator_helm_repo }}"
+  changed_when: false
 
 - name: update helm repos
   command: /usr/local/bin/helm repo update
+  changed_when: false
 
 - name: Remove existing GPU Operator labels to accommodate potenial configuration changes
-  shell: kubectl label node --all nvidia.com/gpu.deploy.container-toolkit-  nvidia.com/gpu.deploy.dcgm- nvidia.com/gpu.deploy.dcgm-exporter- nvidia.com/gpu.deploy.device-plugin- nvidia.com/gpu.deploy.driver- nvidia.com/gpu.deploy.gpu-feature-discovery- nvidia.com/gpu.deploy.node-status-exporter- nvidia.com/gpu.deploy.operator-validator-
```
supertetelman, line 13: I just added this line and if I remember there was an edge-case error that required me to use shell here instead of command. Let me double check that this is passing the tests.

roles/nvidia-gpu-operator/tasks/nvaie.yml
```diff
@@ -5,7 +5,8 @@
     gpu_operator_enable_migmanager: false
 
 - name: Create namespace for GPU Operator resources
-  shell: kubectl create namespace {{ gpu_operator_namespace }} -o yaml --dry-run=client | kubectl apply -f -
+  shell: set -o pipefail && kubectl create namespace {{ gpu_operator_namespace }} -o yaml --dry-run=client | kubectl apply -f -
+  changed_when: false
```
supertetelman, line 9: This should be be changed_when: true, shouldn't it?

scripts/deepops/ansible-lint-roles.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
supertetelman, line 1: Needs a line or 2 of comments documenting the usage of the script.

supertetelman (dismissed):
Learned a few neat things from going through all this linting. I had 1 or 2 concerns for the changes on some of the less used/testable playbooks. I also question adding additional lines for "changed_when: False" for every single helm command, but I am assuming the official policy is to set that for any command or shell?

Otherwise LGTM and merge it in.

roles/nvidia-gpu-operator/tasks/k8s.yml
supertetelman: Looks like no.

roles/lmod/tasks/main.yml
ajdecon: Yes - this needed to be dropped because we aren't executing commands anymore, we are using the built-in file module. So there isn't even a `lmod_bugfix_posix.rc` variable to check anymore!

roles/nvidia-gpu-operator-node-prep/tasks/main.yml
ajdecon: AFAICT, the major difference is that the task isn't marked as `failed` in the final Ansible results summary.

ajdecon:
@supertetelman : Thanks for walking through this! I definitely want good feedback on this PR, because it touches so many files. As you can see most of the changes are pretty mechanical, but that doesn't mean they're all perfect.

My major goal with this PR is really to get automated linting in place so that we can catch possible problems in the future, especially in code paths that aren't normally executed. The cleanups I did on the individual roles were intended to get us to an initial "passing" state to make the automated tests useful for future PRs.

Re: high-risk areas, I think anything not exercised by the end-to-end tests may be risky. I *think* most of the changes I made are pretty obviously benign. But I'm also happy to break risky roles out into their own PR for more thorough testing, and just add them to `ANSIBLE_EXCLUDE_LINT` for this PR.

A few high-level comments based on the things your highlighted:

- `ansible-lint` flags tasks that can create new files without specifying a `mode` parameter, because Ansible doesn't actually have a default value for this. 😬 Instead, it runs in the environment of whatever the `ansible_user` is, and uses the environment's `umask` to determine permissions.
    The mode I selected in most cases (`0644` for files, `0755` for directories) is a good general default based on the default umask for the root user on most distros (`022`).

- The `changed_when: false` additions to commands are being added for task idempotence purposes. This is a little bit of a hack -- ideally, commands would only be executed when needed. But where a command is *always* execution, the best practice I can find is to suppress the `changed` message, because that execution is the expected behavior on a "clean" fun.

- The `failed_when: false` replacements for `ignore_errors: true` is the most problematic change here. Ideally we'd be examining the output of each task and reporting failure based on parsing that. This PR was intended to be a "quick" addition of linting, so I just suppressed the expected failures completely... However, if you think it's a better idea, we can undo those changes. The presence of `ignore_errors: true` appears to be a linting warning rather than an error, so the overall lint process should still pass.

roles/nvidia-gpu-operator/tasks/nvaie.yml
ajdecon: See my high-level comment.

roles/easy-build-packages/defaults/main.yml
ajdecon: Most recent commit in https://github.com/DeepOps/easyconfigs . `ansible-lint` doesn't like `HEAD` as a git checkout target.

roles/nvidia-gpu-operator-node-prep/tasks/main.yml
supertetelman: From what I can tell the difference is that ignore_errors makes the output red and failed_when doesn't.

roles/lmod/tasks/main.yml
supertetelman: Makes sense!

roles/easy-build-packages/defaults/main.yml
supertetelman: Okay, sounds good. I also prefer pointing to a specific hash.

roles/nvidia-mig-manager/tasks/main.yml
```diff
@@ -3,13 +3,16 @@
...
 - name: check for MIG capable devices
-  shell: nvidia-smi --query-gpu=mig.mode.current --format=csv,noheader | grep -v 'N/A'
+  shell: set -o pipefail && nvidia-smi --query-gpu=mig.mode.current --format=csv,noheader | grep -v 'N/A'
   register: has_mig
   failed_when: false
+  changed_when: false
+
 - name: check for NVIDIA MIG parted
-  shell: which nvidia-mig-parted
+  command: which nvidia-mig-parted
```
supertetelman, line 12: I think this will break.

roles/roce_backend/tasks/mofed-install.yaml
```diff
@@ -10,7 +11,7 @@
         path: /usr/bin/ofed_info
       register: ofed_installed
     - name: Check Mellanox OFED version
-      shell: /usr/bin/ofed_info -n
+      command: /usr/bin/ofed_info -n
```
supertetelman, line 14: I'm not sure about this one.

ajdecon:
@supertetelman : I think those are probably ok, but also those roles are tricky to test. So I've swapped those two cases back to `shell` and added `# noqa` comments.

scripts/deepops/ansible-lint-roles.sh
ajdecon: Done, see below.

supertetelman (approved):
Great work!

PR #1139, opened 2022-03-24
Add more tests to cover containerd/driver container k8s deployments

supertetelman:
We now have the following configurations to test.

This change covers these all in the nightlies and it will probably take around 4 hours end to end.

I tried to mix-and-match the actual tests being run across them so that we have some confidence without filling out a dense test matrics. Kubeflow is tested once on GPU Operator and once on device plugin.

I skipped the local-registry tests on containerd installs, but kept it in for docker installs. This may or may not work given the recent changes.

I made sure that the monitoring stack is tested with at least one configuration of device plugin, docker, containerd, and driver-container configurations.

### DIFF
--- a/workloads/jenkins/Jenkinsfile
+++ b/workloads/jenkins/Jenkinsfile
@@ -83,6 +83,11 @@ pipeline {
             timeout 180 bash -x ./workloads/jenkins/scripts/test-dashboard.sh
           '''
 
+          echo "Start new virtual environment pre-Slurm checks"
+          sh '''
+            bash -x ./workloads/jenkins/scripts/vagrant-startup.sh
+          '''
+
           echo "Set up Slurm"
           sh '''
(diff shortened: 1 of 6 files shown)


workloads/jenkins/Jenkinsfile
```diff
@@ -83,6 +83,11 @@ pipeline {
             timeout 180 bash -x ./workloads/jenkins/scripts/test-dashboard.sh
           '''
 
+          echo "Start new virtual environment pre-Slurm checks"
```
ajdecon, line 86: Do we need to explicitly tear down the VMs before we do this?

supertetelman: Tearing them down is part of this script.

ajdecon (approved):
LGTM. Confirmed that tests passed in nightly builds.

PR #1141, opened 2022-03-25
Misc fixes for Jenkins CI

ajdecon:
## Summary

Following the merge of #1043 to update our Kubespray version and migrate to containerd, we're seeing some failures in Jenkins CI.

This PR is intended to let me work through these failures until we get a successful build. 😄 

Changes include:

- https://github.com/NVIDIA/deepops/commit/a7c89bfe6de16367e8af5ea5a321aeeb9bba671b: Update to maintained version of docker-registry chart
- https://github.com/NVIDIA/deepops/pull/1141/commits/e271f74c0700b1eb0581137ac30a7a23e9ba0300: Fixes for local registry test in Kubernetes with use of containerd
- https://github.com/NVIDIA/deepops/pull/1141/commits/5e7c681c2b8910e8a5133b10d3abee300539dcdd: Fixes to monitoring tests to update stack to one that works with later Kubernetes, and account for use of GPU operator

## Test plan

All CI tests should pass

### DIFF
--- a/config.example/group_vars/k8s-cluster.yml
+++ b/config.example/group_vars/k8s-cluster.yml
@@ -69,6 +69,15 @@ docker_insecure_registries: "{{ groups['kube-master']|map('regex_replace', '^(.*
 crio_insecure_registries: "{{ groups['kube-master']|map('regex_replace', '^(.*)$', '\\1:5000')|list + ['registry.local:31500']}}"
 docker_registry_mirrors: "{{ groups['kube-master'] | map('regex_replace', '^(.*)$', 'http://\\1:5000') | list }}"
 
+# TODO: Add support in containerd for automatically setting up registry
+# mirrors, not just the k8s-local registry
+containerd_insecure_registries:
+  "registry.local:31500": "http://registry.local:31500"
+
+# Work-around for https://github.com/kubernetes-sigs/kubespray/issues/8529
+nerdctl_extra_flags: " --insecure-registry"
+image_command_tool: "crictl"
+
 ################################################################################
 # Logging with rsyslog                                                         #
 ################################################################################
--- a/workloads/jenkins/Jenkinsfile
+++ b/workloads/jenkins/Jenkinsfile
@@ -122,11 +122,6 @@ pipeline {
           sh '''
             timeout 60 bash -x ./workloads/jenkins/scripts/test-slurm-gpu.sh
           '''
-
-          echo "Test DCGM metrics"
-          sh '''
-            timeout 500 bash -x ./workloads/jenkins/scripts/test-dcgm-metrics.sh slurm-node
-          '''
         }
       }
     }
--- a/workloads/jenkins/scripts/test-monitoring.sh
+++ b/workloads/jenkins/scripts/test-monitoring.sh
@@ -123,7 +123,15 @@ set -e # The loop is done, and we got debug if it failed, re-enable fail on erro
 
 # Get some debug for Pods that did/didn't come up and verify DCGM metrics
 kubectl get all -n monitoring
-bash -x ./workloads/jenkins/scripts/test-dcgm-metrics.sh slurm-node # We use slurm-node here because it is GPU only, kube-node includes the mgmt plane
+
+# Check for dcgm-exporter pods that are not running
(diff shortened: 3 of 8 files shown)


config.example/group_vars/k8s-cluster.yml
```diff
@@ -69,6 +69,11 @@ docker_insecure_registries: "{{ groups['kube-master']|map('regex_replace', '^(.*
 crio_insecure_registries: "{{ groups['kube-master']|map('regex_replace', '^(.*)$', '\\1:5000')|list + ['registry.local:31500']}}"
 docker_registry_mirrors: "{{ groups['kube-master'] | map('regex_replace', '^(.*)$', 'http://\\1:5000') | list }}"
 
+# TODO: Add support in containerd for automatically setting up registry
+# mirrors, not just the k8s-local registry
+containerd_insecure_registries:
+        "registry.local:31500": "http://registry.local:31500"
+
```
dholt, line 76: Is there a reason this doesn't have the same format as the similar values above?

ajdecon: The format for this variable was actually just changed in Kubespray! See release notes for v2.18.1: https://github.com/kubernetes-sigs/kubespray/releases/tag/v2.18.1

Interestingly, I originally misread this and added it too early, when I thought it applied to v2.18.0.

However, v2.18.1 appears to have bugfixes we need anyway to get insecure registries to work. I'm testing an update to that release now, in which case this change will still apply. :-)

config.example/group_vars/k8s-cluster.yml
```diff
@@ -69,6 +69,15 @@ docker_insecure_registries: "{{ groups['kube-master']|map('regex_replace', '^(.*
 crio_insecure_registries: "{{ groups['kube-master']|map('regex_replace', '^(.*)$', '\\1:5000')|list + ['registry.local:31500']}}"
 docker_registry_mirrors: "{{ groups['kube-master'] | map('regex_replace', '^(.*)$', 'http://\\1:5000') | list }}"
 
+# TODO: Add support in containerd for automatically setting up registry
+# mirrors, not just the k8s-local registry
+containerd_insecure_registries:
+  "registry.local:31500": "http://registry.local:31500"
+
+# Work-around for https://github.com/kubernetes-sigs/kubespray/issues/8529
+nerdctl_extra_flags: " --insecure-registry"
```
supertetelman, line 78: Do we need to call out somewhere that we are defaulting to use this flag?

workloads/jenkins/Jenkinsfile
```diff
@@ -122,11 +122,6 @@ pipeline {
           sh '''
             timeout 60 bash -x ./workloads/jenkins/scripts/test-slurm-gpu.sh
           '''
-
-          echo "Test DCGM metrics"
-          sh '''
-            timeout 500 bash -x ./workloads/jenkins/scripts/test-dcgm-metrics.sh slurm-node
```
supertetelman, line 128: Why are we removing this test?

workloads/jenkins/scripts/test-monitoring.sh
```diff
@@ -123,7 +123,15 @@ set -e # The loop is done, and we got debug if it failed, re-enable fail on erro
...
-bash -x ./workloads/jenkins/scripts/test-dcgm-metrics.sh slurm-node # We use slurm-node here because it is GPU only, kube-node includes the mgmt plane
+
+# Check for dcgm-exporter pods that are not running
+if kubectl get pods -n gpu-operator-resources -l app=nvidia-dcgm-exporter | grep nvidia-dcgm-exporter | grep -v Running; then
+  echo "Some nvidia-dcgm-exporter pods are not in state Running"
+  exit 1
+fi
+
+# Commented out test-dcgm-metrics test, as with gpu-operator we no longer expose port 9400 directly on the nodes
+#bash -x ./workloads/jenkins/scripts/test-dcgm-metrics.sh slurm-node # We use slurm-node here because it is GPU only, kube-node includes the mgmt plane
```
supertetelman, line 134: We should key off the "DEEPOPS_K8S_OPERATOR" variable for this. If set to "true we are using the operator. If left blank or set to false we are using device plugin.

We still claim to support both and the nightly testing goes down both paths.

supertetelman (approved):
LGTM, need to modify the DCGM tests to still run in some cases, but I will handle that in the other PR where we are adding more K8s test scenarios and already touching the same scripts.

PR #1150, opened 2022-04-01
Update jenkins matrix test to support driver container vs host driver

supertetelman:
* Add DEEPOPS env vars to jeknins script output for debug
* Add ability to specify host or container drivers in manual matrix test in Jenkins


ajdecon (approved):
LGTM

PR #1160, opened 2022-04-18
Update to nvidia.nvidia_driver role v2.2.0

ajdecon:
Updates to use R510 driver by default.

## Test plan

* Passing Jenkins tests


dholt (approved):
approving since jenkins passed, but i'm curious why all the molecule tests failed

ajdecon:
@dholt : For the molecule tests, see #1158

PR #1166, opened 2022-04-28
Update NVIDIA signing key

ajdecon:
## Summary

Repository signing keys have changed, so we need to update the automation to match. https://developer.nvidia.com/blog/updating-the-cuda-linux-gpg-repository-key/

- [x] Update install process for nvidia_cuda and nvidia_dcgm
- [x] Remove nvidia-ml as this repository is deprecated and will not be updated
- [x] Update `roles/requirements.txt` for updated version of `nvidia.nvidia_driver` role (depends on https://github.com/NVIDIA/ansible-role-nvidia-driver/pull/59)

## Test plan

Passing CI tests.


roles/nvidia_cuda/tasks/install-ubuntu.yml
```diff
@@ -4,25 +4,22 @@
...
-- name: Ubuntu | add pin file
-  copy:
-    src: "cuda-ubuntu.pin"
-    dest: "/etc/apt/preferences.d/cuda-repository-pin-600"
-    owner: "root"
-    group: "root"
-    mode: "0644"
-
-- name: Ubuntu | add key
+- name: Ubuntu | ensure old key is absent
```
supertetelman, line 7: Do we want to eventually remove this or does it stay here forever?

supertetelman (approved):
LGTM, let's see it pass testing.

roles/nvidia_cuda/tasks/install-ubuntu.yml
ajdecon: Unclear. In principle I think this could be removed once we were confident all existing installs were updated. But it's hard to know when that is. 🤷🏻‍♂️ 

If I were being ultra-conservative, I'd say this could be removed when the [last released drivers signed by the old key are EOL](https://docs.nvidia.com/datacenter/tesla/drivers/#cuda-drivers) (r470, July 2024). We might be able to come up with a faster schedule than that though.

However, I don't think this task is doing any *harm* (apart from code clutter). So I don't think there would be any rush to remove this.

supertetelman (approved):
It's passing

PR #1182, opened 2022-06-15
Have same slurm.conf among nodes and controller

seyong-um:
Share slurm.conf to have same of it - in case of configuring nodes and/or partitions.
Locate slurm.conf at /sw/.slurm/slurm.conf of controller where the /sw is nfs mount point to the all nodes.
Let /etc/slurm/slurm.conf be a soft link to it from nodes and controller.

By Hyundai Motor Group / AIRS Company
Signed-off-by: Seyong Um <seyong.um@hyundai.com>

### DIFF
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -91,6 +91,9 @@ slurm_enable_nfs_client_nodes: true
 nfs_server_group: "slurm-master[0]"
 nfs_client_group: "slurm-master[1:],slurm-node"
 
+# Flags for sharing slurm.conf among compute and control nodes
+slurm_conf_symlink: false
+
 ################################################################################
 # SOFTWARE MODULES (SM)                                                        #
 #   May be built with either EasyBuild or Spack                                #
--- a/roles/pyxis/defaults/main.yml
+++ b/roles/pyxis/defaults/main.yml
@@ -1,6 +1,7 @@
 slurm_install_pyxis: true
 slurm_install_prefix: /usr/local
 slurm_config_dir: /etc/slurm
+slurmctl_config_dir: /sw/.slurm
 slurm_pyxis_version: 0.11.1
 slurm_pyxis_tarball_url: "https://github.com/NVIDIA/pyxis/archive/v{{ slurm_pyxis_version }}.tar.gz"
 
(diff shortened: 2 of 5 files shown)


ajdecon (changes requested):
@seyong-um : This functionality looks useful to support, and I know this is a common pattern in a lot of Slurm clusters.

However, I don't think we want to enable this by default, as we have several known use cases where the Slurm controller and the compute nodes don't share an NFS filesystem.

Would you be willing to put this behind a flag variable? I.e., add a boolean variable like `slurm_conf_symlink` which defaults to `false`, and only set up the symlink if it's set to `true`. Otherwise, use the existing logic.

seyong-um:
I added share-slurm-conf flag as you suggested. The default value is false, it is existing logic, once it has been set to true then new logic will run.

config.example/group_vars/slurm-cluster.yml
```diff
@@ -91,6 +91,9 @@ slurm_enable_nfs_client_nodes: true
 nfs_server_group: "slurm-master[0]"
 nfs_client_group: "slurm-master[1:],slurm-node"
 
+# Flags for sharing slurm.conf among compute and control nodes
+slurm_conf_symlink: false
```
ajdecon, line 95: Can you duplicate this line in `roles/slurm/defaults/main.yml` as well as well?

This will make sure the role itself is functional, even if this line isn't present in a cluster config. This will impact upgrades of older clusters and some of our CI workflows.

roles/pyxis/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
 slurm_install_pyxis: true
 slurm_install_prefix: /usr/local
 slurm_config_dir: /etc/slurm
+slurmctl_config_dir: /sw/.slurm
```
ajdecon, line 4: Why is this in the pyxis role?

ajdecon (changes requested):
Our CI tests are currently having issues, but manual testing of the cluster setup process looks good from my end.

Added a couple of inline comments with additional changes. Once those are done I can merge.

roles/pyxis/defaults/main.yml
seyong-um: Removed.

config.example/group_vars/slurm-cluster.yml
seyong-um: Flag is now placed in that file for the same.

ajdecon (approved):
Thank you! All code review comments are addressed, all CI tests are passing, and manual tests work as well. Looks good to me!

PR #1183, opened 2022-06-15
Fix ssh disconnection on compute nodes

seyong-um:
Once slurm job completed, slurm terminates slurm user's all processes by epilog script. 

Suppose the environment that a user connects to compute node by ssh and looks job progress in there, the ssh connection will be lost by killing user processes.

Not to kill the ssh and it's shell (bash) so, the ssh connection will be kept.

by Hyundai Motor Group / AIRS Company

### DIFF
--- a/roles/slurm/templates/etc/slurm/epilog.d/40-lastuserjob-processes
+++ b/roles/slurm/templates/etc/slurm/epilog.d/40-lastuserjob-processes
@@ -1,8 +1,17 @@
 #!/usr/bin/env bash
 set -ex
 
-if [ "$SLURM_JOB_USER" != root ]; then
-    if killall -9 -u "$SLURM_JOB_USER" ; then
+# Retrieve slurm user processes
+ps h -u $SLURM_JOB_USER | while read PROCESS
+do
+    PID=`echo $PROCESS | awk '{print $1}'`
+    PNAME=`echo $PROCESS | awk '{print $5}'`
+
+    # Kill residual user processes except: user ssh connection to compute nodes
+    # "sshd:" : A process created for ssh connection
+    # "-bash" : A shell process which is child of ssh connection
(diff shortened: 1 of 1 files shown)


itzsimpl:
I think this effectively gives the user (once they crated a job on a compute node) an infinite access to it. There is no gurantie that the user will ever disconnect. What happens with resource limitation? E.g. number of cores, access to GPUs, memory; it should be torn down/reverted/relaxed once the SLURM job has finished (managed by slurm). If I'm not mistaken, running ssh to a SLURM compute node, pins the user to the specific process under which the SLURM job is run (effectively imposing the limitations about allocated resources set by the job). So what effect does it have to leave the ssh process open?

On another note, let @ajdecon correct me if I'm wrong, but flags
https://github.com/NVIDIA/deepops/blob/e91ba4959b5508ea14ccca02ab845d38ef91933a/config.example/group_vars/slurm-cluster.yml#L43-L55
will let you disable the restriction to ssh on a compute node if not running a job, and also the killing of all active processes once the job terminates. In essence what you are trying to override.

ajdecon:
In most cases, we do want to kill SSH access to a node when a user's job ends.

This is an intentional measure to prevent indefinite access (as @itzsimpl highlights), as well as to ensure that no resources are used by users who do not have an active job.

@itzsimpl points out a couple of variables for allowing specific users indefinite SSH access, as well as being able to configure "login on compute" functionality if your login node is also running Slurm jobs.

Also, if you want to use a customized set of prolog/epilog scripts on your cluster, I'll point out the [`prolog_fileglob` and `epilog_fileglob` variables](https://github.com/NVIDIA/deepops/blob/master/roles/slurm/defaults/main.yml#L82) that will let you specify paths to an alternate set of scripts.

I'm closing this PR, but please let us know if I'm misunderstanding your use case, or if one of the alternate solutions above isn't acceptable.

PR #1187, opened 2022-06-21
Fix bugs preventing slurm reinstall or rebuild

biocyberman:
Bugs:

1. docker.slurm-exporter.service creates subdirectories
`/usr/loca/bin/{sinfo,sdiag,squeue}` when the excutables are absent
(due to uninstall process). When deepops tries to install the executables
again, they are put under the unwanted subdirectories (e.g.
/usr/local/bin/sinfo/sinfo instead of /usr/local/bin/sinfo)

2. Slurm 'make uninstall' doesn't remove files under `"{{ slurm_install_prefix
}}/lib/slurm"`. Therefore the directory needs to be removed. A safer solution
would be to run the correct 'make uninstall' command (at the correct location)
to uninstall the files. Without this handling, slurm commands fails to run because of
library version conflicts.

### DIFF
--- a/roles/slurm/tasks/build.yml
+++ b/roles/slurm/tasks/build.yml
@@ -110,6 +110,13 @@
   when: slurm_build
   block:
 
+    # This service should be stopped before uninstalling slurm
+  - name: stop docker slurm exporter service to prevent subfolder creation
+    service:
+      name: "docker.slurm-exporter.service"
+      state: stopped
+    when: slurm_build and (ansible_distribution == 'Ubuntu') and ("{{ansible_hostname}}" not in groups['slurm-node'])
+
   - name: check that we have an old build dir
     stat:
       path: "{{ slurm_build_dir }}"
@@ -125,6 +132,11 @@
   - name: remove old build dir
     file:
       path: "{{ slurm_build_dir }}"
+      state: absent
+
+  - name: remove old lib dir
+    file:
+      path: "{{ slurm_install_prefix }}/lib/slurm"
       state: absent
 
 - name: make build directory
--- a/roles/slurm/tasks/build.yml
+++ b/roles/slurm/tasks/build.yml
@@ -106,16 +106,19 @@
   when: ansible_os_family == 'RedHat'
   failed_when: false
 
+- name: gather service facts
+  ansible.builtin.service_facts:
+
 - name: when building, try to uninstall first before downloading source
   when: slurm_build
   block:
-
     # This service should be stopped before uninstalling slurm
   - name: stop docker slurm exporter service to prevent subfolder creation
     service:
       name: "docker.slurm-exporter.service"
       state: stopped
-    when: slurm_build and (ansible_distribution == 'Ubuntu') and ("{{ansible_hostname}}" not in groups['slurm-node'])
+    when: (ansible_distribution == 'Ubuntu') and 
+      ("docker.slurm-exporter.service" in ansible_facts.services)
 
   - name: check that we have an old build dir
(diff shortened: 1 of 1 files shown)


roles/slurm/tasks/build.yml
```diff
@@ -110,6 +110,13 @@
   when: slurm_build
   block:
 
+    # This service should be stopped before uninstalling slurm
+  - name: stop docker slurm exporter service to prevent subfolder creation
+    service:
+      name: "docker.slurm-exporter.service"
+      state: stopped
+    when: slurm_build and (ansible_distribution == 'Ubuntu') and ("{{ansible_hostname}}" not in groups['slurm-node'])
```
ajdecon, line 118: This fails on initial installs because the exporter hasn't been installed yet. The error looks like this:

```
fatal: [virtual-login01-2]: FAILED! => changed=false 
  invocation:
    module_args:
      daemon_reexec: false
      daemon_reload: false
      enabled: null
      force: null
      masked: null
      name: docker.slurm-exporter.service
      no_block: false
      scope: system
      state: stopped
  msg: 'Could not find the requested service docker.slurm-exporter.service: host'
```

You should ensure that service facts are being collected, then add a condition so this only executes when the service exists. I.e., something like this:

```
- name: gather service facts
  ansible.builtin.service_facts:

- name: stop docker slurm exporter service to prevent subfolder creation
  service:
      name: "docker.slurm-exporter.service"
      state: stopped
  when: 
  - slurm_build
  - ansible_distribution == 'Ubuntu'
  - "{{ansible_hostname}}" not in groups['slurm-node']
  - "docker.slurm-exporter.service" in ansible_facts.services
```

(I haven't tested this change directly, but I'm pretty sure the syntax looks like that!)

roles/slurm/tasks/build.yml
```diff
@@ -127,6 +134,11 @@
       path: "{{ slurm_build_dir }}"
       state: absent
 
+  - name: remove old lib dir
```
ajdecon, line 137: This should only be executed when `slurm_build` is true. Otherwise this makes the Slurm role non-idempotent, and could break functional clusters.

biocyberman: > This should only be executed when `slurm_build` is true. Otherwise this makes the Slurm role non-idempotent, and could break functional clusters.

It is already under the same block with `when: slurm_build` with other item above it.

roles/slurm/tasks/build.yml
biocyberman: @ajdecon  I incorporated your suggestion and also restarted the service back again in the end. Ran a test deployment and also rebased to master.

biocyberman: By the way, may be because of older ansible version, the `when` list doesn't work so I use the explicit logical `and`.

ajdecon (changes requested):
This looks great, thank you! I can confirm our CI tests are passing and my manual tests also look good.

Last thing: please sign off your commits according to our [CONTRIBUTING.md](https://github.com/NVIDIA/deepops/blob/master/CONTRIBUTING.md) doc. Once this is done, I can merge this PR.

biocyberman:
> This looks great, thank you! I can confirm our CI tests are passing and my manual tests also look good.
> 
> Last thing: please sign off your commits according to our [CONTRIBUTING.md](https://github.com/NVIDIA/deepops/blob/master/CONTRIBUTING.md) doc. Once this is done, I can merge this PR.
Sounds great!
I amended the last commit and signed it off. Thanks.

PR #1191, opened 2022-07-01
Specify runtime_path partition size

seyong-um:
Problem:
 - In the case a docker image is bigger than `/run` partition size then slurm job fails: no space left on device
 - Cf.) nvcr.io/nvidia/pytorch is about 14GB, aws g4dn.2xlarge instance `/run` is about 3GB
 - https://github.com/NVIDIA/pyxis/issues/53

Cause:
 - Pyxis uses `runtime_path` to store squashfs file temporarily
 - The default location of `runtime_path` is `/run/pyxis`.
 - `/run` size might be up to distribution, however ubuntu is 10% of physical memory

Suggestion:
 - Add `resize_run_partition` option to specify `/run` partition size 

I know we can change `runtime_path` to other location, but I believe it is meaningful to use `tmpfs` residing in memory faster than other filesystems if compute nodes have sufficient memory.

### DIFF
--- a/config.example/group_vars/slurm-cluster.yml
+++ b/config.example/group_vars/slurm-cluster.yml
@@ -179,6 +179,12 @@ slurm_install_enroot: true
 slurm_install_pyxis: true
 slurm_pyxis_version: 0.11.1
 
+# /run is default partition of pyxis runtime_path
+resize_run_partition: false
+
+# /run tmpfs size. ubuntu default is 10%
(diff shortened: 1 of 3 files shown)


ajdecon (approved):
LGTM!

- All CI tests are passing, so the default case is working well
- Manual test with `resize_run_partition: false` behaved as expected
- Manual test with `resize_run_partition: true` increased the size of `/run` correctly

PR #1198, opened 2022-07-08
Add alertmanager for slurm cluster

0leaf:
Currently when I deploy slurm cluster environment with deepops, alertmanager is not deployed.
(In the [k8s manifest files](https://github.com/NVIDIA/deepops/blob/master/workloads/examples/k8s/services/prometheus-monitor.yml#L475), I checked the alertmanager configuration and followed it as closely as possible.)

- It is configured based on docker like prometheus and grafana.
- alert_rules.yml.j2 is configured similarly to the sample in k8s, and is configured to receive an alarm when there is a problem with the exporter of each node.
- It follows the structure and conventions of deepops as much as possible.

**It is a big change that an alertmanager is additionally configured in the login node, so please comment if there are any anticipated problems.**


roles/alertmanager/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
+---
+- name: create config dir
+  file:
+    path: "{{ alertmanager_config_dir }}"
+    state: directory
+    mode: "0755"
+
+- name: install config file
+  template:
+    src: templates/alertmanager.yml.j2
```
ajdecon, line 10: As written, this will make it a little difficult to adjust the AlertManager config without directly modifying the DeepOps role.

Instead of providing a hard-coded config file, can you make this a variable? E.g.,

```
- name: install config file
  template: "{{ alertmanager_config_src }}"
...
```

With `alertmanager_config_src: templates/alertmanager.yml.j2` in `defaults/main.yml`.

This will allow the user to substitute in their own config file without modifying DeepOps, and making subsequent DeepOps release updates more difficult.

roles/prometheus/tasks/main.yml
```diff
@@ -11,6 +11,20 @@
...
 
+- name: create alarm rules dir
+  file:
+    path: "{{ prometheus_config_dir }}/rules"
+    state: directory
+    mode: "0755"
+
+- name: install alarm rules file
+  template:
+    src: templates/alert_rules.yml.j2
```
ajdecon, line 22: Similar to the comment on `roles/alertmanager/tasks/main.yml`, please make this a variable to allow the user to easily substitute in their own rules file. E.g.,

```
- name: install alarm rules file
  template:
    src: "{{ alert_rules_src }}"
```

ajdecon (changes requested):
Thank you for putting this together! I'm excited to merge it, there are just a few changes I'd like to see.

1. I added two inline comments with suggested changes around how configuration files are managed. This will hopefully make it easier for the user to modify the configuration.
2. Additionally, can you please add a short section on AlertManager to the [Slurm monitoring doc](https://github.com/NVIDIA/deepops/blob/master/docs/slurm-cluster/slurm-monitor.md)?
    This does not need to be too extensive, but should include a link to the [AlertManager website](https://prometheus.io/docs/alerting/latest/alertmanager/) and tell the user which files to change to configure alert rules and alert delivery (e.g., via Slack as in your example file).

roles/alertmanager/tasks/main.yml
0leaf: Thanks for your comments. :)
When I looked at the [Prometheus configuration](https://github.com/NVIDIA/deepops/blob/master/roles/prometheus/tasks/main.yml#L16), I checked that it was configured as a relative path and followed it as it is. However, I agree with your comment and changed the configuration of Prometheus as well.

roles/prometheus/tasks/main.yml
0leaf: Thanks, I fixed not only this part but also [Prometheus config part](https://github.com/NVIDIA/deepops/pull/1198/commits/2927a595ca91390e0a2568a1473cdd341313b80d).

0leaf:
Thanks for your kind feedback.

I agree with your opinion and have reflected on the following, please comment if there is any content that needs to be corrected or that needs to be kept in the convention.
- Changed some hard-coded parts to a variable method.
- Added Alertmanager part on README.md

ajdecon (approved):
Nice! Your additional changes look good to me, and all tests are still passing. Approved and will merge.

PR #1249, opened 2022-12-19
Kubeflow v1.6.1 Upgrade & drop failing docker runtime tests

supertetelman:
* Tweaked some of the tests to pass. Now that we are at K8s v1.24 containerd testing is all that is needed, as docker support is dropped in K8s-proper.
* Bumped Kubeflow version to v1.6.1


ajdecon (approved):
LGTM

PR #1261, opened 2023-04-15
Conform to standard gpu operator namespacing

supertetelman:
When the GPU Operator was originally launched it installed into "default" and as a best-practice we installed it into "gpu-operator-resources" namespace. They now install everything into "gpu-operator" and cleaned up some bugs that prevented this cleaner install.

This PR conforms to their naming standard and upates a few of our installation scripts and tests.

I'm hoping this will make alignment with official documentation and additional addon installs such as kube-prometheus-stack easier to follow going forward.


ajdecon (approved):
LGTM

PR #1279, opened 2023-06-28
Minor: Fix hardcoded slurm username

jeremyfix:
The task hardcodes the slurm username while the variable file does contain a field for this property.


morsoinferno:
It seems ok to me

dholt (approved):

PR #1294, opened 2023-08-03
Release updates

dholt:
Updates to software versions

Move version config out to roles so it's easier to manage


supertetelman (approved):
LGTM, just put in release notes the versions we are pulling out/changing here.

PR #1355, opened 2026-07-01
fix: handle non-dict parsed JSON in firmware version manifest parser

dholt:
## Summary
- `roles/nvidia-dgx-firmware/files/parse_manifest.py` `parse_versioning` keeps the last `json.loads`-parseable line from the firmware container output, then indexes it with a string key (`manifest_json['ErrorWritingVersioning']`).
- If that last parseable line decodes to a non-dict (e.g. a JSON array), the string-key index raises `TypeError`. Only `KeyError` was caught, so the exception escaped and aborted the task.
- Guard that the parsed value is a `dict` before indexing; a non-dict now falls through to the existing "No JSON could be loaded" path and exits `1` cleanly instead of crashing.

## Validation
- `git diff --check origin/master...HEAD`
- `py_compile` of the changed script
- Reproduced the crash before the fix (non-dict last line → uncaught `TypeError`) and confirmed the clean `exit 1` after the fix.
- Added a self-contained regression test and ran it:
  `python -m unittest discover -s roles/nvidia-dgx-firmware/tests -v` → `Ran 1 test ... OK`
- Public sanitizer pass on the PR body and added diff lines.

## Notes
- Behavior is unchanged for well-formed manifests; this only hardens the parser against malformed/non-dict input.
- No pre-existing tests covered this script/role; the new test uses only the standard library and a tempfile, no external fixtures.

### DIFF
--- a/roles/nvidia-dgx-firmware/files/parse_manifest.py
+++ b/roles/nvidia-dgx-firmware/files/parse_manifest.py
@@ -3,7 +3,7 @@
 import json
 
 from collections import OrderedDict
-from sys import argv
+from sys import argv, exit
 
 
 def return_json(payload):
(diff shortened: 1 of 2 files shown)


dholt:
@michael-balint Gentle ping on this one when you have a moment — it's a small, self-contained robustness fix for `parse_manifest.py` with a stdlib-only regression test. All 31 checks are green and it's mergeable. Thanks!

michael-balint (approved):

PR #1363, opened 2026-07-15
fix: restart containerized exporters when Docker restarts

### LINKED ISSUE #1267: node exporters don't work after initial run of slurm playbook
The node stat exporters don't work after an initial install until either docker has been restarted or the node rebooted.
I suspect the playbooks are just missing a restart of docker after setting up the exporters / ip_forwarding configuration on the node.

dholt:
## Summary
- Add `PartOf=docker.service` to the six containerized monitoring service units (node exporter, DCGM exporter, Slurm exporter, Prometheus, Grafana, Alertmanager).
- Root cause of #1267: the units declare `Requires=docker.service` without `PartOf=`, so an explicit Docker restart propagates as a one-way stop to the dependents and nothing starts them again. A first `slurm-cluster.yml` run on GPU nodes triggers exactly this — the NVIDIA container runtime configuration restarts Docker after the node exporter is already up — leaving exporters down until reboot.
- With `PartOf=`, a Docker restart propagates as a restart, so the units come back automatically (including after any manual `systemctl restart docker`).

## Validation
- `git diff --check`; full role lint: 0 failures, 0 warnings.
- Live single-node GPU validation summary will be posted as a follow-up comment before review: fresh `slurm-cluster.yml` run, then asserting the node exporter is active at the end of the run and recovers after an explicit Docker restart.

## Notes
- Fixes #1267 (stale-closed; the maintainer had twice marked it "not stale" — the failure mode is present on master until this change).
- One line per unit; no change to start ordering or normal operation.

### DIFF
--- a/roles/alertmanager/templates/docker.alertmanager.service.j2
+++ b/roles/alertmanager/templates/docker.alertmanager.service.j2
@@ -2,6 +2,7 @@
 Description=Alert Manager
 After=docker.service
 Requires=docker.service
+PartOf=docker.service
 
 [Service]
 TimeoutStartSec=0
--- a/roles/grafana/templates/docker.grafana.service.j2
+++ b/roles/grafana/templates/docker.grafana.service.j2
@@ -2,6 +2,7 @@
 Description=Grafana
 After=docker.service
 Requires=docker.service
+PartOf=docker.service
 
 [Service]
 TimeoutStartSec=0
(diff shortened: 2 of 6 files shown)


dholt:
## Live validation summary

Validated on a disposable dual-GPU server (Ubuntu 22.04) with a fresh full `slurm-cluster.yml` deployment from a branch carrying this change:

- The rendered unit files contain `PartOf=docker.service`.
- **Reproduction of the reported failure sequence:** the node exporter was deployed and active, then the DCGM exporter playbook's NVIDIA container runtime configuration fired its `restart docker` handler — the exact sequence that leaves exporters dead on master. With this change the node exporter was **active at the end of the run**.
- **Explicit restart recovery:** `systemctl restart docker` on the node; both the node exporter and Slurm exporter units returned to active automatically, and `:9100/metrics` served node metrics.
- Slurm remained healthy throughout (node idle, `srun --gpus=1 nvidia-smi` passing).

michael-balint (approved):

Issue #1307, opened 2024-05-27
Errors in deepops/slurm-exporter


fa-ina-tic:
OS on Master Nodes: Ubuntu 22.04 LTS

Hello,

I've encountered some issues with monitoring our Slurm cluster. Below are the details:

Problem 1: Service Failure
The `docker.slurm-exporter.service` keeps failing with the following log output:
```
May 12 02:35:56 login01 systemd[1]: Starting Prometheus Slurm Exporter...
May 12 02:35:56 login01 docker[2222118]: Error response from daemon: No such container: docker.slurm-exporter.service
May 12 02:35:56 login01 docker[2222131]: Error: No such container: docker.slurm-exporter.service
May 12 02:35:57 login01 docker[2222143]: latest: Pulling from deepops/prometheus-slurm-exporter
May 12 02:35:57 login01 docker[2222143]: Digest: sha256:9a0c657a465fe20209093aec38d44e8bc55d5de282a62827e623698ee9944048
May 12 02:35:57 login01 docker[2222143]: Status: Image is up to date for deepops/prometheus-slurm-exporter:latest
May 12 02:35:57 login01 docker[2222143]: docker.io/deepops/prometheus-slurm-exporter:latest
May 12 02:35:42 login01 systemd[1]: Started Prometheus Slurm Exporter.
May 12 02:35:43 login01 docker[2222028]: time="2024-05-12T02:35:43Z" level=info msg="Starting Server: :8080" source="main.go:43"
May 12 02:35:56 login01 docker[2222028]: 2024/05/12 02:35:56 exit status 1
May 12 02:35:56 login01 systemd[1]: docker.slurm-exporter.service: Main process exited, code=exited, status=1/FAILURE
May 12 02:35:56 login01 systemd[1]: docker.slurm-exporter.service: Failed with result 'exit-code'.
May 12 02:35:56 login01 systemd[1]: docker.slurm-exporter.service: Scheduled restart job, restart counter is at 16205.
May 12 02:35:56 login01 systemd[1]: Stopped Prometheus Slurm Exporter.
```

To solve problem 1, I ran the Docker image manually and checked the sinfo, squeue, and sdiag commands (as the service attempts to do). The following error messages were encountered:
```
root@mlops-slurm-master-01:/lib64# sdiag
sdiag: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by sdiag)
[...]

PR #1365, opened 2026-07-15
fix: build the Slurm exporter image locally with a current base

### LINKED ISSUE #1328: The Docker image 'deepops/prometheus-slurm-exporter' is based on too old version of Ubuntu.
It's been already known that the current version of prometheus-slurm-exporter image occurs issue due to the old Ubuntu version of the base image.
[Issue 1307](https://github.com/NVIDIA/deepops/issues/1307)

vpenso's prometheus-slurm-exporter is not maintained actively now, but I checked it still works fine with the Ubuntu 22.04 version and Golang >=1.24.

Here's the [Dockerfile](https://gist.github.com/jungyh0218/3ad03dbb79bfbf66be3a0fdd510af98c) I wrote. I think the current version of deepops/prometheus-slurm-exporter & docker.slurm-exporter.service file should be updated.

I also have a plan to make the pull request in timely manner.

dholt:
## Summary
- Stop pulling `deepops/prometheus-slurm-exporter:latest` — a 2020-era external image with a single unpinned tag and no in-repo build path. Its old glibc is incompatible with the Slurm client binaries bind-mounted from modern hosts, so the service crash-loops on current deployments (#1328, #1307).
- Build the exporter image locally from a bundled Dockerfile instead: vpenso `prometheus-slurm-exporter` 0.20 (same metric names as the shipped Grafana dashboard) on an Ubuntu 24.04 base with the system libraries the mounted Slurm binaries require. The tag is pinned; no registry access is needed.
- Sites can keep using their own image with `slurm_exporter_build_image: false` and a `slurm_exporter_container` override (the unit then retains the pull step).
- Mount the full Slurm install `lib` directory (superset of the previous `lib/slurm` plugin mount) so `libslurm` resolves alongside the plugins, and mount `sshare`, which the exporter's fair-share collector executes but the unit never provided.
- Create the `slurm` system user in the image (the mounted Slurm clients refuse to parse `slurm.conf` without it) and patch an upstream exporter bug at build time: the node collector's `sinfo -O` call uses default field widths, so node names of 20+ characters merge into the next column and panic the exporter on every scrape (present through upstream master; patch is a candidate for upstream contribution).

## Validation
- YAML parse and full role lint: 0 failures; `ansible-playbook --syntax-check playbooks/slurm-cluster.yml`.
- Live single-node GPU validation summary will be posted as a follow-up comment before review: fresh Slurm deployment with the exporter role, asserting the image builds locally, the service stays active (no crash loop), and `/metrics` serves `slurm_*` series.

## Notes
- Fixes #1328. The previously shipped Grafana dashboard is unchanged — the 0.20 exporter emits the same metric names.
- Alternative considered: switching to the actively maintained rivosinc exporter; rejected for this fix because its metric names would break the shipped dashboard. Candidate for a future train.

### DIFF
--- a/roles/prometheus-slurm-exporter/defaults/main.yml
+++ b/roles/prometheus-slurm-exporter/defaults/main.yml
@@ -1,4 +1,9 @@
-slurm_exporter_container: "deepops/prometheus-slurm-exporter:latest"
+# Built locally from the bundled Dockerfile by default; set
+# slurm_exporter_build_image: false and point slurm_exporter_container at a
+# site image to use a registry instead.
+slurm_exporter_container: "deepops/prometheus-slurm-exporter:0.20-1"
+slurm_exporter_build_image: true
+slurm_exporter_build_dir: /opt/deepops/build/slurm-exporter
 slurm_exporter_svc_name: "docker.slurm-exporter.service"
 slurm_exporter_state: started
 slurm_exporter_enabled: yes
--- a/roles/prometheus-slurm-exporter/files/docker/Dockerfile
+++ b/roles/prometheus-slurm-exporter/files/docker/Dockerfile
@@ -1,0 +1,34 @@
+# Prometheus Slurm exporter image, built locally by the
+# prometheus-slurm-exporter role (no registry pull required).
+#
+# The container bind-mounts the host's Slurm client binaries and libraries,
+# so the runtime stage provides a current glibc plus the system libraries
+# those binaries link that are not part of the Slurm install prefix.
+ARG SLURM_EXPORTER_BASE_IMAGE=ubuntu:24.04
+
+FROM golang:1.24 AS build
+ARG SLURM_EXPORTER_VERSION=0.20
+# The sed below fixes an upstream parsing bug present through vpenso master:
+# the node collector calls sinfo -O without field widths, so node names of
+# 20+ characters truncate into the next column and the exporter panics with
+# "index out of range". The ": " width suffix disables truncation.
+RUN git clone --depth 1 --branch ${SLURM_EXPORTER_VERSION} \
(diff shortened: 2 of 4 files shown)


dholt:
## Live validation summary

Validated on a disposable dual-GPU server (Ubuntu 22.04) with a fresh Slurm 26.05.1 deployment; the exporter role was applied from this branch. Live validation surfaced three real defects that are all addressed in the final head — each was reproduced, fixed, and re-verified on the node:

1. The container needs a resolvable `slurm` system user or the mounted Slurm clients refuse to parse `slurm.conf` — added to the image.
2. The exporter's fair-share collector executes `sshare`, which the unit never bind-mounted (a latent gap that predates this change) — mount added.
3. An upstream exporter bug present through vpenso master: the node collector calls `sinfo -O` without field widths, so node names of 20+ characters truncate into the next column and the exporter panics with `index out of range` on every scrape. Patched at image build with explicit no-truncation field widths; candidate for an upstream contribution.

Final state on the node: image `deepops/prometheus-slurm-exporter:0.20-1` built locally, service active with `NRestarts=0` (the original issue was a crash loop), `/metrics` returning HTTP 200 with 48 `slurm_*` series (`slurm_nodes_idle 1`, `slurm_cpus_total 32`, `slurm_queue_pending 0`), and the service recovering automatically after a Docker restart.

michael-balint (approved):

dholt:
The upstream exporter bug patched at image build here (node collector panic from `sinfo -O` field-width truncation on 20+ character node names) has been submitted upstream: vpenso/prometheus-slurm-exporter#128. If it merges, a future release can drop the build-time patch by bumping the pinned exporter version.

PR #1374, opened 2026-07-28
fix(setup): make setup.sh work when it is not run interactively

dholt:
## The bug

The usual no-argument `./scripts/setup.sh` invocation silently does nothing and
exits 0 when run non-interactively, including from CI outside this repository,
remote commands, pipelines, and agents.

```console
$ ./scripts/setup.sh </dev/null ; echo "exit=$?"
exit=0

$ bash scripts/setup.sh
... works normally
```

## Why

The old shebang was `#!/bin/bash --init-file`. `--init-file FILE` consumes the
script path as its option argument, leaving Bash with no script operand.

- In an interactive terminal, Bash sources the file as an init file and opens
  a shell with the virtual environment activated.
- Non-interactively, `--init-file` is ignored. Bash reads EOF from stdin and
  exits successfully without running the setup.

The old workflow invoked `bash scripts/setup.sh`, which supplied a script
operand and bypassed the shebang, so it did not cover the failing path.

## Why removing the trick is safe

The interactive convenience is redundant:

- The script already adds the activation command to `~/.bashrc`, so the
  environment activates automatically in later shells.
- The script prints the explicit activation command when it finishes.
- `--init-file` replaces `~/.bashrc`, so the subshell omitted the user's normal
  aliases, functions, and prompt.

Running `./scripts/setup.sh` from a terminal will no longer leave an activated
subshell; use the command it prints or open a new shell.

## Adjacent fixes

- The root check previously printed an error and used a bare `exit`, which
  inherited `echo`'s successful status. It now exits 1 and writes to stderr.
- The unsupported-OS message referenced the nonexistent `${DEPS_RPM[@]}`
  array. It now prints the actual RPM and Debian dependency arrays.

The unsupported-OS branch remains non-fatal because later setup is
distribution-independent and the existing `virtualenv` check still fails
loudly when a required dependency is genuinely absent.

## Regression coverage

The setup workflow now runs:

```bash
./scripts/setup.sh </dev/null
```

That exercises the executable/shebang path on both Ubuntu 22.04 and Ubuntu
24.04 instead of bypassing it with `bash scripts/setup.sh`.

Validation on this head includes:

- Bash syntax
- workflow YAML parsing
- direct non-interactive executable-path reproduction
- root-refusal exit status
- the Ubuntu 22.04 and Ubuntu 24.04 setup workflow jobs

## Follow-up

Add Rocky Linux to the dependency-install case so the distribution already
supported elsewhere in DeepOps receives automatic setup dependencies instead
of the manual fallback message. AlmaLinux and Oracle Linux can be evaluated
separately as Red Hat-compatible candidates rather than being implied as
currently supported.

### DIFF
--- a/.github/workflows/setup.yml
+++ b/.github/workflows/setup.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: run setup.sh
-        run: bash scripts/setup.sh
+        run: ./scripts/setup.sh </dev/null
 
       - name: verify ansible in venv
         run: |
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,4 +1,4 @@
-#!/bin/bash --init-file
+#!/usr/bin/env bash
 
 # DeepOps setup/bootstrap script
 #   This script installs required dependencies on a system so it can run Ansible
@@ -39,8 +39,8 @@ export DEBIAN_FRONTEND=noninteractive
 
 # Exit if run as root
 if [ $(id -u) -eq 0 ] ; then
-    echo "Please run as a regular user"
-    exit
+    echo "Please run as a regular user" >&2
+    exit 1
 fi
 
 # Proxy wrapper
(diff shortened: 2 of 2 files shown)


dholt:
Regression coverage is now part of this PR: the executable path runs on both
Ubuntu 22.04 and Ubuntu 24.04, and all checks pass on `ff760eef`. The PR body
has been updated to match the tested behavior. Ready for review.

michael-balint (approved):

PR #1375, opened 2026-08-07
docs: replace stale container image mirroring workflow

dholt:
## Summary
- replace the stale container-replicator workflow with a Skopeo `dir:` transport that preserves Docker manifest-list bytes and digests on Skopeo 1.22.2
- preserve multi-architecture images and require source-to-destination digest equality after the offline registry import
- retain a Docker pull/save/load fallback for single-platform images and document its architecture boundary
- keep credentials in temporary auth directories, require separately retained checksum/digest trust anchors, and create transfer ISOs unprivileged with strict shell failure handling
- separate the loopback-only test registry from the production TLS, authentication, reachability, and push/pull acceptance checklist

## Validation
- `git diff --check origin/master...HEAD`
- all 11 Bash examples pass `bash -n`
- public sanitizer pass on the PR body, full diff, and changed documentation
- live reproduction with Skopeo 1.22.2 confirmed `oci:` plus `--all --preserve-digests` fails for the documented CUDA manifest list; the corrected `dir:` transport preserves the source digest
- independent technical/security and field-operator review findings addressed

## Notes
- the documented OCI transport limitation is scoped to Skopeo 1.22.2 and the tested Docker-media-type multi-architecture source
- the production path requires checksum and digest trust anchors retained separately from the transfer media
- the Skopeo path requires exact digest equality after the round trip; an archive checksum alone is not considered sufficient

### DIFF
--- a/docs/airgap/mirror-docker-images.md
+++ b/docs/airgap/mirror-docker-images.md
@@ -179,7 +179,8 @@ REGISTRY_DIGEST="$(skopeo inspect --format '{{.Digest}}' "docker://${REGISTRY_SO
 printf '%s\n' "${REGISTRY_DIGEST}" > /trusted/registry-3.1.1.source-digest
 
 docker pull "registry@${REGISTRY_DIGEST}"
-docker save -o /tmp/images/registry-3.1.1.tar "registry@${REGISTRY_DIGEST}"
+docker tag "registry@${REGISTRY_DIGEST}" registry:3.1.1
+docker save -o /tmp/images/registry-3.1.1.tar registry:3.1.1
 (
     set -euo pipefail
     cd /tmp/images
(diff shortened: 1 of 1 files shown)


dholt:
Updated this PR after a fresh dual independent review of the previous head:

- Rebased onto current `master` (now includes #1374).
- The primary Skopeo workflow and the import path now provision and use a configurable `TRUSTED_RECORD_DIR` (default `~/deepops-trusted-records`) instead of writing to a hardcoded, unprovisioned `/trusted`, which would abort the `set -euo pipefail` procedures on any host without that directory.
- The Skopeo 1.22.2 digest-preservation limitation is now scoped to the transport it was actually observed on (`oci-archive:`), and the doc ships a runnable reproducer with explicit digest-equality assertions so readers can evaluate their own Skopeo version and transports.

All shell blocks pass `bash -n`; public checks are rerunning on the new head.

michael-balint (approved):

PR #1390, opened 2026-09-03
nvidia-peer-memory: prefer in-tree nvidia_peermem over legacy nv_peer_mem

100-JM:
## Problem
`roles/nvidia-peer-memory` assumes the out-of-tree `nv_peer_mem` DKMS module on DGX systems. Since driver R470.42.01 the equivalent `nvidia_peermem` module ships in-tree with the driver and `nv_peer_mem` is not available, so the role fails on current DGX OS:

    modprobe: FATAL: Module nv_peer_mem not found in directory /lib/modules/6.8.0-106-generic

There is no variable to disable the role from `slurm-cluster.yml`; only the `nvidia-peer-memory` tag.

Reproduced on DGX OS 7.5.0, driver 580.126.20.

## Fix
Probe for `nvidia_peermem` with `modinfo`. When present, retire any legacy state first — stop and disable `nv_peer_mem.service` if it exists, unload `nv_peer_mem` — then load `nvidia_peermem`, and skip the legacy DKMS/`nv_peer_mem`/service path. When the in-tree module is absent, behaviour is unchanged.

Every step is idempotent (`service_facts` gates the service task; `modprobe` only acts when the module state differs), so a second run reports `ok` throughout.

## Verification
Failure reproduced on DGX OS 7.5.0 / driver 580.126.20. On the same system `lsmod` shows the in-tree module already loaded (`nvidia_peermem 16384 0`, used by `ib_uverbs`), so the new probe takes the in-tree branch and the legacy DKMS/nv_peer_mem path is skipped.

### DIFF
--- a/roles/nvidia-peer-memory/tasks/main.yml
+++ b/roles/nvidia-peer-memory/tasks/main.yml
@@ -4,11 +4,32 @@
     path: /etc/dgx-release
   register: is_dgx
 
+# Driver R470.42.01 and later ship nvidia_peermem in-tree. On those
+# systems the out-of-tree nv_peer_mem DKMS module does not exist.
+- name: Check for in-tree nvidia_peermem module
+  command: modinfo nvidia_peermem
+  register: nvidia_peermem_info
+  failed_when: false
+  changed_when: false
+  when:
+    - ansible_local['gpus']['count']
+    - is_dgx.stat.exists
+
+- name: Load in-tree nvidia_peermem
+  modprobe:
+    name: nvidia_peermem
+    state: present
+  when:
+    - ansible_local['gpus']['count']
+    - is_dgx.stat.exists
+    - nvidia_peermem_info.rc == 0
+
 - name: Autoinstall DKMS modules
   command: dkms autoinstall
   when:
     - ansible_local['gpus']['count']
     - is_dgx.stat.exists
+    - nvidia_peermem_info.rc != 0
 
 - name: Modprobe nv_peer_mem
   modprobe:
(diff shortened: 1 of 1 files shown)


dholt (changes requested):
Before loading nvidia_peermem, handle the legacy state explicitly: stop and disable the nv_peer_mem service when present, unload nv_peer_mem as safely required, and then load nvidia_peermem. The transition should also remain successful on a second Ansible run.

---

100-JM:
Pushed 45c03ba4 to address this.

When the in-tree module is available the role now: gathers `service_facts`; stops and disables `nv_peer_mem.service` if it exists; unloads `nv_peer_mem` via `modprobe state: absent`; then loads `nvidia_peermem`. The legacy DKMS path is unchanged and still taken when `modinfo nvidia_peermem` fails.

Second run: `service_facts` gates the service task, and `community.general.modprobe` only calls `modprobe -r` / `modprobe` when `/proc/modules` disagrees with the requested state, so every step reports `ok` once converged. Exercised the branch selection locally against ansible-core 2.17.14 for non-DGX, in-tree with legacy service present, in-tree without it, and legacy-only hosts.

One deliberate choice: if `nv_peer_mem` is in use and cannot be unloaded, the task fails rather than loading `nvidia_peermem` alongside it. Happy to relax that if you prefer a warning.

100-JM:
@dholt The requested changes are pushed (see the comment above for details) — ready for another look when you have a moment.

dholt (approved):
The revised legacy-provider transition addresses my earlier change request: the old service is stopped/disabled and its module unloaded before the in-tree provider is loaded. The skipped-probe handling is also explicit. CI is green.

Please preserve the original commits with a merge commit rather than squash. A human maintainer will make the final merge decision.

PR #1398, opened 2026-09-07
chore(slurm): derive PAM module directory from target architecture

100milliongold:
### What this is

A hygiene change, not a bug fix. **The current code works on ARM** — this
explains why, and why the hardcoded tuple is still worth removing.

`roles/slurm/vars/ubuntu.yml` hardcodes the Debian multiarch tuple:

```yaml
slurm_pam_lib_dir: /lib/x86_64-linux-gnu/security
```

On an aarch64 host this makes the play create and use
`/lib/x86_64-linux-gnu/security`, which is not that host's multiarch path.

### Why nothing breaks today

`slurm_pam_lib_dir` has exactly two direct consumers, and they cancel out:

| Consumer | Use |
|---|---|
| `roles/slurm/defaults/main.yml:19-20` | `--with-pam_dir={{ slurm_pam_lib_dir }}` in `slurm_configure` |
| `roles/slurm/tasks/build.yml:209-219` | indirect — builds `contribs/pam_slurm_adopt` and `make install`s it into the `--with-pam_dir` set above |
| `roles/slurm/tasks/compute.yml:165,167,180` | the absolute path written into `/etc/pam.d/sshd` |

`roles/slurm/tasks/build.yml:166,172` runs `slurm_configure`, so DeepOps builds
Slurm from source, and lines 209-219 then build and install `pam_slurm_adopt`
itself (209-213 and 215-219, both guarded by `when: slurm_build`):

```yaml
- name: install pam_slurm_adopt
  shell: "make -j$(nproc) install >> ../../build.log 2>&1"
  args:
    chdir: "{{ slurm_build_dir }}/contribs/pam_slurm_adopt"
```

That `make install` honours the `--with-pam_dir` from configure. So the same
variable decides *where the build puts the module* and *where PAM looks for it*.
Both move together, `make install` creates the directory, and the module loads.
The PAM lines are also prefixed with `-`, so even a missing module would be
skipped rather than fail the login.

### Why change it anyway

1. **It creates a directory named for the wrong architecture.** An aarch64 node
   ends up with `/lib/x86_64-linux-gnu/security`, outside the standard multiarch
   layout, which misleads anyone inspecting the node.
2. **The sibling vars file is already architecture-agnostic.**
   `roles/slurm/vars/redhat.yml:2` is `/lib64/security`, correct on every
   architecture. Only the Ubuntu path carries a hardcoded tuple.
3. **It stops being self-consistent as soon as the module is not built here** —
   for example an air-gapped flow that ships a prebuilt binary, where the
   install path is decided elsewhere.
4. **It breaks if `--with-pam_dir` goes away** — switching to the distro package
   (`slurm-wlm`, which installs into the real multiarch path) or to any variant
   that does not pass that flag.

### Change

Resolve the tuple from `ansible_architecture` through an explicit map. The map is
explicit rather than an `<arch>-linux-gnu` format string because two real
architectures do not follow that shape:

| ansible_architecture | resolved directory | note |
|---|---|---|
| x86_64  | `/lib/x86_64-linux-gnu/security` | unchanged from today |
| aarch64 | `/lib/aarch64-linux-gnu/security` | |
| arm64   | `/lib/aarch64-linux-gnu/security` | input alias, see below |
| armv7l  | `/lib/arm-linux-gnueabihf/security` | not `armv7l-linux-gnu` |
| ppc64le | `/lib/powerpc64le-linux-gnu/security` | not `ppc64le-linux-gnu` |

A format string would have produced `armv7l-linux-gnu` and `ppc64le-linux-gnu`,
neither of which exists on those hosts. The fallback keeps format-string
behaviour for an unlisted architecture so the role does not hard-fail, and the
comment tells the next person to add an entry.

`arm64` is included to match existing repository practice:
`roles/nvidia-dgx/vars/ubuntu-24.04.yml:2` already tests
`ansible_architecture in ['aarch64', 'arm64']`.

x86_64 resolves to the same value as before, so this is a no-op on existing
clusters.

### Verification

Directories checked inside real Debian-family containers on each architecture:

```
$ docker run --rm --platform linux/arm64 ubuntu:24.04 sh -c 'uname -m; ls -d /lib/*/security'
aarch64
/lib/aarch64-linux-gnu/security

$ docker run --rm --platform linux/amd64 ubuntu:24.04 sh -c 'uname -m; ls -d /lib/*/security'
x86_64
/lib/x86_64-linux-gnu/security

$ docker run --rm --platform linux/arm/v7 debian:12 sh -c 'uname -m; ls -d /lib/*-linux-*'
armv7l
/lib/arm-linux-gnueabihf

$ docker run --rm --platform linux/ppc64le debian:12 sh -c 'uname -m; ls -d /lib/*-linux-*'
ppc64le
/lib/powerpc64le-linux-gnu
```

Template rendering checked for all five map keys plus one unlisted architecture;
each resolved to the table above.

### DIFF
--- a/roles/slurm/tasks/compute.yml
+++ b/roles/slurm/tasks/compute.yml
@@ -158,6 +158,35 @@
   tags:
     - localusers
 
+# slurm_pam_lib_dir is architecture-derived, so an existing installation can
+# hold its module at a path this run no longer selects. build.yml skips the
+# rebuild when the installed version already matches, which would otherwise
+# leave PAM pointing at a file that is not there -- and a wrong -account line
+# in /etc/pam.d/sshd locks operators out of the node. Check the module before
+# writing either block, and name the rebuild switch in the failure.
+- name: locate the pam_slurm_adopt module
+  stat:
+    path: "{{ slurm_pam_lib_dir }}/pam_slurm_adopt.so"
+  register: slurm_pam_adopt_module
+  tags:
+    - config
+  when: slurm_enable_pam_slurm_adopt
+
+- name: fail when pam_slurm_adopt is missing from the selected directory
+  fail:
+    msg: >-
+      {{ slurm_pam_lib_dir }}/pam_slurm_adopt.so does not exist, so
+      /etc/pam.d/sshd will not be pointed at it. This usually means Slurm was
+      built against a different PAM directory and build.yml skipped the
+      rebuild because the installed version already matches. Rebuild with
+      slurm_force_rebuild=true, or set slurm_pam_lib_dir to the directory the
+      installed module actually lives in.
+  tags:
+    - config
+  when:
+  - slurm_enable_pam_slurm_adopt
+  - not slurm_pam_adopt_module.stat.exists
+
 - name: configure /etc/pam.d/sshd to restrict node access without jobs
   blockinfile:
     path: /etc/pam.d/sshd
--- a/roles/slurm/tasks/compute.yml
+++ b/roles/slurm/tasks/compute.yml
@@ -167,6 +167,7 @@
 - name: locate the pam_slurm_adopt module
   stat:
     path: "{{ slurm_pam_lib_dir }}/pam_slurm_adopt.so"
+    follow: true
   register: slurm_pam_adopt_module
   tags:
     - config
@@ -175,8 +176,10 @@
 - name: fail when pam_slurm_adopt is missing from the selected directory
   fail:
     msg: >-
-      {{ slurm_pam_lib_dir }}/pam_slurm_adopt.so does not exist, so
-      /etc/pam.d/sshd will not be pointed at it. This usually means Slurm was
+      {{ slurm_pam_lib_dir }}/pam_slurm_adopt.so is not an existing regular
+      file, so /etc/pam.d/sshd will not be pointed at it. A dangling symlink,
+      a directory or a FIFO at that path fails this check as well.
+      This usually means Slurm was
       built against a different PAM directory and build.yml skipped the
       rebuild because the installed version already matches. Rebuild with
       slurm_force_rebuild=true, or set slurm_pam_lib_dir to the directory the
@@ -185,7 +188,7 @@
     - config
   when:
   - slurm_enable_pam_slurm_adopt
-  - not slurm_pam_adopt_module.stat.exists
+  - not slurm_pam_adopt_module.stat.exists or not (slurm_pam_adopt_module.stat.isreg | default(false))
 
 - name: configure /etc/pam.d/sshd to restrict node access without jobs
   blockinfile:
--- a/roles/slurm/vars/ubuntu.yml
+++ b/roles/slurm/vars/ubuntu.yml
@@ -1,5 +1,17 @@
 ---
-slurm_pam_lib_dir: /lib/x86_64-linux-gnu/security
+# Debian multiarch tuple for the PAM module directory. This is an explicit map
+# because the tuple is not derivable from ansible_architecture: armv7l maps to
+# arm-linux-gnueabihf and ppc64le to powerpc64le-linux-gnu, neither of which
+# follows the "<arch>-linux-gnu" shape. Add an entry here for a new architecture.
+# 'arm64' matches existing practice in roles/nvidia-dgx/vars/ubuntu-24.04.yml.
+slurm_pam_arch_map:
+  x86_64: x86_64-linux-gnu
+  aarch64: aarch64-linux-gnu
+  arm64: aarch64-linux-gnu
+  armv7l: arm-linux-gnueabihf
+  ppc64le: powerpc64le-linux-gnu
+
+slurm_pam_lib_dir: "/lib/{{ slurm_pam_arch_map.get(ansible_architecture, ansible_architecture + '-linux-gnu') }}/security"
 
 slurm_build_deps:
   - build-essential


dholt (changes requested):
The architecture mapping is useful, but an existing-install migration case needs a guard before this merges.

`build.yml` skips rebuilding when the installed Slurm version already matches. `compute.yml` still rewrites PAM to the newly selected module path. An existing ARM/POWER installation can therefore keep its module at the old path while PAM is changed to reference a missing file.

Verify the effective module exists before activating the new PAM configuration, and provide a controlled rebuild/migration path. Please test the same-version upgrade case as well as a fresh build; do not silently change a working login configuration to a nonexistent module.

100milliongold:
Guard added in `1faf8f0`.

## The gap

`build.yml:51` skips the rebuild when the installed version already matches:

```yaml
- name: don't build slurm if it's already installed, unless forced
  set_fact:
    slurm_build: no
  when: slurmd_reported_version.stdout == slurm_version and not slurm_force_rebuild
```

`compute.yml` rewrote `/etc/pam.d/sshd` regardless, so an existing ARM or POWER node kept its module at the old path while PAM was pointed at the newly derived one.

## The guard

`compute.yml` now stats the module before either `blockinfile` task and stops if it is absent, naming `slurm_force_rebuild` — the controlled rebuild path already in `build.yml` — and `slurm_pam_lib_dir` for sites whose module genuinely lives elsewhere:

```yaml
- name: locate the pam_slurm_adopt module
  stat:
    path: "{{ slurm_pam_lib_dir }}/pam_slurm_adopt.so"
  register: slurm_pam_adopt_module
  when: slurm_enable_pam_slurm_adopt

- name: fail when pam_slurm_adopt is missing from the selected directory
  fail:
    msg: >-
      ... Rebuild with slurm_force_rebuild=true, or set slurm_pam_lib_dir to
      the directory the installed module actually lives in.
  when:
  - slurm_enable_pam_slurm_adopt
  - not slurm_pam_adopt_module.stat.exists
```

Both existing `blockinfile` tasks already run only under `slurm_enable_pam_slurm_adopt`, and the guard carries the same condition, so deployments that do not use `pam_slurm_adopt` are unaffected.

## Evidence

Run with `ansible-playbook` against a temporary PAM file, `slurm_pam_lib_dir` pointed at a directory that either holds `pam_slurm_adopt.so` or does not.

| case | exit | `/etc/pam.d/sshd` rewritten |
| --- | --- | --- |
| fresh build — module present at the selected path | 0 | yes |
| same-version upgrade — module still at the old path | 2 | no, fails with the message above |

Without the guard, the second case is the one that matters:

```
exit code = 0
# BEGIN ANSIBLE MANAGED BLOCK (ansible-role-slurm)
-account sufficient .../aarch64-linux-gnu/security/pam_slurm_adopt.so action_no_jobs=ignore
-account required .../aarch64-linux-gnu/security/pam_slurm_adopt.so
# END ANSIBLE MANAGED BLOCK (ansible-role-slurm)

module present at that path: no
```

A successful run, and `-account required` now naming a file that is not there.

roles/slurm/tasks/compute.yml
```diff
@@ -158,6 +158,35 @@
...
 
+# slurm_pam_lib_dir is architecture-derived, so an existing installation can
+# hold its module at a path this run no longer selects. build.yml skips the
+# rebuild when the installed version already matches, which would otherwise
+# leave PAM pointing at a file that is not there -- and a wrong -account line
+# in /etc/pam.d/sshd locks operators out of the node. Check the module before
+# writing either block, and name the rebuild switch in the failure.
+- name: locate the pam_slurm_adopt module
+  stat:
+    path: "{{ slurm_pam_lib_dir }}/pam_slurm_adopt.so"
```
dholt, line 169:
```suggestion
    path: "{{ slurm_pam_lib_dir }}/pam_slurm_adopt.so"
    follow: true
```

roles/slurm/tasks/compute.yml
```diff
@@ -158,6 +158,35 @@
...
+      /etc/pam.d/sshd will not be pointed at it. This usually means Slurm was
+      built against a different PAM directory and build.yml skipped the
+      rebuild because the installed version already matches. Rebuild with
+      slurm_force_rebuild=true, or set slurm_pam_lib_dir to the directory the
+      installed module actually lives in.
+  tags:
+    - config
+  when:
+  - slurm_enable_pam_slurm_adopt
+  - not slurm_pam_adopt_module.stat.exists
```
dholt, line 188:
```suggestion
  - not slurm_pam_adopt_module.stat.exists or not (slurm_pam_adopt_module.stat.isreg | default(false))
```

dholt (changes requested):
The new pre-activation guard addresses the missing-module case. One small gap remains: the default stat behavior admits a dangling symlink, and an existence check alone also admits a directory or FIFO named `pam_slurm_adopt.so`.

The two suggestions below follow symlinks and require the resolved object to be a regular file. I checked these guards with Ansible 2.17.14 using six temporary-file cases against the old and corrected code; only regular files and valid links to regular files pass the corrected guard. This was an isolated guard test, not a PAM deployment.

Please apply both suggestions and adjust the diagnostic to say “not an existing regular file” rather than only “does not exist.”

100milliongold:
Both suggestions applied in `7e5d6579`, and the diagnostic now says "is not an existing regular file" instead of "does not exist".

Your reading of the default `stat` behaviour matches what I measured. I ran the six cases in a single play on Ansible core 2.15.5, against the old condition and the corrected one:

| path | old guard passes | corrected guard passes |
|---|---|---|
| regular file | yes | yes |
| symlink to regular file | yes | yes |
| **dangling symlink** | **yes** | no |
| **directory** | **yes** | no |
| **FIFO** | **yes** | no |
| absent | no | no |

So the old guard admitted three of the four bad shapes, not just the dangling symlink. `follow: true` turns the underlying `lstat` into a `stat`, which is what drops the dangling link, and `isreg` drops the directory and the FIFO.

`| default(false)` on `isreg` is load-bearing: with `follow: true` and a missing target the fact is absent rather than false, so the bare attribute would raise instead of failing the check.

Like yours, this was an isolated guard test on temporary files, not a PAM deployment.

roles/slurm/tasks/compute.yml
100milliongold: Applied in `7e5d6579`. `follow: true` is what drops the dangling symlink: it turns the underlying `lstat` into a `stat`, which fails on a link with no target.

roles/slurm/tasks/compute.yml
100milliongold: Applied in `7e5d6579`. Measured on Ansible core 2.15.5: the old condition passed a dangling symlink, a directory and a FIFO; this one passes only a regular file or a symlink resolving to one. `| default(false)` matters because with `follow: true` and a missing target the `isreg` fact is absent rather than false.

dholt (approved):
The architecture mapping and pre-activation guard now address the requested migration and file-shape corrections. Independent review found no remaining functional blocker.

Validation: 28 isolated Ansible invocations covered architecture/configure/PAM path agreement, both PAM modes, six module-path shapes, disabled adoption, config tags, same-version migration refusal, retry/idempotence, and historical negative controls. A separate 12-case run on Ansible core 2.17.14 and focused Slurm-role syntax check also passed. Only regular files and symlinks resolving to regular files pass; rejected migration cases leave the prior PAM fixture unchanged. All 15 current-head CI checks are green.

The combined merge preview with the GPU-helper PR passed the rendered-helper and PAM-guard regressions as well. These are isolated tests, not a real Slurm build or PAM/SSH deployment. Retaining the regression matrix in CI is a nonblocking follow-up.

PR #1401, opened 2026-09-08
docs: refresh the Enroot and Pyxis example versions

100milliongold:
The snippet under "Updating Enroot and Pyxis" still suggests enroot_version
3.2.0, released 2020-12-02, and slurm_pyxis_version 0.11.1. Point them at
current releases instead.

Raising enroot_version on its own is not enough on RHEL-family hosts. The
nvidia.enroot role builds its RPM URLs with a hardcoded el7 tag, and Enroot
moved to el8 artifacts in 3.4.1 and stopped publishing el7 ones, so any
version from 3.4.1 onward 404s. Document the enroot_rpm_packages override
next to the version.

Also flag 3.4.0, which needs more than the tag: its deb and rpm carry
different release numbers, and the role derives both filenames from a single
enroot_release. It is the only release out of eighteen where they disagree.

### DIFF
--- a/docs/deepops/update-deepops.md
+++ b/docs/deepops/update-deepops.md
@@ -538,9 +538,25 @@ ansible-playbook -l slurm-node playbooks/slurm-cluster/nvidia-dcgm-exporter.yml
 To update Pyxis and/or Enroot, edit your DeepOps configuration and specify the new versions you wise to use:
 
 ```bash
-slurm_pyxis_version: "0.11.1"
-enroot_version: "3.2.0"
+slurm_pyxis_version: "0.24.0"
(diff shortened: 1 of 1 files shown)


dholt (commented):
The refreshed examples and RPM workaround are useful. One nonblocking wording clarification: Enroot 3.4.0 cannot use one common release value across DEB/RPM hosts, but per-host or per-family release overrides can handle that difference. Existing site configurations still need deliberate adoption of the changes in #1402.

CI is green. Please preserve the original commit with a merge commit rather than squash.

dholt (approved):

PR #1403, opened 2026-09-09
Spare localusers from the epilog's killall

100milliongold:
## What happens

`epilog.d/41-lastuserjob-ssh` checks `/etc/slurm/localusers.backup` before it touches a user:

```bash
if grep -q -w "$SLURM_JOB_USER" /etc/slurm/localusers.backup ; then
    exit 0  # don't revoke access for these users
fi
```

`epilog.d/40-lastuserjob-processes` does not, even though it is the broader of the two:

```bash
if [ "$SLURM_JOB_USER" != root ]; then
    if killall -9 -u "$SLURM_JOB_USER" ; then
```

`killall -9 -u` reaps **every** process the user owns on that node. That includes an interactive login shell and the `sshd` session carrying it.

So an operator listed in `localusers.backup` — listed there precisely so their access is not cut off — is disconnected the moment their job's epilog runs.

## How we hit it

Two-node cluster where the controller is also a compute node, so the account submitting jobs also has a login shell on a node that runs the epilog.

```
11:09:55  Running /etc/slurm/epilog.d/41-lastuserjob-ssh ...
11:09:55  Running /etc/slurm/epilog.d/42-lastuserjob-cleanup ...
11:10:03  epilog for JobId=2 ran for 8 seconds
```

and on the operator's terminal, at the same moment:

```
Connection to <node> closed by remote host.
```

`40` leaves no log line of its own here, because its `logger` call is inside `if killall ...; then` and `killall` takes down the shell that would have carried the message.

There is a second, downstream symptom. Node Health Check runs `check_ps_service -u root -d sshd sshd`; on a socket-activated `sshd` (Ubuntu 24.04) there is no persistent `sshd` process once the last session dies, so NHC failed 2.5 minutes later:

```
error: health_check failed: rc:1 output:ERROR: nhc: Health check failed:
       check_ps_service: Service sshd (process sshd) owned by root not running
```

That NHC check is arguably wrong on its own for socket activation, and is not addressed here.

## The change

Apply the guard `41` already uses, unchanged in form.

Cleanup behaviour is identical for every user **not** in `localusers.backup`. And with `ProctrackType=proctrack/cgroup` the job's own processes are already reaped by Slurm, so this script is a sweep for strays that escaped the cgroup rather than the primary teardown — skipping it for a handful of operator accounts does not leave the node dirty.

## Verified

Applied to both nodes of the cluster above:

```
guard fires for ubuntu (in localusers.backup)  : YES — killall skipped
guard fires for an ordinary user               : no  — killall still runs
```

## Not changed

`42-lastuserjob-cleanup` has the same asymmetry — it removes the user's files under `/tmp` and `/dev/shm` with no exemption check. Its blast radius is much smaller than `killall -9`, so it is left alone here rather than widened into this PR. Happy to follow up if you would rather the three scripts agree.

### DIFF
--- a/roles/slurm/templates/etc/slurm/epilog.d/40-lastuserjob-processes
+++ b/roles/slurm/templates/etc/slurm/epilog.d/40-lastuserjob-processes
@@ -1,6 +1,18 @@
 #!/usr/bin/env bash
 set -ex
 
+# Sibling 41-lastuserjob-ssh already consults this list before it touches a
+# user, so that operator accounts do not lose access when their job ends.
+# `killall -9 -u` is broader than anything 41 does -- it reaps every process
+# the user owns on this node, including their interactive login shell and its
+# sshd session -- so it has to honour the same exemption.
+#
+# Without this an operator who submits from a login shell on a compute node is
+# disconnected the moment the job's epilog runs.
+if grep -q -w "$SLURM_JOB_USER" /etc/slurm/localusers.backup ; then
+    exit 0  # don't kill these users' login sessions
+fi
+
 if [ "$SLURM_JOB_USER" != root ]; then
     if killall -9 -u "$SLURM_JOB_USER" ; then
         logger -s -t slurm-epilog 'Killed residual user processes'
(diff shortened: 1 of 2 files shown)


100milliongold:
Closing in favour of #1404. This branch was cut from a topic branch rather than `master`, so it carried the two commits already under review in #1391. #1404 has the same change on its own, one file, +12/-0.

